PR #1, opened 2017-05-09
add ssd300x300 with coco support

SeaOfOcean:


jason-dai:
Change the name space to `com.intel.analytics.zoo.pipeline`

PR #3, opened 2017-05-16
add python notebook support

SeaOfOcean:
preprocess in python
reuse model.Predict in notebook


zhichao-li:
Roughly the pipeline of prediction looks good. One thing I can think of is those meta&label info can be store into an metaRDD and then post-processing like ```resultRDD.zip(metaRDD).map(postprocessing) ``` in case the data is too large to pull back to driver. 
but my other concerns is not sure what's behavior  of ```zip``` when resultRDD and the metaRDD have the same elements but with different partitions. Maybe we can use dataframe or RDD[(img, meta)] as a safe alternatives.

zhichao-li:
```imgs:RDD[(img, meta)] 
val prediction = ssdmodel.predict(imgs.map(_.img))
val predictionAndMeta = prediction.zip(imgs.map(_.meta)) 
predictionAndMeta.map((img, meta) => postprocessing)```

SeaOfOcean:
https://github.com/SeaOfOcean/analytics-zoo/blob/notebook/notebook/example/SSD.ipynb

PR #5, opened 2017-05-21
refine inference process

hhbyyh:
1. fix maven build by adding aliyun repo.
2. refactor inference evaluate to include both hdfs and local
3. refactor param, remove host, add dataPath and modelPath.
4. use originalSizeCol to replace windows column in decoder
5. add some sample audio files so the customers learn the required data structure.
6. style refine.


hhbyyh:
ping @JerryYanWan @megaSpoon for review.

PR #6, opened 2017-05-22
add dictionary based decoder

hhbyyh:
add dictionary based decoder, also move inference result to main


hhbyyh:
ping @JerryYanWan this is ready now.

PR #26, opened 2017-09-01
add deepspeech2 notebook(in progress)

megaSpoon:
@jason-dai


jason-dai:
I don't think you need to manually manage py4j in the notebook; instead, using pyspark doe that. @jenniew @zhichao-li

PR #42, opened 2017-10-15
add image reader for Spark ML interface.

hhbyyh:
BGRImageReader extends Spark Transformer and provides function for loading image from
local or HDFS paths.

When used as a Transformer, the input column should be of string type, and contains a local
or HDFS path. The output column contains a tuple of imageData (data: Array[Bytes], height: Int,
width: Int, numChannel: Int). The data is stored in a pixel-by-pixel BGR row-wise order.
This follows the OpenCV convention.

Object BGRImageReader also provides static function for loading all the images in a
specific path, where the path support wildcard matching, e.g. /path/1*.jpg

In the case of reading failure (image file malformed), the output column presents null.

unit test added to compare with original BigDL image reader


hhbyyh:
cc @jason-dai @SeaOfOcean

transform/vision/src/main/scala/com/intel/analytics/zoo/transform/vision/image/feature/ReadImageUtil.scala
```diff
@@ -0,0 +1,125 @@
...
+
+object ReadImageUtil extends Serializable {
+
+  OpenCV.loadIfNecessary()
+
+  /**
+   * @param fileContent bytes from an image file.
+   * @return OpenCV Mat
+   */
+  private[vision] def readImageAsMat(fileContent: Array[Byte]): OpenCVMat = {
```
SeaOfOcean: this is same as OpenCVMat.toMat, we can reuse it

transform/vision/src/main/scala/com/intel/analytics/zoo/transform/vision/image/feature/ReadImageUtil.scala
```diff
@@ -0,0 +1,125 @@
...
+      val floats = bytes.map(b => (b & 0xff) / divisor)
+      (floats, h, w, c)
+    } catch {
+      case e: Exception =>
+        println(e)
+        null
+    }
+  }
+
+  private[vision] def mat2Bytes(mat: OpenCVMat): (Array[Byte], Int, Int, Int) = {
```
SeaOfOcean: This is similar to OpenCVMat.toBytesBuf, maybe we can merge these two?

transform/vision/src/main/scala/com/intel/analytics/zoo/transform/vision/image/feature/ReadImageUtil.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  /**
+   * @param fileContent bytes representation for an image file.
+   * @param smallSideSize the size of the smallest side after resize
+   * @param divisor divide each pixel by divisor. E.g. if divisor = 255, each pixel is in [0, 1]
+   * @return (data, height, width, numChannel)
+   */
+  def readImageAsFloats(
```
SeaOfOcean: similar to toFloatBuf & toFloats, can we merge them?

SeaOfOcean:
There are some functions in common between BGRImageReader and OpenCVMat, maybe we can merge these two classes as one?

jason-dai:
I think we would like to unify these concepts into a hierarchy: 
1) An image is a bytesImage, floatsImage or OpenCVMatImage
2) ImageFeature represents a single sample that contains the image data (in addition to label, etc.)
3) Currently we use RDD[ImageFeature] for a set of images; maybe we can build something like ImageFrame?

PR #51, opened 2017-10-27
update readme

zhangxiaoli73:
update readme, add "--driver-class-path deepspeech2-0.1-SNAPSHOT-jar-with-dependencies.jar", otherwise there will be error about ClassNotFound


hhbyyh:
is it possible use --jar?
so that the script can be used in cluster.

zhangxiaoli73:
the error is caused by loading model. I have tried use --jars, but it doesn't work and error happens.
As long as we can load model on driver, there will be ok on cluster.

hhbyyh:
OK. Let's merge this first

PR #53, opened 2017-10-30
Object detection model zoo scala python example

SeaOfOcean:


SeaOfOcean:
python example: apps/ssd/SSD.ipynb apps/ssd/utils.py
scala example: com.intel.analytics.zoo.pipeline.common.modelzoo
@jason-dai

PR #56, opened 2017-10-30
update to bigdl-0.3

zhangxiaoli73:
update to bigdl dependency to bigdl-0.3


pipeline/deepspeech2/pom.xml
```diff
@@ -80,7 +80,7 @@
         <dependency>
             <groupId>com.intel.analytics.bigdl</groupId>
             <artifactId>bigdl-SPARK_2.0</artifactId>
-            <version>0.1.0</version>
+            <version>0.3.0-SNAPSHOT</version>
```
hhbyyh: I assume this change will break the current example, right? since our model is still just a Java serialization. 

Let's try to get a DS 2 model saved with the BigDL 0.3 first, sounds good?

zhangxiaoli73:
I have changed some deepspeech class to bigdl, and generate a new model.
With new model, InferenceExample output is same with before.
InferenceEvaluate output is same with before. Output is:

hhbyyh:
Thanks @zhangxiaoli73 is this ready for review? Please also upload a new model saved with 0.3

zhangxiaoli73:
Sorry, I am not sure how to upload new model

zhangxiaoli73:
Besides, I add the model definition in this pr (Class DeepSpeech2ModelLoader). If you think it is no need, I will delete it.

hhbyyh:
model definition is alright. For uploading the model, any place that's accessible is OK. google, dropbox, or even baiduyun. 

I remember we can access google with proxy, it's not working now?

pipeline/deepspeech2/README.md
```diff
@@ -25,4 +25,20 @@ https://drive.google.com/open?id=0B9zID9CU9HQeU1luc2ZKSHA1MjA
    ```
 
 
+### Run inference with evaluate:
 
+1. Download model file "dp2.bigdl" from the link above.
+
+2. Import the project into IDE or build with "mvn clean package".
+
+3. script to run ds2 inference evaluate:
```
hhbyyh: How about rename this as 
4. script to run ds2 inference evaluate:

and remove 1, 2 and the new section.

pipeline/deepspeech2/README.md
```diff
@@ -25,4 +25,20 @@ https://drive.google.com/open?id=0B9zID9CU9HQeU1luc2ZKSHA1MjA
...
+
```
hhbyyh: For evaluation, we need both the audio files and transcripts.

We need to provide some instruction about how to prepare the evaluation data if we want to add the example.

pipeline/deepspeech2/README.md
zhangxiaoli73: OK

pipeline/deepspeech2/README.md
zhangxiaoli73: OK, we have supported audio files, so we have to add transcripts? Is there any example about how to prepare?

hhbyyh: Eh, evaluation means compare the model output with the transcripts. So if you've tried the evaluation example, you should already have them.

Just write a few sentences about the file structure that evaluation requires.

jason-dai:
Maybe we can upload the model to S3 @zhangxiaoli73

pipeline/deepspeech2/README.md
zhangxiaoli73: ok, I changed the readme and you can have a review

zhangxiaoli73:
ok, I will have a try.

zhangxiaoli73:
@hhbyyh I have uploaded new model and updated readme, please have a review if possible. Thank you.

hhbyyh:
LGTM. Thanks for the update.

PR #62, opened 2017-11-08
Add more utilities for vision and some code refactor

SeaOfOcean:
* Add LocalImageFrame, DistributedImageFrame
* Read local/distributed folder
* Read/write image parque file
* Add more comments

PR #72, opened 2017-11-21
Add model predict api

SeaOfOcean:
This pr is based on the change of https://github.com/intel-analytics/analytics-zoo/pull/62

### DIFF
--- a/pipeline/objectDetection/common/src/main/scala/com/intel/analytics/zoo/pipeline/common/model/Model.scala
+++ b/pipeline/objectDetection/common/src/main/scala/com/intel/analytics/zoo/pipeline/common/model/Model.scala
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.intel.analytics.bigdl.pipeline.common.model
+
+import com.intel.analytics.bigdl._
+import com.intel.analytics.bigdl.dataset.{SampleToMiniBatch}
+import com.intel.analytics.bigdl.nn.SpatialShareConvolution
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.zoo.transform.vision.image._
+import com.intel.analytics.zoo.pipeline.common.model.Preprocessor._
+import com.intel.analytics.bigdl.models.utils.ModelBroadcast
+import com.intel.analytics.bigdl.utils.{Engine, MklBlas}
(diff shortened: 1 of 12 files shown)


pipeline/objectDetection/common/src/main/scala/com/intel/analytics/zoo/pipeline/common/model/Model.scala
```diff
@@ -0,0 +1,129 @@
...
+      val miniBatch = localTransformer(partition)
+      miniBatch.flatMap(batch => {
+        val result = localModel.forward(batch.getInput()).toTensor[T]
+        val batchOut = if (result.dim() == 1) {
+          Array(result)
+        } else {
+          result.split(1)
+        }
+        batch.imageFeatures.zip(batchOut).map(x => {
+          x._1(ImageFeature.feature) = x._2
```
jason-dai: Change to ImageFeature.predit or ImageFeature.output

jason-dai:
We can first merge the ImageFrame support to BigDL, and then add `predict(ImageFrame)` to AbstractModule and Predictor.

And instead of introducing new ImageFeatureMiniBatch and ImageFeatureToBatch, we can probably reuse the existing MiniBatch and SampleToMiniBatch as follows in `predict(ImageFrame)` :

```scala
    data.rdd.mapPartitions { partition =>
      val localModel = modelBroad.value()
      val localTransformer = broadcastToBatch.value.cloneTransformer()
      val localPostProcessor = broadcastPostProcess.value
      
      partition.grouped(size).flatMap { b =>
        val batch = localTransformer(b.iterator).next // need to move buffers out as class members
        val result = localModel.forward(batch.getInput()).toTensor[T]
        val batchOut = if (result.dim() == 1) {
          Array(result)
        } else {
          result.split(1)
        }
        b.zip(batchOut).map({
           ...
        })
      }
```

pipeline/objectDetection/common/src/main/scala/com/intel/analytics/zoo/pipeline/common/model/Model.scala
```diff
@@ -0,0 +1,129 @@
...
+
+class Model[A <: Activity : ClassTag, B <: Activity : ClassTag, T: ClassTag]
+(model: AbstractModule[A, B, T])(
+  implicit ev: TensorNumeric[T]) {
+
+  SpatialShareConvolution.shareConvolution(model)
+  model.evaluate()
+
+  // get tag
+  val modelTag = "ssd_vgg16_300x300_pascal"
```
wzhongyuan: Supposed to get  from moduleName ?

pipeline/objectDetection/common/src/main/scala/com/intel/analytics/zoo/pipeline/common/model/Model.scala
```diff
@@ -0,0 +1,129 @@
...
+  val versionTag = "0.3.0"
+  val defaultBatchPerPartition = 4
+
+  private def defaultToBatch(totalBatch: Int, partitionNum: Int) = SampleToMiniBatch[T](
+    batchSize = totalBatch,
+    partitionNum = Some(partitionNum))
+
+  private def defaultPostProcessor: (ImageFeature => ImageFeature) = null
+
+  private def getResources(imf: ImageFrame, partitionNum: Int) = modelTag match {
```
wzhongyuan: How about move it into a separate utility ?

SeaOfOcean:
@jason-dai 

```
      partition.grouped(size).flatMap { b =>
        val batch = localTransformer(b.iterator).next // need to move buffers out as class members
        val result = localModel.forward(batch.getInput()).toTensor[T]
        val batchOut = if (result.dim() == 1) {
          Array(result)
        } else {
          result.split(1)
        }
        b.zip(batchOut).map({
           ...
        })
      }
```

In this way, the transformer will run twice?

PR #83, opened 2017-12-22
 Refine Model Zoo API and example 

SeaOfOcean:


models/src/main/python/bigdl/zoo/models.py
```diff
@@ -0,0 +1,141 @@
...
+    def predict(self, image_frame, output_layer=None,
+                share_buffer=False, predict_key="predict"):
+        res = callBigDlFunc(self.bigdl_type, "modelZooPredict", self.value,
+                             image_frame, output_layer, share_buffer, predict_key)
+        return ImageFrame(res)
+
+class Configure(JavaValue):
+    """
+    predictor configure
+    :param pre_processor preprocessor of ImageFrame before model inference
```
wzhongyuan: Maybe of => on ?

models/src/main/python/bigdl/zoo/models.py
```diff
@@ -0,0 +1,141 @@
...
+    :param post_processor postprocessor of ImageFrame after model inference
+    :param batch_per_partition batch size per partition
+    :param label_map label mapping
```
wzhongyuan: label mapping => mapping from prediction result indexes to real dataset labels ?

models/src/main/python/bigdl/zoo/models.py
```diff
@@ -0,0 +1,141 @@
...
+                pre_processor,
+                post_processor,
+                batch_per_partition,
+                label_map)
+
+    def label_map(self):
+        return callBigDlFunc(self.bigdl_type, "getLabelMap", self.value)
+
+
+
```
wzhongyuan: style : to many empty lines

models/src/main/scala/com/intel/analytics/zoo/models/Predictor.scala
```diff
@@ -17,47 +17,75 @@
...
    * @param imageFrame local or distributed imageFrame
    * @param outputLayer output layer name, if it is null, use output of last layer
    * @param shareBuffer share buffer of output layer
    * @param predictKey key to store prediction result
    * @return imageFrame with prediction
    */
-  def predict[A <: Activity : ClassTag, B <: Activity : ClassTag, T: ClassTag]
-  (model: AbstractModule[A, B, T],
-    imageFrame: ImageFrame,
+  def predict(imageFrame: ImageFrame,
```
wzhongyuan: In case of no default config and no explicit config being passed in , add a null check here for a reminder of passing config?

SeaOfOcean: It has been checked when the Predictor is created

jason-dai:
We need to add Python examples for both classification and object detection.

And we need to properly structures the Python APIs into proper hierarchy?

jason-dai:
We need to add Python examples for both classifications and object detection

And we need to put the Python APIs into proper package hierarchy

@SeaOfOcean @wzhongyuan

wzhongyuan:
@jason-dai   ok, will add the examples, for the API, do we need to separate them into different packages for object detection and classification respectively ?

PR #92, opened 2018-01-09
 Optimize predictor performance and add PaddingParam for variant input  sizes

SeaOfOcean:


models/src/main/scala/com/intel/analytics/zoo/models/Predictor.scala
```diff
@@ -60,9 +67,21 @@ class Predictor[T: ClassTag](
...
+            configure.featurePaddingParam,
+            configure.batchPerPartition)
+        distributedPredictor.predictImage(
+          distributedImageFrame, outputLayer, shareBuffer, predictKey)
+      case localImageFrame: LocalImageFrame =>
+        if (null == localPredictor) localPredictor =
+          new LocalPredictor[T](model,
+            configure.featurePaddingParam,
+            configure.batchPerPartition)
+        localPredictor.predictImage(localImageFrame, outputLayer, shareBuffer, predictKey)
```
wzhongyuan: can this be moved out (together with line 76) ?

SeaOfOcean: distributedPredictor and localPredictor are different classes

wzhongyuan: I mean can we change the first param type  of predictImage(....) for both local distributed predictor to ImageFrame instead of LocalImageFram and DistriIMageFrame so that they have the same signature ?

PR #97, opened 2018-01-18
Storm example

wzhongyuan:
ImageClassification  streaming example with Storm


wzhongyuan:
@jason-dai   do we need to open a new branch for analytics-zoo as well ?  I've made a fix to bigdl to make ImageFrame serializable so in models we have to switch dependency to 0.5, but potentially there would be risk as 0.5 is still under development and not stable. @SeaOfOcean has an application relying on `model` so it's better not to change the indirect dependency

PR #105, opened 2018-02-02
get label map from predictor

wzhongyuan:
A simple change to get label map from predictor instead of reading it  separately

Verified by the example itself

### DIFF
--- a/models/src/main/python/bigdl/zoo/example/ImageClassificationPredict.py
+++ b/models/src/main/python/bigdl/zoo/example/ImageClassificationPredict.py
@@ -32,10 +32,10 @@ def predict(modelPath, imgPath, topN):
     print ("Model Path %s" % modelPath)
     print ("Image Path %s" % imgPath)
     print  ("Top N : %d" % topN)
-    labelMap = read_imagenet_label_map()
     model = Model.loadModel(modelPath)
     imageFrame = ImageFrame.read(imgPath)
     predictor = Predictor(model)
+    labelMap = predictor.configure.label_map()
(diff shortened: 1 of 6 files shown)


models/src/main/python/bigdl/zoo/example/ImageClassificationPredict.py
```diff
@@ -32,10 +32,10 @@ def predict(modelPath, imgPath, topN):
     print ("Model Path %s" % modelPath)
     print ("Image Path %s" % imgPath)
     print  ("Top N : %d" % topN)
-    labelMap = read_imagenet_label_map()
     model = Model.loadModel(modelPath)
     imageFrame = ImageFrame.read(imgPath)
     predictor = Predictor(model)
+    labelMap = predictor.configure.label_map()
```
jason-dai: Shall we add label_map() method to Predictor? @SeaOfOcean

models/src/main/scala/com/intel/analytics/zoo/models/Predictor.scala
```diff
@@ -89,6 +89,14 @@ class Predictor[T: ClassTag](
     // apply post process if defined
     if (null != configure.postProcessor) configure.postProcessor(result) else result
   }
+
+  /**
+   * get label map for predictor
+   */
+
```
SeaOfOcean: remove this empty line

jason-dai:
Please update the scala examples as well.

wzhongyuan:
@jason-dai   scala example updated as well

jason-dai:
LGTM

PR #113, opened 2018-02-28
distributed reinforce

yangw1234:


pipeline/reinforce/src/main/python/rl/reinforce_example.py
```diff
@@ -0,0 +1,153 @@
...
+
+    return records.map(lambda x: (x[0], x[1], (x[2] - mean) / (std + eps)))
+
+
+if __name__ == "__main__":
+    spark_conf = create_spark_conf()
+
+    sc = SparkContext(appName="REINFORCE_CartPole-v1", conf=spark_conf)
+    init_engine()
+    init_executor_gateway(sc)
```
shane-huang: The implementation of this new function is in BigDL, right? is it merged to BigDL master? Does it mean this will not work with the latest release BigDL 0.4.0 because the missing of this function? Any chance we make this work with 0.4.0 version?

pipeline/reinforce/src/main/python/rl/agent.py
```diff
@@ -0,0 +1,81 @@
...
+        raise NotImplementedError
+
+class REINFORCEAgent(Agent):
+
+    def __init__(self, env, horizon = 1000):
+        self.env = env
+        self.horizon = horizon
+        self.sampler = Sampler(env, horizon)
+
+    def sample(self, model, num_steps=None, num_trajs=None):
```
shane-huang: maybe add some simple comments on major functions about the meaning of argument? e.g. num_trajs is not self-explainatory in this interface. 
Also I'd recommend to use the name `act` instead of `sample` for this function, since it's a common practice in many external libs.  Internal function names such as _sample_to_max_steps can keep that way.

pipeline/reinforce/src/main/python/rl/trajectory.py
```diff
@@ -0,0 +1,55 @@
...
+
+class Trajectory(object):
+
+    fields = ["observations", "actions", "rewards", "terminal"]
+
+    def __init__(self):
+        self.data = {k: [] for k in self.fields}
+        self.last_r = 0.0
+
+    def add(self, **kwargs):
```
shane-huang: some simple comments to demonstrate how to use this interface.

pipeline/reinforce/src/main/python/rl/trajectory.py
```diff
@@ -0,0 +1,55 @@
...
+        for k, v in kwargs.items():
+            self.data[k] += [v]
+
+    def is_terminal(self):
+        return self.data["terminal"][-1]
+
+
+class Sampler(object):
```
shane-huang: some simple comments about what this class does.

pipeline/reinforce/src/main/python/rl/reinforce_example.py
yangw1234: It has been merged to BigDL master. Kind of hard to make it work with 0.4.0. In 0.4.0, we can not call `forward` on executor.

pipeline/reinforce/src/main/python/rl/reinforce_example.py
```diff
@@ -0,0 +1,156 @@
...
+# limitations under the License.
+#
+
+from bigdl.nn.layer import *
+from bigdl.nn.criterion import *
+from bigdl.optim.optimizer import *
+from bigdl.util.common import *
+from agent import *
+from utils import *
+from environment import *
```
shane-huang: I only see envrionment is imported. But no new files added. Could it be missing?

yangw1234: added

PR #117, opened 2018-03-14
[WIP, waiting release 0.5.0]update objectdetection to 0.5.0

qiuxin2012:


pipeline/objectDetection/ssd/src/main/scala/com/jd/cv/bigdl/ScoringMain.scala
```diff
@@ -0,0 +1,5 @@
+package com.jd.cv.bigdl
```
qiuxin2012: delete later.

PR #126, opened 2018-05-08
auto version check

yangw1234:
auto version check

### DIFF
--- a/pyzoo/zoo/common/nncontext.py
+++ b/pyzoo/zoo/common/nncontext.py
@@ -55,10 +55,10 @@ def _check_spark_version(sc, report_warn):
     (c_major, c_feature, c_maintenance) = _split_full_version(version_info['spark_version'])
     (r_major, r_feature, r_maintenance) = _split_full_version(sc.version)
     error_message = "The compile time spark version is not compatible with " + \
-        "the Spark runtime version. Compile time version is %s, " % version_info['spark_version'] + \
-        "runtime version is %s. If you want to bypass this check, please set " % sc.version + \
-        "spark.analytics.zoo.versionCheck to false, and if you want to only " + \
-        "report an warning message, please set spark.analytics.zoo.versionCheck.warning to true."
+      "the Spark runtime version. Compile time version is %s, " % version_info['spark_version'] + \
+      "runtime version is %s. If you want to bypass this check, please set " % sc.version + \
+      "spark.analytics.zoo.versionCheck to false, and if you want to only " + \
+      "report an warning message, please set spark.analytics.zoo.versionCheck.warning to true."
     if c_major != r_major:
         if not report_warn:
             print("***************************Usage Error*****************************")
@@ -67,9 +67,10 @@ def _check_spark_version(sc, report_warn):
         else:
             warnings.warn(error_message)
     elif not (c_maintenance == r_maintenance and c_feature == r_feature):
-        warnings.warn("The compile time spark version may not compatible with " +\
-        "the Spark runtime version. Compile time version is %s, " % version_info['spark_version'] +\
-        "runtime version is %s" % sc.version)
+        warnings.warn("The compile time spark version may not compatible with " +
+          "the Spark runtime version. " +
+          "Compile time version is %s, " % version_info['spark_version'] +
+          "runtime version is %s" % sc.version)
 
 
 def _get_bigdl_verion_conf():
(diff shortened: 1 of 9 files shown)


scripts/zoo-build-info
```diff
@@ -0,0 +1,36 @@
...
+#   spark_version - The version of spark currently built against
+#   scala_version - The version of scala currently built against
+#   java_version - The version of java currently built against
+
+RESOURCE_DIR="$1"
+mkdir -p "$RESOURCE_DIR"
+ZOO_BUILD_INFO="${RESOURCE_DIR}"/zoo-version-info.properties
+
+echo_build_properties() {
+  echo zoo_version=$1
```
jason-dai: analytics_zoo_verion

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
```diff
@@ -40,7 +45,81 @@ object NNContext {
...
+                            runtimeVersion: String,
+                            compileTimeVersion: String,
+                            project: String): Unit = {
+    val Array(runtimeMajor, runtimeFeature, runtimeMaintance) =
+      runtimeVersion.split("\\.").map(_.toInt)
+    val Array(compileMajor, compileFeature, compileMaintance) =
+      compileTimeVersion.split("\\.").map(_.toInt)
+    if (!(runtimeMajor == compileMajor && runtimeFeature == compileFeature)) {
+      throw new RuntimeException(s"The compile time $project version is not compatible with" +
+        s" the Spark runtime version. Compile time version is $compileTimeVersion, runtime version" +
```
jason-dai: "the runtime $project version"

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
```diff
@@ -40,7 +45,81 @@ object NNContext {
...
+
+  private def checkVersion(
+                            runtimeVersion: String,
+                            compileTimeVersion: String,
+                            project: String): Unit = {
+    val Array(runtimeMajor, runtimeFeature, runtimeMaintance) =
+      runtimeVersion.split("\\.").map(_.toInt)
+    val Array(compileMajor, compileFeature, compileMaintance) =
+      compileTimeVersion.split("\\.").map(_.toInt)
+    if (!(runtimeMajor == compileMajor && runtimeFeature == compileFeature)) {
```
jason-dai: add an option to ignore the error and just report a warning.

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
```diff
@@ -40,7 +45,81 @@ object NNContext {
     val bigdlConf = Engine.createSparkConf(conf)
     val sc = SparkContext.getOrCreate(bigdlConf)
     Engine.init
+    checkSparkVersion(sc)
```
jason-dai: add an option to bypass the check

scripts/zoo-build-info
jason-dai: yes, better to change to ANALYTICS_ZOO

jason-dai:
In addition to check in getNNContext, can we also add an command line tool such as analytics-zoo.common.CheckVersion?

And for Python, shall we check using Python code or Scala code?

yangw1234:
@jason-dai I have added the command line tool. As for python, I will use python code to check, as when version is not right, calling java function will fail.

yangw1234:
pr validation pass

yangw1234:
@zhichao-li Could you take a look at the python change?

pyzoo/zoo/common/nncontext.py
```diff
@@ -30,7 +31,60 @@ def get_nncontext(conf=None):
     """
 
     sc = get_spark_context(conf)
+    conf = sc._conf
+    if conf.get("spark.analytics.zoo.versionCheck", "True").lower() == "true":
```
zhichao-li: should we always enable the checking as there's no overhead for doing this ?

pyzoo/zoo/common/nncontext.py
```diff
@@ -30,7 +31,60 @@ def get_nncontext(conf=None):
...
+        report_warn = conf.get(
+            "spark.analytics.zoo.versionCheck.reportWarn", "False").lower() == "true"
+        _check_spark_version(sc, report_warn)
     redire_spark_logs()
     show_bigdl_info_logs()
     init_engine()
     return sc
+
+def _split_full_version(version):
```
zhichao-li: I think it would be better to wrap these utilities into a class or static methods. i.e `check_spark_version` , `get_compile_spark_version` , or `get_zoo_compile_properties` or pick any name which meaningful to you.

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
```diff
@@ -43,83 +129,16 @@ object NNContext {
    */
   def getNNContext(conf: SparkConf = null): SparkContext = {
     val bigdlConf = Engine.createSparkConf(conf)
+    if (bigdlConf.getBoolean("spark.analytics.zoo.versionCheck", true)) {
```
jason-dai: "analytics.zoo.versionCheck", not "spark.analytics.zoo.versionCheck"

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
```diff
@@ -30,6 +30,92 @@ object NNContext {
...
+        s".versionCheck.reportWarn to true."
+      val diffLevel = if (runtimeMajor != compileMajor) {
+        1
+      } else if (runtimeFeature != compileFeature) {
+        2
+      } else {
+        3
+      }
+      if (diffLevel <= level && !reportWarning) {
+        throw new RuntimeException(warnMessage + errorMessage)
```
jason-dai: better to report error and exist, instead of throwing an exception.

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
```diff
@@ -43,83 +129,16 @@ object NNContext {
    */
   def getNNContext(conf: SparkConf = null): SparkContext = {
     val bigdlConf = Engine.createSparkConf(conf)
+    if (bigdlConf.getBoolean("spark.analytics.zoo.versionCheck", true)) {
+      val reportWarning =
+        bigdlConf.getBoolean("spark.analytics.zoo.versionCheck.reportWarn", false)
```
jason-dai: change to "spark.analytics.zoo.versionCheck.warning"

yangw1234:
pr validation pass

PR #127, opened 2018-05-08
Python WideAndDeep Recommender

hkvision:
WideAndDeep Python model with unit tests.
Add some utilities compatible with Scala side.


hkvision:
Synced with @jason-dai offine. Merge it.

PR #143, opened 2018-05-09
NNFrames: model create with size

hhbyyh:
create NNModel and NNClassifierModel with (featuresize).

### DIFF
--- a/pyzoo/zoo/pipeline/nnframes/nn_classifier.py
+++ b/pyzoo/zoo/pipeline/nnframes/nn_classifier.py
@@ -337,6 +337,24 @@ class NNModel(JavaTransformer, HasFeaturesCol, HasPredictionCol, HasBatchSize, J
         return NNModel(model, chained_transformer, jvalue, bigdl_type)
 
     @classmethod
+    def createWithSize(cls, model, feature_size, jvalue=None, bigdl_type="float"):
+        """
+        Construct a NNModel with a feature size. The constructor is useful
+        when the feature column contains the following data types:
+        Float, Double, Int, Array[Float], Array[Double], Array[Int] and MLlib Vector. The
+        feature data are converted to Tensors with the specified sizes before sending
+        to the model.
+
+        :param model: BigDL Model to be trained.
+        :param feature_size: The size (Tensor dimensions) of the feature data. e.g. an image
+                            may be with width * height = 28 * 28, featureSize = Array(28, 28).
+        :param jvalue: Java object create by Py4j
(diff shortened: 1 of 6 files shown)


hhbyyh:
cc @dding3 @jason-dai

pyzoo/zoo/pipeline/nnframes/nn_classifier.py
```diff
@@ -443,10 +475,28 @@ def __init__(self,  model, feature_preprocessing, jvalue=None, bigdl_type="float
...
+        feature data are converted to Tensors with the specified sizes before sending
+        to the model.
+        :param model: BigDL Model to be trained.
+        :param feature_size: The size (Tensor dimensions) of the feature data. e.g. an image
+                            may be with width * height = 28 * 28, featureSize = Array(28, 28).
+        :param jvalue: Java object create by Py4j
+        :param bigdl_type: optional parameter. data type of model, "float"(default) or "double".
+        """
+        chained_transformer = ChainedPreprocessing([SeqToTensor(feature_size), TensorToSample()])
+        return NNModel(model, chained_transformer, jvalue, bigdl_type)
```
dding3: return NNClassifierModel?

jason-dai:
I think we can use the same method in Python which dynamically check the type of argument?

hhbyyh:
Thanks for the suggestion @dding3 and @jason-dai . I can understand the concern for the names. Indeed it feels complicated right now.

I guess this is a typical overload/dispatch issue for Python. cc @zhichao-li @hkvision to see if there's better suggestion.

We can use a single entrance of the init, with default params or dynamically switch according to param inputs. Yet that would create a mess in the **API documents**, and probably is more difficult for users to learn the API. We will have to write something like, 
"User can set samplePreprocessing, (doc for sample preprocessing).
Alternatively, user can set featurePreprocessing and labelPreprocessing, (doc for feature and label)
alternatively, user can set feature size and label size."

The pros for factory methods are that they can be managed separately and easier for maintain, trouble-shooting and extend/deprecate. **My suggestion** is for this release, we keep using the factory method to support different param inputs, but we can try to create better names for them, like 
`create_with_Tensor_Preprocessing` and `create_with_Tensor_size`. In next release, we can investigate on other possible options or third party libs like `multipledispatch`

jason-dai:
See one such example in Python: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/pipeline/api/keras/engine/topology.py#L41

hhbyyh:
close this as the function has been merged with https://github.com/intel-analytics/analytics-zoo/pull/162

PR #152, opened 2018-05-11
ImageModel to load other model

jenniew:
Add support for ImageModel to load other model.
Add test cases.
Related Issue: https://github.com/intel-analytics/zoo/issues/167


zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageConfigure.scala
```diff
@@ -59,7 +59,7 @@ object ImageConfigure {
     val splits = tag.split(splitter)
     require(splits.length >= 4, s"tag ${tag}" +
       s" needs at least 4 elements, publisher, model, dataset, version")
-    require(splits(0) == "analytics-zoo", "the model publisher needs to be analytics-zoo")
+//    require(splits(0) == "analytics-zoo", "the model publisher needs to be analytics-zoo")
```
dding3: I think we can remove it

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
```diff
@@ -78,10 +85,27 @@ object ImageModel {
...
+    val imageModel = if (model.isInstanceOf[ImageModel[T]]){
+      model.asInstanceOf[ImageModel[T]]
+    }else{
+      val specificModel = modelType.toLowerCase() match {
+        case "objectdetection" => {
+          new ObjectDetector[T]()
+        }
+        case "imageclassification" => {
+          new ImageClassifier[T]()
+        }
```
jason-dai: What is there is no match?

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
```diff
@@ -78,10 +85,27 @@ object ImageModel {
...
+      }
+      specificModel.addModel(model)
+      val name = model.getName()
+      val publisher = name.split(ImageConfigure.splitter)(0)
+      val zooName = name.replace(publisher, "analytics-zoo")
```
jason-dai: Why do we do this? If it's not published by analytics zoo, can we ensure the correctness of pre- & post-precessing?

pyzoo/zoo/models/image/common/image_model.py
```diff
@@ -44,7 +44,7 @@ def get_config(self):
         return ImageConfigure(jvalue=config)
 
     @staticmethod
-    def load_model(path, weight_path=None, bigdl_type="float"):
+    def load_model(path, weight_path=None, model_type=None, bigdl_type="float"):
```
jason-dai: where is this function used?

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
```diff
@@ -78,10 +85,27 @@ object ImageModel {
    * @tparam T Numeric type of parameter(e.g. weight, bias). Only support float/double now.
    * @return
    */
-  def loadModel[T: ClassTag](path: String, weightPath: String = null)
+  def loadModel[T: ClassTag](path: String, weightPath: String = null, modelType: String = null)
```
hkvision: Add some doc for modelType so that users can know what to do?

hkvision: Oh users won't explicitly call ImageModel.loadModel right...

pyzoo/zoo/models/image/common/image_model.py
hkvision: If users don't call this function, then we can remove it from Python side?

zoo/src/test/scala/com/intel/analytics/zoo/models/image/imageclassification/ImageClassificationSpec.scala
```diff
@@ -64,23 +78,49 @@ class ImageClassificationSpec extends FlatSpec with Matchers {
...
+      "imagenet/bigdl_inception-v1-quantize_imagenet_0.4.0.model")
+  }
+
+  "ImageClassifier" should "predict bigdl inception-v1-quantize" in {
+    predict("https://s3-ap-southeast-1.amazonaws.com/bigdl-models/imageclassification/imagenet/" +
+      "bigdl_inception-v1-quantize_imagenet_0.4.0.model")
+  }
 }
 
 class ImageClassifierSerialTest extends ModuleSerializationTest {
```
hkvision: Wondering it is easy to extract some save and load unit test to this class for runtime check whether there's ut for serialization? If not that easy, just ignore this comment and we can postpone it.

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageConfigure.scala
jenniew: removed

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
jenniew: change to just set to original name

pyzoo/zoo/models/image/common/image_model.py
jenniew: I think we don't want user to call this directly, remove it.

jenniew: If we don't want user to see ImageModel class, we can remove the whole class.

jenniew: removed this method

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
```diff
@@ -64,9 +68,15 @@ abstract class ImageModel[T: ClassTag]()(implicit ev: TensorNumeric[T])
   }
 
   def getConfig(): ImageConfigure[T] = config
+
+  def setConfig(imageConfigure: ImageConfigure[T]): Unit = {
```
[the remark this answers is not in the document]
jason-dai: but it can be set once and then used many times, yes?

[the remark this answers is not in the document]
jenniew: I think we can remove setConfig() now.

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
jenniew: We log this error. Do we need to throw exception or exit the program in this situation?

jason-dai: yes

jenniew: updated

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageConfigure.scala
```diff
@@ -59,7 +59,6 @@ object ImageConfigure {
     val splits = tag.split(splitter)
     require(splits.length >= 4, s"tag ${tag}" +
```
jason-dai: I think there are two options if the tag is not from analytics zoo
1) It can report error saying we only support models published by analytics zoo
2) Or it can report a warning and don't set a default config

Either way, we need to update the checking logic here.

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
```diff
@@ -78,10 +85,31 @@ object ImageModel {
...
-    val model = ZooModel.loadModel(path, weightPath).asInstanceOf[ImageModel[T]]
-    model.config = ImageConfigure.parse(model.getName())
-    model
+    val model = Module.loadModule[T](path, weightPath)
+    val imageModel = if (model.isInstanceOf[ImageModel[T]]){
+      model.asInstanceOf[ImageModel[T]]
+    }else{
+      if (modelType == null){
+        logger.error(s"model type is not set.")
+        throw new NullPointerException("model type is null, please set model type")
```
jason-dai: set model type to either "imageclassification" or "objectdetection".

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
jason-dai: only "imageclassification" and "objectdetection" are currently supported.

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageConfigure.scala
jenniew: updated

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
jenniew: updated

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageModel.scala
jenniew: updated

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageConfigure.scala
jason-dai: There is no guarantee that the tag will have 4 splits if the model is not published by analytics zoo

jason-dai:
Some my comments inline (for tag check); otherwise it looks good to me.

zoo/src/main/scala/com/intel/analytics/zoo/models/image/common/ImageConfigure.scala
jenniew: updated

jason-dai: I don't understand the logic here. Which case are you supporting here - 1 or 2?

jason-dai: Please add more comments in the code on how we handle the case that the model is not published by analytics zoo (in which case the tag may not have 4 splits, and does not start with "analytics-zoo").

jason-dai:
LGTM

jenniew:
Jenkins pass: http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/348/

PR #153, opened 2018-05-11
Add site generation for ZOO docs

simanw304:
The file generates the site well. It works with -m (--startserver) command but doesn't work with -s (add scala docs) and -p (add python docs) commands. I am still trying to figure out the reason.

### DIFF
--- a/docs/gen_site.py
+++ b/docs/gen_site.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python
+
+import argparse
+import sys
+import os
+from subprocess import Popen, PIPE
+import subprocess
+
+
+def run_cmd(cmds, err_msg, s=False):
+    cmd = cmds
+    if s:
+        cmd = ' '.join(cmds)
+    try:
+        # print cmd
+        p = Popen(cmd, shell=s)
+        p.communicate()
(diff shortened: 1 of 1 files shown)


simanw304:
Now it only doesn't work with -s (add scala docs) command. I suppose the reason is that it lacks support for necessary plugins in pom.xml.

hkvision:
cc @zhichao-li whether there's something missing in the pom.

pyzoo/docs/doc-web.py
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright 2016 The BigDL Authors.
```
hkvision: Change the header. Refer to other files in Analytics Zoo.

docs/gen_site.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python
+
```
hkvision: Insert the header here.

hkvision:
Is it ready for merge? @helenlly

helenlly:
LGTM

zhichao-li:
let's revert the changes as it would break the PR verification.

PR #157, opened 2018-05-11
Expose gradient clipping API for Keras model

hkvision:


jason-dai:
How about Python?

hkvision:
@jason-dai Doing it now :)

hkvision:
jenkins passed http://172.168.2.101:8080/job/ZOO-PR-Validation/293/

hkvision:
I can't find any unit tests related to gradient clipping. Thus I just add this functionality in original unit tests to test that it can be invoked successfully. cc @jason-dai @dding3

jason-dai:
LGTM

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -97,6 +97,42 @@ def set_checkpoint(self, path, over_write=True):
...
+                      float(max))
+
+    def set_gradient_clipping_by_l2_norm(self, clip_norm):
+        """
+        Call this if you would like to clip gradient to a maximum L2-Norm
+        during the training process.
+        In order to take effect, it needs to be called before fit.
+
+        # Arguments
+        min: The minimum value to clip by. Float.
```
hkvision: wrong doc. modify

dding3:
gradient clipping ut is in a pr, you may want to refer https://github.com/intel-analytics/BigDL/pull/2381/files#diff-9c7647e61b67e51ee5d7d53cd5d60c1c to add some uts

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -117,6 +120,35 @@ abstract class KerasNet[T: ClassTag](implicit ev: TensorNumeric[T])
...
+  }
+
+  /**
+   * Call this if you would like to set constant gradient clipping during the training process.
+   * In order to take effect, it needs to be called before fit.
+   *
+   * @param min The minimum value to clip by. Double.
+   * @param max The maximum value to clip by. Double.
+   */
+  def setConstantGradientClipping(min: Double, max: Double): Unit = {
```
dding3: Looks like when passing min, max to optimizer, they have been converted to float. Can we also use float here so we don't need do the convert.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -117,6 +120,35 @@ abstract class KerasNet[T: ClassTag](implicit ev: TensorNumeric[T])
...
+    this.constantGradientClippingParams = (min, max)
+  }
+
+  /**
+   * Call this if you would like to clip gradient to a maximum L2-Norm during the training process.
+   * In order to take effect, it needs to be called before fit.
+   *
+   * @param clipNorm Gradient L2-Norm threshold. Double.
+   */
+  def setGradientClippingByL2Norm(clipNorm: Double): Unit = {
```
dding3: same as abover

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
hkvision: Oh, I see. The reason why I change to Double is that seems when interacting with python, almost all have type Double instead of Float.
In this case, I will change to Float.

hkvision:
@dding3 I saw the unit tests and seems a little bit complicated... For this release, I think we can just make sure it can be called successfully? Just as here https://github.com/intel-analytics/analytics-zoo/pull/157/files#diff-35ca7e479e057c57fa62b250ea4efe7cR73 
Later if your PR is ready to be merged into BigDL/or Analytics Zoo, I can add further unit tests if you want.
Since what I do here is just a simple wrap and nothing else. If in BigDL this can work properly, then here it will for sure have no problem. Hope that this is OK with you :).

hkvision:
jenkins passed http://172.168.2.101:8080/job/ZOO-PR-Validation/299/

dding3:
Sounds good to me.

hkvision:
Yeah, thanks. For this release, I merge it first and later we can enrich the unit tests together :)

dding3:
Ok. Sure. :)

PR #164, opened 2018-05-11
add object detection into integration test

dding3:


dding3:
integration jekins pass. http://172.168.2.101:8080/job/ZOO-NB-Integration-Test/46/

PR #180, opened 2018-05-15
[Bug Fix] Disable version check on worker

yangw1234:
Disable version check on worker


yangw1234:
@jason-dai In scala, the version check is placed in getnncontext(), which, I assume, will not be called on worker. In python, the version check is placed in the __init__ of zoo package, and this may be called on worker.

yangw1234:
pr validation pass

PR #161, opened 2018-05-11
add image augmentation notebook example and corresponding APIs

changlinzhang:
add image augmentation notebook example
add transformMat method in existing APIs
add other corresponding APIs to Analytics Zoo

PR #184, opened 2018-05-15
image augmentation example using bigdl api

changlinzhang:


jason-dai:
Why do we need this one as we already have https://github.com/intel-analytics/analytics-zoo/pull/161

changlinzhang:
@jason-dai 
Because #161 using ImageSet of analytics-zoo and cant work on python3.
Thus, I want to using ImageFrame of bigdl to try if it could work on python3.
Now it shows both ImageSet and ImageFrame fail to work on python3. I have opened an issue at https://github.com/intel-analytics/zoo/issues/216

PR #190, opened 2018-05-15
localImageset for python3 bug fix

dding3:
Fix https://github.com/intel-analytics/zoo/issues/216

### DIFF
--- a/pyzoo/zoo/feature/image/imageset.py
+++ b/pyzoo/zoo/feature/image/imageset.py
@@ -93,8 +93,8 @@ class LocalImageSet(ImageSet):
             self.value = jvalue
         else:
             # init from image ndarray list and label rdd(optional)
-            image_tensor_list = map(lambda image: JTensor.from_ndarray(image), image_list)
-            label_tensor_list = map(lambda label: JTensor.from_ndarray(label), label_list)\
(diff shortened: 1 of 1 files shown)


pyzoo/zoo/feature/image/imageset.py
```diff
@@ -93,8 +93,8 @@ def __init__(self, image_list=None, label_list=None, jvalue=None, bigdl_type="fl
             self.value = jvalue
         else:
             # init from image ndarray list and label rdd(optional)
-            image_tensor_list = map(lambda image: JTensor.from_ndarray(image), image_list)
-            label_tensor_list = map(lambda label: JTensor.from_ndarray(label), label_list)\
+            image_tensor_list = list(map(lambda image: JTensor.from_ndarray(image), image_list))
```
zhichao-li: should we add a unittest to cover this?

zhichao-li: wondering what's the reason for wrapping this with list?

dding3: @changlinzhang is working on a notebook which will use localImageSet and it will be added to our nightly test.

dding3: in python3, map will return a map object which cannot be serialized by pickle.

zhichao-li:
LGTM

PR #196, opened 2018-05-16
Remove bigdl in jar and add it as a pip dependency

hkvision:
Fix https://github.com/intel-analytics/zoo/issues/226

cc @jason-dai 
I try if we remove bigdl in our package, and add bigdl as a dependency, it can work and the whl file size will definitely be smaller.
I test using the following code after pip install the whl file and start python REPL:
```
from zoo.common.nncontext import *
sc = get_nncontext()
from zoo.pipeline.api.keras.layers import *
from zoo.pipeline.api.keras.models import *
model = Sequential()
model.add(Dense(3, input_shape=(10, )))
```

My preference is to use `bigdl==0.5.0` as a dependency and in this case we don't need to worry anything about the pypi file size. Moreover, if we use the current way, there will possibly be potential problem if users have already install bigdl.
Possibly we need a jar that includes all other dependencies except bigdl in the case where analytics-zoo have some dependencies that bigdl doesn't have.

Wondering if you want to link to other released versions of bigdl (ie. up to a certain commit...)

### DIFF
--- a/pyzoo/setup.py
+++ b/pyzoo/setup.py
@@ -71,20 +71,20 @@ def init_env():
 
 
 def setup_package():
-    jar_file = 'lib/analytics-zoo-'+VERSION+'.jar'
-    if ".dev0" in jar_file:
-        jar_file = jar_file.replace(".dev0", "-SNAPSHOT")
     metadata = dict(
         name='analytics-zoo',
         version=VERSION,
         description='Analytics + AI platform for Apache Spark and BigDL',
         author='Analytics Zoo Authors',
-        author_email='zoo-user-group@googlegroups.com',
+        author_email='analytics-zoo-user-group@googlegroups.com',
         license='Apache License, Version 2.0',
         url='https://github.com/intel-analytics/analytics-zoo',
         packages=['zoo',
                   'zoo.common',
                   'zoo.examples',
+                  'zoo.examples.autograd',
+                  'zoo.examples.imageclassification',
+                  'zoo.examples.nnframes',
                   'zoo.examples.objectdetection',
                   'zoo.examples.textclassification',
                   'zoo.feature',
@@ -93,6 +93,7 @@ def setup_package():
                   'zoo.models.common',
                   'zoo.models.image',
                   'zoo.models.image.common',
+                  'zoo.models.image.imageclassification',
                   'zoo.models.image.objectdetection',
                   'zoo.models.recommendation',
                   'zoo.models.textclassification',
@@ -108,8 +109,8 @@ def setup_package():
         install_requires=['bigdl==0.5.0'],
         dependency_links=['https://d3kbcqa49mib13.cloudfront.net/spark-2.0.0-bin-hadoop2.7.tgz'],
         include_package_data=True,
-        package_dir={"bigdl": '../backend/bigdl/pyspark/bigdl', "zoo.share": TEMP_PATH},
-        package_data={"zoo.share": [jar_file, 'conf/*', 'bin/*',
+        # package_dir={"bigdl": '../backend/bigdl/pyspark/bigdl', "zoo.share": TEMP_PATH},
+        package_data={"zoo.share": ['lib/analytics-zoo*with-dependencies.jar', 'conf/*', 'bin/*',
                                     'extra-resources/*']},
         classifiers=[
             'License :: OSI Approved :: Apache Software License',
--- a/pyzoo/setup.py
+++ b/pyzoo/setup.py
@@ -71,6 +71,9 @@ def init_env():
 
 
 def setup_package():
+    jar_file = 'lib/analytics-zoo-'+VERSION+'.jar'
+    if ".dev0" in jar_file:
+        jar_file = jar_file.replace(".dev0", "-SNAPSHOT")
     metadata = dict(
         name='analytics-zoo',
         version=VERSION,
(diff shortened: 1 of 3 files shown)


jason-dai:
This is only for pip, not maven or zip download, yes?

hkvision:
Yes, only for pip. For all other downloads, I change nothing. I don't change make_dist.sh. All the jars will be created as normal.
What I do is just the following:
For pip install, I now want to only include `the jar without dependencies` (under the circumstance that all the dependencies in analytics-zoo are covered in BigDL) or the jar `with dependencies but without bigdl` (in case there is some dependency in analytics-zoo but not in BigDL) and require dependency `bigd==0.5.0`.
Not sure it is OK for you @jason-dai

jason-dai:
Sounds good to me

hkvision:
@jason-dai OK. Thanks for the comment. I will then use this way. We may need some interns to do more test for the verification.

hkvision:
Verified by me and @simanw304 if use the original way (include bigdl in the package); if users have already installed bigdl==0.5.0, it will be OK. If users have already installed bigdl==0.4.0, the following error will be throw when creating a Sequential:
```
[‎5/‎17/‎2018 11:08 AM] Wang, Siman: 
No Title
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/simanw/MyPython2.7/lib/python2.7/site-packages/zoo/pipeline/api/keras/models.py", line 40, in __init__
    super(Sequential, self).__init__(jvalue, **kwargs)
  File "/home/simanw/MyPython2.7/lib/python2.7/site-packages/bigdl/nn/keras/layer.py", line 76, in __init__
    super(KerasCreator, self).__init__(jvalue, bigdl_type, *args)
  File "/home/simanw/MyPython2.7/lib/python2.7/site-packages/bigdl/util/common.py", line 108, in __init__
    bigdl_type, self.jvm_class_constructor(), *args)
  File "/home/simanw/MyPython2.7/lib/python2.7/site-packages/bigdl/util/common.py", line 588, in callBigDlFunc
    raise e
py4j.protocol.Py4JJavaError: An error occurred while calling o30.createZooKerasSequential.
: java.lang.NoSuchMethodError: com.intel.analytics.bigdl.nn.keras.KerasLayer.inputShapeValue_$eq(Lcom/intel/analytics/bigdl/utils/Shape;)V
        at com.intel.analytics.bigdl.nn.keras.KerasLayer.<init>(KerasLayer.scala:168)
        at com.intel.analytics.zoo.pipeline.api.keras.models.KerasNet.<init>(Topology.scala:44)
        at com.intel.analytics.zoo.pipeline.api.keras.models.Sequential.<init>(Topology.scala:447)
        at com.intel.analytics.zoo.pipeline.api.keras.models.Sequential$.apply(Topology.scala:558)
        at com.intel.analytics.zoo.pipeline.api.keras.python.PythonZooKeras.createZooKerasSequential(PythonZooKeras.scala:61)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:497)
        at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
        at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:357)
        at py4j.Gateway.invoke(Gateway.java:280)
        at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
        at py4j.commands.CallCommand.execute(CallCommand.java:79)
        at py4j.GatewayConnection.run(GatewayConnection.java:214)
        at java.lang.Thread.run(Thread.java:745) 
```

hkvision:
Added a profile.
`make-dish.sh -P without_bigdl` will produce the jar `analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar` that removes bigdl and all its dependencies. Currently the size is only around 2M.
Verified on adding another dependency in pom
```
<dependency>
      <groupId>org.antlr</groupId>
      <artifactId>antlr-runtime</artifactId>
      <version>3.5</version>
</dependency>
```
After running with profile `without_bigdl`, this will be included in the final jar.

For normal use rather than pip install, `make_dist.sh -P spark_2.x`, nothing will be affected.
cc @jason-dai @zhichao-li

pyzoo/setup.py
```diff
@@ -71,6 +71,9 @@ def init_env():
 
 
 def setup_package():
+    jar_file = 'lib/analytics-zoo-'+VERSION+'.jar'
+    if ".dev0" in jar_file:
```
zhichao-li: what's purpose for such mapping?

zhichao-li:
The profile changes looks good to me

pyzoo/setup.py
hkvision: This is previously for get the jar without dependencies. Removed.

hkvision:
The current implementation is workable.

hkvision:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/394/

hkvision:
If no problem, I will merge it first for nightly build and further test.

zhichao-li:
LGTM

PR #198, opened 2018-05-16
Fix a typo in pip release twine whl name

hkvision:


hkvision:
https://github.com/pypa/pip/issues/1150 Here I saw someone says 
```I think the replacing of - with _ is part of Wheel.
PEP440 does not allow a dash```
Seems if I set the name to `analytics-zoo`, the whl file will start with `analytics_zoo` and if I set to `analytics-zoo-bigdl`, the file will start with `analytics_zoo_bigdl`.
I merge it first for test.
Can use another PR to change to name if there's a solution.

PR #200, opened 2018-05-16
Fix recommender for Python 3

hkvision:
This is actually not a problem for recommender but for all python warps that involve `numpy.int64` will have this error http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Python-AppTests/46/consoleFull 

This PR fixes this for recommender and corresponding APIs; namely cast input to int.
So that Python notebooks for recommenders can pass jenkins.

https://github.com/intel-analytics/zoo/issues/222


jason-dai:
LGTM

hkvision:
jenkins passed http://172.168.2.101:8080/job/ZOO-PR-Validation/375/

hkvision:
wait for colleagues if any can solve this problem rather than doing the cast

PR #202, opened 2018-05-16
Explicitly claim imported library

dding3:
Currently in our code, there is "from bigdl.util.common import *". It may overwrite the class defined in zoo(eg, Brightness) when loading. It would be better if we can explictly declare what's needed from BigDL,

### DIFF
--- a/pyzoo/zoo/models/image/common/image_config.py
+++ b/pyzoo/zoo/models/image/common/image_config.py
@@ -17,8 +17,6 @@
 import sys
 from bigdl.util.common import JavaValue
 from bigdl.util.common import callBigDlFunc
-from bigdl.util.common import *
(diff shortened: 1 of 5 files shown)


jenniew:
LGTM

dding3:
Thanks for your review.

Jekins pass http://172.168.2.101:8080/job/ZOO-PR-Validation/378/

jason-dai:
Maybe we should rename all image preprocessing to ImageXXX to prevent such errors, e.g., ImageCenterCrop, etc.

dding3:
Sounds good. I will submit another PR to update it.

hhbyyh:
There's an issue tracking this. https://github.com/intel-analytics/zoo/issues/232

PR #203, opened 2018-05-17
refactor the app-tests and fix wrong ftp url

glorysdj:
passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Python-AppTests/48/

This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/203)

### DIFF
--- a/apps/run-app-tests.sh
+++ b/apps/run-app-tests.sh
@@ -11,9 +11,9 @@ export ANALYTICS_ZOO_CONF=${ANALYTICS_ZOO_HOME_DIST}/conf/spark-bigdl.conf
 export PYTHONPATH=${ANALYTICS_ZOO_PYZIP}:$PYTHONPATH
 
 chmod +x ./apps/ipynb2py.sh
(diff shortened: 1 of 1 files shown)


apps/run-app-tests.sh
```diff
@@ -52,7 +51,7 @@ if [ -f "$FILENAME" ]
 then
     echo "$FILENAME already exists" 
 else
-    wget $FTP_URI/analytics-zoo-data/data/apps/object-detection/train_dog.mp4 -P $ANALYTICS_ZOO_HOME/apps/object-detection/
+    wget $FTP_URI/analytics-zoo-data/apps/object-detection/train_dog.mp4 -P $ANALYTICS_ZOO_HOME/apps/object-detection/
```
hkvision: Wondering why for the wrong ftp url these apps can still pass jenkins?

glorysdj: because if downloading from ftp failed it will download from s3

PR #208, opened 2018-05-17
Enable math operation for model construction

zhichao-li:
Previously, we only return `node`  without any meaningful math operation. In this PR, Variable would be introduced which would enable us to write some mathematical combination freely.
To some extend, it can be seems as an alternative of  `LambdaLayer`.
I'm going to merge this once the 0.1 branch created as more unitest still need to be added in.

```
        input = Input(shape=[2, 20])
        time = TimeDistributed(layer=Dense(30))(input)
        t1 = time.index_select(1, 0)
        t2 = time.index_select(1, 1)
        diff = auto.abs(t1 - t2)
        model = Model(input, diff)
        data = np.random.uniform(0, 1, [10, 2, 20])
        output = model.forward(data)
        print(output.shape)
```


jason-dai:
Not sure if we can add `Net()` for scala

zhichao-li:
fix https://github.com/intel-analytics/zoo/issues/285

PR #213, opened 2018-05-18
change `FeatureTransformer` used in Zoo to `Preprocessing`

dding3:
What changes were proposed in this pull request?
1. update imageset from tranform `FeatureTransformer` to `Preprocessing`
2. Change pre & post processors in ImageConfigure from FeatureTransformer to Preprocessing
3. add `BytesToMat`, `MatToFloats`, `ImageSetToSample`

How was this patch tested?
unit test and manual test

Related links or issues (optional)
https://github.com/intel-analytics/zoo/issues/209
https://github.com/intel-analytics/zoo/issues/237

### DIFF
--- a/pyzoo/zoo/models/image/common/image_config.py
+++ b/pyzoo/zoo/models/image/common/image_config.py
@@ -41,11 +41,11 @@ class ImageConfigure(JavaValue):
             self.value = jvalue
         else:
             if pre_processor:
-                assert pre_processor.__class__.__bases__[0].__name__ == "FeatureTransformer", \
-                    "the pre_processor should be subclass of FeatureTransformer"
+                assert pre_processor.__class__.__bases__[0].__name__ == "Preprocessing", \
+                    "the pre_processor should be subclass of Preprocessing"
             if post_processor:
-                assert post_processor.__class__.__bases__[0].__name__ == "FeatureTransformer", \
-                    "the pre_processor should be subclass of FeatureTransformer"
+                assert post_processor.__class__.__bases__[0].__name__ == "Preprocessing", \
+                    "the pre_processor should be subclass of Preprocessing"
             self.value = callBigDlFunc(
                 bigdl_type, JavaValue.jvm_class_constructor(self),
                 pre_processor,
--- a/zoo/src/main/scala/com/intel/analytics/zoo/feature/common/Preprocessing.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/feature/common/Preprocessing.scala
@@ -60,7 +60,12 @@ class ChainedPreprocessing[A, B, C](first: Preprocessing[A, B], last: Preprocess
   }
 
   def apply(imageSet: ImageSet): ImageSet = {
-    imageSet.transform(this.asInstanceOf[Preprocessing[ImageFeature, ImageFeature]])
+    if (this.isInstanceOf[Preprocessing[ImageFeature, ImageFeature]]) {
+      imageSet.transform(this.asInstanceOf[Preprocessing[ImageFeature, ImageFeature]])
+    } else {
+      throw new UnsupportedOperationException("imageSet only supports transform" +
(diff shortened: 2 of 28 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/feature/common/Preprocessing.scala
```diff
@@ -58,12 +58,23 @@ class ChainedPreprocessing[A, B, C](first: Preprocessing[A, B], last: Preprocess
   override def apply(prev: Iterator[A]): Iterator[C] = {
     last(first(prev))
   }
+
+  def apply(imageSet: ImageSet): ImageSet = {
+    imageSet.transform(this.asInstanceOf[Preprocessing[ImageFeature, ImageFeature]])
```
hhbyyh: This will fail if Preprocessing is of different types. Can we put it into ImageProcessing?

jason-dai: This should be Preprocessing[ImageFeature, _], yes?

Maybe we can check it and report an error if needed.

zoo/src/main/scala/com/intel/analytics/zoo/feature/image/BytesToMat.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import com.intel.analytics.bigdl.transform.vision.image.ImageFeature
+import com.intel.analytics.zoo.feature.common.{ImageProcessing}
+import com.intel.analytics.bigdl.transform.vision.image.{BytesToMat => BBytesToMat}
+
+/**
+ * Transform byte array(original image file in byte) to OpenCVMat
+ * @param byteKey key that maps byte array
+ */
+class BytesToMat(byteKey: String = ImageFeature.bytes) extends ImageProcessing {
```
jason-dai: Change to ImageBytesToMat

zoo/src/main/scala/com/intel/analytics/zoo/feature/common/Preprocessing.scala
dding3: There is `apply` in ImageProcessing and we put another `apply` in `ChainPreprocessing` to
1. make feature parity with python code.
2. usually this function will be called only when user calling `chainPreprocessing(imageset)`, Compared to `imageset -> chainPreprocessing`, the latter way should be more popular.

As `ImageSet` only support transform `Preprocessing[ImageFeature, ImageFeature]`, we need casting `ChainPreprocessing` here and have updated the code to throw exception if Preprocessing with different types.

hhbyyh: In that case
1. are we sure we need to provide the apply interface for ImageSet specifically? Since there's already another set of API that can reach the same target. (ImageSet ->)
2. If still we want to add the apply method, I think we can put it into `Preprocessing` directly.

jason-dai: Let's just move this to Preprocessing.

jason-dai:
Is this ready for merge?

pyzoo/zoo/models/image/common/image_config.py
```diff
@@ -41,11 +41,11 @@ def __init__(self, pre_processor=None,
             self.value = jvalue
         else:
             if pre_processor:
-                assert pre_processor.__class__.__bases__[0].__name__ == "FeatureTransformer", \
-                    "the pre_processor should be subclass of FeatureTransformer"
+                assert pre_processor.__class__.__bases__[0].__name__ == "Preprocessing", \
```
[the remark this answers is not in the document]
dding3: There is a chance that pre_processor is ChainPreprocessing

dding3:
Jekins pass. http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/417/

PR #232, opened 2018-05-21
fix conflict in pom.xml

simanw304:
fix conflict in pom.xml
add unit tests for some operators


zhichao-li:
let's merge this once pass the PR test.

PR #234, opened 2018-05-21
modify doc structure 

helenlly:
modify doc structure and delete old docs


zhichao-li:
@helenlly We should use "Squash and Merge" to squash the commits

PR #235, opened 2018-05-21
Add more metrics

zhichao-li:


zhichao-li:
cc @dding3 Please check if the default value for AUC is OK.

zhichao-li:
jenkins passed.

pyzoo/zoo/pipeline/api/keras/utils.py
```diff
@@ -66,12 +66,22 @@ def to_bigdl_criterion(criterion):
...
 
+def to_bigdl_metric(metric):
+    from zoo.pipeline.api.keras.metrics.auc import AUC
+    metric = metric.lower()
+    if metric == "accuracy":
+        return Top1Accuracy()
+    elif metric == "mae":
+        return MAE()
+    elif metric == "auc":
+        return AUC(1000)
```
dding3: I think 1000 is too large. Can we pass a parameter here and let user set the default value? As we allow user set this value when they create AUC, it may be better if to_bigdl_metric have the same behavior.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/utils/KerasUtils.scala
```diff
@@ -169,12 +170,17 @@ object KerasUtils {
...
-      List(new Top1Accuracy[T]())
-    }
-    else {
-      throw new IllegalArgumentException(s"Unsupported metrics: ${metrics.mkString(", ")}")
+    } else {
+      metrics.map { metric =>
+        metric.toLowerCase() match {
+          case "accuracy" => new Top1Accuracy[T]()
+          case "mae" => new MAE[T]()
+          case "auc" => new AUC[T](1000)
```
dding3: Same as above

pyzoo/zoo/pipeline/api/keras/utils.py
jason-dai: you can only pass a string for now; what is a good default value?

dding3: I noticed in tensorflow it's using 200. I think we can refer this value.

zhichao-li: I would give AUC a default value 200. For metrics, you can pass in a list of metrics object if want to specify the parameters.

zhichao-li: https://github.com/intel-analytics/analytics-zoo/pull/238

PR #241, opened 2018-05-22
Check environment variable/sys.path when adding

hkvision:
Before appending to environment variables, check if it already exists.
Fix this in Analytics Zoo. But as BigDL also has such code, print "Adding ... to ..." multiple times is still unavoidable unless we fix in BigDL.

As far as I test, now even if the message is printed multiple times, `BIGDL_JARS` only has one jar as we desire. Without the check, there will be two duplicate jars.

### DIFF
--- a/pyzoo/zoo/util/engine.py
+++ b/pyzoo/zoo/util/engine.py
@@ -69,8 +69,9 @@ def __prepare_analytics_zoo_env():
 
     def append_path(env_var_name, path):
         try:
-            print("Adding %s to %s" % (path, env_var_name))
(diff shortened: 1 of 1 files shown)


hkvision:
Related issue: https://github.com/intel-analytics/zoo/issues/239

zhichao-li:
We can accept this as a workaround for now, but I think still we need to figure out why the appending happen multiple times.

hkvision:
jenkins passed http://172.168.2.101:8080/job/ZOO-PR-Validation/457/
Merge it first as it could at least aviod adding multiple times in Analytics Zoo.

PR #243, opened 2018-05-22
renamed faraudDetection notebook

xuex2017:


xuex2017:
related issue: intel-analytics/zoo#261

PR #245, opened 2018-05-22
Fix getParameterTable() for Keras layers and models

hkvision:
Fix https://github.com/intel-analytics/zoo/issues/22


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/Net.scala
```diff
@@ -34,6 +34,27 @@ import scala.reflect.ClassTag
  * A placeholder to add layer's utilities
  */
 trait Net {
+  def getParametersTable[T: ClassTag](): Table = {
+    val module = this.asInstanceOf[AbstractModule[Activity, Activity, T]]
+    val params = module.parameters()
```
dding3: Can we use "module.getParametersTable" here? Looks like the implementation is the same.

hkvision: Yeah. Fixed not pushed the commit yet...

PR #249, opened 2018-05-22
support preprocessing when reading image

dding3:
Fix https://github.com/intel-analytics/zoo/issues/257


jason-dai:
Does it solve the memory issue?
1) ImageResize does not remove the original bytes
2) ImagePreprocessing will add Mat, which can be large size?

dding3:
Close this PR since it doesn't work.

PR #251, opened 2018-05-23
add summary for zoo model

dding3:


zoo/src/main/scala/com/intel/analytics/zoo/models/common/ZooModel.scala
```diff
@@ -88,6 +91,21 @@ abstract class ZooModel[A <: Activity: ClassTag, B <: Activity: ClassTag, T: Cla
...
+   * @return HashMap(layername/parametername -> sizes)
+   */
+  def summary(): HashMap[String, Array[Int]] = {
+    val variables = new HashMap[String, Array[Int]]()
+    model.getParametersTable().foreach { case (name: String, params: Table) =>
+      params.foreach { case (paramName: String, param: Tensor[Float]) =>
+        variables += (name + "/" + paramName -> param.size())
+      }
+    }
+    variables
```
hkvision: Can you give an example what will print summary() give?

dding3: ObjectDetectorSpec has a ut which will print summary()

hkvision: I saw it. I mean can you paste the print result here so that I can have a look what it looks like without running the ut?

dding3: Ok. For the ut, it will print sth like:
conv6/dw/weight 256X1X1X3X3
conv10/weight 1X512X512X1X1
conv1/dw/weight 32X1X1X3X3
conv6/gradWeight 1X512X256X1X1
conv17_1/gradBias 64
conv2/dw/gradBias 64

hkvision: Thanks for the update. cc @jason-dai

dding3:
duplicated pr.

PR #274, opened 2018-05-25
NNEstimator: Add option to handle invalid data during fit and transform

hhbyyh:
Add option `handleInvalid` to process faulty data.

This is a prototype and does not target for the first release.


hhbyyh:
cc @jenniew

zoo/src/main/scala/com/intel/analytics/zoo/feature/common/FeatureLabelPreprocessing.scala
```diff
@@ -37,13 +37,19 @@ class FeatureLabelPreprocessing[F, L, T: ClassTag](
...
+          case Some(l) =>
+            val labelTensor = labelStep(Iterator(l)).next()
+            Sample[T](featureTensor, labelTensor)
+          case None =>
+            Sample[T](featureTensor)
+        }
+      } catch {
+        case e: Exception =>
+          println(e.getMessage)
+          null
```
jenniew: If we choose to set handleInvalid as error, I think we need to throw the original exception not the NullPointerException which is confusing to the user.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/nnframes/NNEstimator.scala
```diff
@@ -538,8 +577,11 @@ class NNModel[T: ClassTag] private[zoo] (
               "unexpected batchResult dimension: " + batchResult.size().mkString(", "))
           }
         }
-        rowBatch.toIterator.zip(predictions).map { case (row, predict) =>
+
+        validRow.map(_._1).iterator.zip(predictions).map { case (row, predict) =>
           Row.fromSeq(row.toSeq ++ Seq(predict))
+        } ++ invalidRows.map(_._1).map { row =>
+          Row.fromSeq(row.toSeq ++ Seq(Array.empty))
```
hhbyyh: Row.fromSeq(row.toSeq ++ Seq(outputToPrediction(Tensor(0))))

And update in NNClassifier.scala:
  protected override def outputToPrediction(output: Tensor[T]): Any = {
    if (output.isSameSizeAs(Tensor(0))) {
      Double.NaN
    } else {
      ev.toType[Double](output.max(1)._2.valueAt(1))
    }
  }

zoo/src/main/scala/com/intel/analytics/zoo/feature/common/FeatureLabelPreprocessing.scala
jason-dai: Throw a meaningful exception

hhbyyh: `null` here is the place holder for the invalid data.

jason-dai:
Why don't we simply to implement a Preprocessing to filer invalid images, which can be reused across NNEstimator, NNModel, etc.?

hhbyyh:
I was trying to support the filtering, not only for image data, but for other data types during preprocessing as well.
I do feel this is not in the best state as it's requesting all the Preprocessing to return null for invalid records and not-null for valid records, but we didn't enforce the design in Preprocessing.

hhbyyh:
Options are:
        'drop': invalid data are ignored during training (fit).
                During prediction (transform), rows containing invalid data will be dropped.
        'error': throw an error whenever an invalid data is met.

PR #275, opened 2018-05-25
programing docs for recommendation and anomaly detection

songhappy:


docs/docs/ProgrammingGuide/anomaly-detection.md
```diff
@@ -1 +1,27 @@
+# Analytics Zoo Anomaly detection
 
+Analytics Zoo shows how to detect anomalies in time series data based on RNN network. Currently, a [python notebook](https://github.com/intel-analytics/analytics-zoo/tree/master/apps/anomaly-detection) is provided. 
+In the example, a RNN network using Analytics Zoo Keras-Style API is built, and [NYC taxi passengers dataset](raw.githubusercontent.com/numenta/NAB/master/data/realKnownCause/nyc_taxi.csv) is used to train and test the model.
+
+Build a RNN model for anomaly detection. There are three LSTM layers followed by one Dense layer at the end.
```
shane-huang: The flow is okay while we need to make a few things clearer. Specifically,
1. The way we use RNN for anomaly detection is not the only way RNN can be used for this task, so we better specify the model is used to `predict the future values` instead of just `for anomaly detection`.  
1. Need to tell user the training data is considered all normal

Thus I suggest we modify the text like below: 
We split the entire curve into 2 sections - training data and testing data. The training data section is treated all as normal and an RNN model is trained to fit the training data. The models has three LSTM layers followed by one Dense layer at the end.
<...model code>

Then we train the model using MSE loss (as a regression problem).
<...train code>
 
Then we use this RNN model to predict the testing curve. 
<...predict code>

 Anomalies could be defined by comparing the predictions and actual data values. The current example defines data points as anomalies if the difference of predictions and actual values are larger than a certain value.
...

helenlly:
GQ, we will merge this PR first to build doc website. After you modify as Shane's comments, pls open a new PR , thanks.

PR #104, opened 2018-02-01
update fraud

hhbyyh:
increase learning rate to 3e-2 to increase stability.

update to BigDL 0.4

removed Bagging and directly put the code into notebook.

PR #286, opened 2018-05-29
Change fraud detection app API to zoo

xiaowang1105:
still need to tune model.
By using bagging model, I didn't increase the AUPRC.


jason-dai:
Just compare with the previous result; if they are similar, we can just merge it.

xiaowang1105:
@shane-huang , Actually, by repeating the experiments, I get multiple results (AUPRC = 0.49, AUPRC after bagging 0.49, AUPRC = 0.58, AUPRC after bagging = 0.90), so I think the problem may be trival not related to the model or estimator. Finally, I choose a decent result and re-push it.
The problem related to spark vector error is stated below:
```
Name: org.apache.spark.SparkException
Message: Job aborted due to stage failure: Task 0 in stage 21.0 failed 1 times, most recent failure: Lost task 0.0 in stage 21.0 (TID 218, localhost, executor driver): java.lang.IllegalArgumentException: SeqToTensor only supports Float, Double, Array[Float], Array[Double] or MLlib Vector but got [-0.6942411021638849,-0.04407484719355218,1.6727705625425275,0.9733638055579641,-0.24511615322163896,0.3470673358576231,0.19367859829946352,0.08263713433481461,0.3311272018898202,0.08338539885285959,-0.5404060870051831,-0.6182946323291876,-0.9960971732491746,-0.32460961644902997,1.6040110229374405,-0.5368319260681538,0.24486302414440858,0.030769878583235922,0.4962811553900588,0.32611744349136557,-0.02492332120662103,0.3828537662094823,-0.1769110237561329,0.1105067265574083,0.24658501006984637,-0.39216974306341235,0.33089104174320144,-0.06378103872482441,0.5981926061623525]
```

jason-dai:
It says "SeqToTensor only supports Float, Double, Array[Float], Array[Double] or MLlib Vector", so MLlib Vector is supported? @hhbyyh

shane-huang:
Seems Keras scala model does not support vector. @jason-dai

hhbyyh:
@jason-dai SeqToTensor does not support ML Vector (in Spark 2.0+), since our code need to be compatible with Spark 1.6. I created the issue https://github.com/intel-analytics/zoo/issues/140 and IMO we should include that in the next release.

hhbyyh:
@xiaowang1105 Thanks for working on this. 
We definitely need stable result for the notebook. I used to increase the learning rate to stabilize the result. https://github.com/intel-analytics/analytics-zoo/pull/104/files

xiaowang1105:
@hhbyyh , I have set the learning rate to 3e-2, 3.5e-2, 4e-2, but these parameters still cause some variation in Area under precision-recall curve. Among them, 3e-2 is the best....

hhbyyh:
what's the range of the variation?

xiaowang1105:
single model AUPRC ranging from 0.5 to 0.7
bagging model ranging from 0.5 to 0.7 @hhbyyh

hhbyyh:
Without the change in this PR, are you able to reproduce the result (93% AUPRC) with the master notebook or with the one in legacy branch?
https://github.com/intel-analytics/analytics-zoo/blob/legacy/apps/fraudDetection/Fraud%20Detction.ipynb

xiaowang1105:
The best result I ever got without changing is 0.85 (indicated in this pr) @hhbyyh

apps/fraud-detection/fraud-detection.ipynb
```diff
@@ -101,7 +106,7 @@
     "import spark.implicits._\n",
     "Engine.init\n",
```
jason-dai: Instead of `Engine.init`, use `initNNContext` instead

PR #292, opened 2018-05-30
modify overviewpage format

helenlly:
Fix: https://github.com/intel-analytics/zoo/issues/280


zhichao-li:
Please use `Squashed and merge` instead of `merge` otherwise the commits history would be polluted. cc @helenlly  @jason-dai

PR #297, opened 2018-05-30
add api for inference model

glorysdj:
this pr add simple pojo style inference model to analytics-zoo. 
java user can extends the AbstractInferenceModel and predict with java input and get java output

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/297)

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2018 Analytics Zoo Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.intel.analytics.zoo.pipeline.inference
 
 import com.intel.analytics.bigdl.optim.LocalPredictor
@@ -8,7 +24,7 @@ import com.intel.analytics.bigdl.tensor.Tensor
 import scala.collection.JavaConverters._
 
 case class FloatInferenceModel(
-  model:     AbstractModule[Activity, Activity, Float],
+  model: AbstractModule[Activity, Activity, Float],
   predictor: LocalPredictor[Float]) extends InferenceSupportive {
 
   def forward(data: Tensor[Float])(implicit om: Manifest[Float]): Array[Float] = {
@@ -19,7 +35,8 @@ case class FloatInferenceModel(
     predictor.predictClass(samples)
   }
 
-  def predict(input: java.util.List[java.util.List[java.lang.Float]]): java.util.List[java.lang.Float] = {
+  def predict(input: java.util.List[java.util.List[java.lang.Float]]):
+    java.util.List[java.lang.Float] = {
     timing(s"predict for $input") {
       val tensor = transferInferenceInputToTensor(input)
       val result = model.forward(tensor).asInstanceOf[Tensor[Float]].storage.array.toList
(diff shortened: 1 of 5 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.intel.analytics.bigdl.nn.abstractnn.Activity
+import com.intel.analytics.bigdl.dataset.Sample
+import com.intel.analytics.bigdl.tensor.Tensor
+import scala.collection.JavaConverters._
+
+case class FloatInferenceModel(
+  model: AbstractModule[Activity, Activity, Float],
+  predictor: LocalPredictor[Float]) extends InferenceSupportive {
+
+  def forward(data: Tensor[Float])(implicit om: Manifest[Float]): Array[Float] = {
```
jason-dai: do we need this?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
```diff
@@ -0,0 +1,47 @@
...
+    model.forward(data).asInstanceOf[Tensor[Float]].storage().array()
+  }
+
+  def predict(samples: Array[Sample[Float]]): Array[Int] = {
```
jason-dai: do we need this?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
```diff
@@ -0,0 +1,47 @@
...
+    predictor.predictClass(samples)
```
jason-dai: why predictClass?

jason-dai:
Is this thread safe?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
glorysdj: for scala user?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
glorysdj: I will delete these two functions for now. we can add more apis for both java and scala users in the future., imo.

glorysdj:
For thread safe, @zhichao-li please help to review transferInferenceInputToTensor method which is used to transfer java List to bigdl Tensor. I'm not sure of this part.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModelFactory.scala
```diff
@@ -0,0 +1,34 @@
...
+
+object InferenceModelFactory{
+  def loadFloatInferenceModel(modelPath: String): FloatInferenceModel = {
+    loadFloatInferenceModel(modelPath, null)
+  }
+
+  def loadFloatInferenceModel(modelPath: String, weightPath: String): FloatInferenceModel = {
+    val model = ModelLoader.loadFloatModel(modelPath, weightPath)
+    val predictor = LocalPredictor(model)
+    model.evaluate()
```
zhichao-li: remove this line ? we should suppose `LocalPredictor` has already done such job.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -0,0 +1,58 @@
...
+    val begin = System.currentTimeMillis
+    val result = f
+    val end = System.currentTimeMillis
+    val cost = (end - begin)
+    logger.info(s"$name time elapsed [${cost / 1000} s, ${cost % 1000} ms].")
+    result
+  }
+
+  def transferInferenceInputToTensor(input: java.util.List[java.util.List[java.lang.Float]]):
+    Tensor[Float] = {
```
zhichao-li: should we add another parameter `shape` here?
and also we can simplified the logic like this:
1) convert input to an `Array` , you can reuse the logic below
```
while (d < arrays.length) {
+          System.arraycopy(arrays(d), 0, _buffer.storage().array(), d * size, size)
+          d += 1
+        }
2. create a tensor with `Tensor(data=xxArray, shape=xxshape)`
```

glorysdj: sure, we should add such api in the future.

PR #309, opened 2018-05-31
Fix Net.load_keras()

hkvision:
A bug brought by https://github.com/intel-analytics/analytics-zoo/pull/132
Need to add `return`, otherwise the return result is None.


hkvision:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/525/

PR #330, opened 2018-06-04
Remove edges coming output output in newGraph()

yangw1234:
Remove edges coming output output in newGraph()
Fix transfer learning example bug


yangw1234:
pr validation fail on two unrelated cases, merge it for now

PR #341, opened 2018-06-06
Fix image augmentation notebook output image format

yangw1234:
Just re-run the notebook


yangw1234:
pr validation pass, and manually verified on github web page

PR #359, opened 2018-06-07
Add transfer learning notebook

xuex2017:


jason-dai:
1) Change the folder name to "dogs-vs-casts"
2) Change occurrence of "Zoo" to "Analytics Zoo"
3) Do not show the path specific to the user (e.g., "/home/xiaxu...")
4) Visualize some pictures for the final predictions (with both dogs and cats)

xuex2017:
which folder's name should be changed to "dogs-vs-cat"?

xuex2017:
Do you mean the "transfer-learning" folder be changed to "dogs-vs-cat"?

jason-dai:
Yes, we should change the folder name.

PR #363, opened 2018-06-07
Add util to Export tensorflow graph

yangw1234:
## Util to Export tensorflow graph

The export_tf utility function will
 1. freeze the graph (replace all variables with constants)
 2. strip all unused node as specified by inputs and outpus
 3. add placeholder nodes as needed 
 4. write the frozen graph and inputs/outputs names to the folder

### Usage
```python
export(sess, folder, list_of_input_tensors, list_of_output_tensors)
```
loading from python
```python
net = TFNet.from_export_folder(folder)
```
loading from scala
```scala
val net = TFNet(folder)
```

### DIFF
--- a/pyzoo/zoo/util/tf.py
+++ b/pyzoo/zoo/util/tf.py
@@ -1,0 +1,155 @@
+#
+# Copyright 2018 Analytics Zoo Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


pyzoo/zoo/util/tf.py
```diff
@@ -0,0 +1,163 @@
...
+
+    if not os.path.isdir(folder):
+        os.mkdir(folder)
+
+    with gfile.GFile(os.path.join(folder, "frozen_inference_graph.pb"), "wb") as f:
+        f.write(optimized_graph_def.SerializeToString())
+
+    meta = {
+        "input_names": new_input_names,
+        "output_names": output_names
```
jason-dai: We need to guarantee that `input_names` and `output_names` match exactly the `inputs` and `outputs` tensors specified by the user?

yangw1234:
pr validation pass

PR #373, opened 2018-06-08
Add glove for Embedding layer; update Scala TextClassification example

hkvision:
Initial workable version for review. Needs further discussion and modification. cc @jason-dai 

New layer EmbeddingGloVe.
Still refactoring the code.


hkvision:
This PR is too old... I will reopen a new one

PR #399, opened 2018-06-13
Fix Print info

xiaowang1105:
Even though the code could run flawlessly, the error info would confuse users when no data found.


hkvision:
cc @zhichao-li

PR #405, opened 2018-06-13
Path bugs fixed

wngzhiqi:
Fix bugs due to rename.


jason-dai:
Just replace the previous notebooks, instead of adding new ones.

PR #410, opened 2018-06-14
TFNet should be able to process tensor whose storage.length is not equal to its length

yangw1234:
TFNet should be able to process tensor whose storage.length is not equal to its length


yangw1234:
pr validation pass

PR #414, opened 2018-06-15
add script to start scala notebook

xuex2017:


scripts/jupyter-with-zoo-scala.sh
```diff
@@ -0,0 +1,30 @@
...
+    echo "Please set SPARK_HOME environment variable"
+    exit 1
+fi
+
+# setup paths
+export ANALYTICS_ZOO_JAR=`find ${ANALYTICS_ZOO_HOME}/lib -type f -name "analytics-zoo*jar-with-dependencies.jar"`
+export ANALYTICS_ZOO_CONF=${ANALYTICS_ZOO_HOME}/conf/spark-analytics-zoo.conf
+
+# Check files
+if [ ! -f ${ANALYTICS_ZOO_CONF} ]; then
```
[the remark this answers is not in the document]
xuex2017: no. we can delete these lines

scripts/jupyter-with-zoo-scala.sh
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+fi
+
+if [ -z "${SPARK_HOME}" ]; then
+    echo "Please set SPARK_HOME environment variable"
+    exit 1
+fi
+
+# setup paths
+export ANALYTICS_ZOO_JAR=`find ${ANALYTICS_ZOO_HOME}/lib -type f -name "analytics-zoo*jar-with-dependencies.jar"`
```
hkvision: Wondering about this: as you define ANALYTICS_ZOO_JAR in this script, and you use ANALYTICS_ZOO_JAR in $1, will this cause a recursion or deadlock?...

xuex2017: I have tested this script and the notebook works fine.

xuex2017: I've tested the script and the notebook works fine. shouldn't be a problem..

hkvision: Emmm...even though it works fine, won't it confuse the users? as they don't even know or realize what ANALYTICS_ZOO_JAR is... Pending this for discussion with other colleagues...

hkvision:
Maybe can you try the following?
put `${ANALYTICS_ZOO_JAR}` inside `jupyter-with-zoo-scala.sh` only and let users specify other options passing as parameters?

xuex2017:
updated jupyter-with-zoo-scala.sh

hkvision:
Seems that you are including other's commits in this PR? See files changed...

xuex2017:
fixed

PR #420, opened 2018-06-19
Add yTrueShape for CustomLoss

zhichao-li:
It would be useful if yPred and yTrue have different shape.


zhichao-li:
cc @hhbyyh

zhichao-li:
jenkisn passed. ZOO-PR-Validation/649/

PR #428, opened 2018-06-21
Remove build dependency of spark-bigdl.conf

dding3:
Fix https://github.com/intel-analytics/zoo/issues/312


hkvision:
Wondering if we remove bigdl in make-dist, then will the python zip file contain the bigdl packages?

dding3:
Close the PR as it has been fixed in another one.

PR #431, opened 2018-06-21
add caffe model loading support to AbstractInferenceModel

glorysdj:
With this support, we can use AbstractInferenceModel with Caffe models.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/431)

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
@@ -30,31 +30,7 @@ case class FloatInferenceModel(
   model: AbstractModule[Activity, Activity, Float],
   predictor: LocalPredictor[Float]) extends InferenceSupportive {
 
-  def predict(input: JTensor): JList[JTensor] = {
-    timing("model predict") {
-      val inputData = input.getData
-      val inputShape = input.getShape
-      val sample = transferInputToSample(inputData,
-        inputShape.asScala.toArray.map(_.asInstanceOf[Int]))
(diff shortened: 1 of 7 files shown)


glorysdj:
passed pr validation:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/653/

jason-dai:
Instead of adding different types, you can just add a loadCaffe method

glorysdj:
I'm afraid in the future, there will be more types, loadXXX methods will be a mess accordingly, and also messy reloadXXX methods.

glorysdj:
Ok, I will change to loadCaffe, and will also add JTensor, and refactor predict methods.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModel.scala
```diff
@@ -30,12 +30,55 @@ case class FloatInferenceModel(
   model: AbstractModule[Activity, Activity, Float],
   predictor: LocalPredictor[Float]) extends InferenceSupportive {
 
-  def predict(input: JList[JFloat], shape: JList[JInt]): JList[JFloat] = {
+  def predict(input: JTensor): JList[JTensor] = {
```
jason-dai: why do we need this? why not assume that the input is always a batch?

glorysdj: ok. I will change to only one predict method which take batch inputs.

jason-dai: I don't think we should do that as it will break backward compatibility?

PR #433, opened 2018-06-21
Net.loadTF from folder

hkvision:


hkvision:
cc @yangw1234 seems the pb used in tfnet can't be loaded as a GraphNet(BigDL model)?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/NetUtils.scala
```diff
@@ -159,8 +162,31 @@ object NetUtils {
...
+  implicit val formats = DefaultFormats
+
+  private[zoo] def processTFFolder(folder: String): (String, Seq[String], Seq[String]) = {
+    val folderPath = Path(folder)
+    if (!folderPath.exists) {
+      throw new IllegalArgumentException(s"$folder does not exist")
+    }
+
+    val modelPath = folderPath / Path("frozen_inference_graph.pb")
+    val metaPath = folderPath / Path("graph_meta.json")
```
jason-dai: check if these two files exist

pyzoo/zoo/pipeline/api/net.py
```diff
@@ -138,8 +138,7 @@ def load_torch(path, bigdl_type="float"):
         return GraphNet.from_jvalue(jmodel, bigdl_type)
 
     @staticmethod
-    def load_tf(path, inputs, outputs, byte_order="little_endian",
-                bin_file=None, bigdl_type="float"):
+    def load_tf(**kwargs):
```
jason-dai: I don't we need to change the method signature here; just make inputs and outputs default to None; then in the method, you can check if they are None, and then call different netLoadTF accordingly.

hkvision: Ar yes. Thanks for the comment. Updated.

hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/657/

PR #438, opened 2018-06-25
Support pulling python dependency from maven

zhichao-li:
With this Patch, Analytics-Zoo can pull python dependency from Maven directly and build the whole project base on the given bigdl version.
Commands:
`./make-dist.sh -Pspark_2.x -DskipTests` or specify an alternative bigdl version.
`./make-dist.sh -Pspark_2.x -Dbigdl.version=0.6.0-SNAPSHOT -DskipTests`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.class
 target/
 /dist/
+pyzoo/.pytest_cache/
 
 # SBT, Maven specific
 .cache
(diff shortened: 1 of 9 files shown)


zhichao-li:
cc @jason-dai  @glorysdj

zhichao-li:
Jenkins passed. #671

PR #450, opened 2018-06-29
Fix operator mm

zhichao-li:
Before this patch, `mm` operator would generate wrong result if being executed multiple times.


zhichao-li:
Jenkins pass. 
Merge this for now otherwise the training would be affected.

zhichao-li:
cc @simanw304  I think you would like to rebase the code after this patch.

PR #467, opened 2018-07-05
optimize preprocessing

yangw1234:
JList[JFloat] to Sample time drop from 140ms to 40ms for a single 2590 x 1942 x 3 image on a i7-6700 desktop.


yangw1234:
manually verified

PR #470, opened 2018-07-05
fix inference output tensor order mess up

yangw1234:


hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/728/

PR #484, opened 2018-07-09
add keras2 MaxPooling 2/3D layers and test

yjjiang11:


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKeras2.scala
```diff
@@ -141,5 +141,34 @@ class PythonZooKeras2[T: ClassTag](implicit ev: TensorNumeric[T]) extends Python
       toScalaShape(inputShape))
   }
 
+  def createZooKeras2MaxPooling2D(
+      poolSize: (Int, Int) = (2, 2),
```
hkvision: You should use JList[Int] to avoid the method ... not found error...
See PythonBigDLKeras createZooKerasMaxPooling2D for reference.

PR #489, opened 2018-07-10
Support predict and evaluate on Imageset

hkvision:
Fix https://github.com/intel-analytics/zoo/issues/270

Need to depend on BigDL 0.6.0 and use its `evaluateImage` API

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
@@ -44,7 +44,6 @@ import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 import scala.language.implicitConversions
 
-
 abstract class KerasNet[T: ClassTag](implicit ev: TensorNumeric[T])
   extends KerasLayer[Activity, Activity, T] with Net {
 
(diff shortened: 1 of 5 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -326,6 +333,10 @@ abstract class KerasNet[T: ClassTag](implicit ev: TensorNumeric[T])
     localPredictor.predict(x)
   }
 
+  def predict(x: ImageSet): ImageSet = {
+    ImageSet.fromImageFrame(predictImage(x.toImageFrame()))
```
hkvision: In predictImage, the argument is `batchPerPartition`, any suggestion on how to use batchSize here... seems we can't get the partition number.

dding3:
LGTM

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
dding3: I think partition number can be get by `rdd.partitions.length`

hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/848/

hkvision:
Thanks so much for your review and help!

PR #506, opened 2018-07-13
add a member variable in TFGraphHolder

chengshunzhang:
Add a member val inDriver in class TFGraphHolder in object TFNet so that we do not need to repeatedly call isDriver when on worker node. Test using ssd_mobilenet_v1_coco with no problems.


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFNet.scala
```diff
@@ -253,6 +253,9 @@ object TFNet {
   class TFGraphHolder(@transient var tfGraph: Graph, private var id: String)
     extends Serializable with KryoSerializable {
 
+    @transient
+    lazy val isDr = isDriver
```
yangw1234: change the variable name to a more meaningful one?

yangw1234:
please add more description about what does this pr do and describe your test result

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFNet.scala
```diff
@@ -254,7 +254,7 @@ object TFNet {
     extends Serializable with KryoSerializable {
 
     @transient
-    lazy val isDr = isDriver
+    lazy val theDriver = isDriver
```
yangw1234: change to inDriver?

chengshunzhang: OK.

yangw1234:
LGTM

yangw1234:
pr validation pass

PR #515, opened 2018-07-13
Support list within a list for kerasNet forward

yangw1234:
Support list within a list for kerasNet forward.

Predict doest not work for now, because it need fixes from bigdl predictor.


yangw1234:
pr validation pass

PR #524, opened 2018-07-16
Restrict pyspark version for pip install; doc update

hkvision:
Fix https://github.com/intel-analytics/zoo/issues/336
As this example will fail for spark 2.3.1.

### DIFF
--- a/pyzoo/setup.py
+++ b/pyzoo/setup.py
@@ -93,7 +93,7 @@ def setup_package():
         license='Apache License, Version 2.0',
         url='https://github.com/intel-analytics/analytics-zoo',
         packages=packages,
-        install_requires=['pyspark<2.3.0', 'bigdl==0.6.0'],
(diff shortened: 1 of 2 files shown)


pyzoo/setup.py
```diff
@@ -93,7 +93,7 @@ def setup_package():
         license='Apache License, Version 2.0',
         url='https://github.com/intel-analytics/analytics-zoo',
         packages=packages,
-        install_requires=['bigdl==0.6.0'],
+        install_requires=['pyspark<2.3.0', 'bigdl==0.6.0'],
```
jason-dai: shall we specify pyspark >2.2.0?

hkvision: updated.

hkvision:
Synced with @jason-dai offline, for this release, we restrict pyspark<2.3. In the future, after we test against spark 2.3, we can remove this constraint.

hkvision:
Issued created for spark 2.3 https://github.com/intel-analytics/zoo/issues/338

PR #539, opened 2018-07-17
split recommendation into ncf and wnd

songhappy:


apps/recommendation-ncf/README.md
```diff
@@ -14,7 +14,7 @@ These two notebooks demonstrate how to build neural network recommendation syste
 * Run `export ANALYTICS_ZOO_HOME=the dist directory under the Analytics Zoo project`.
 * Run the following bash command to start the jupyter notebook. Change parameter settings as you need, ie `MASTER = local[physcial_core_number]`.
 ```bash
-MASTER=local[*]
+MASTER=local[4]
```
shane-huang: Why we use local[4] instead of local[*]? What if a user has only 2 cores in his computer?

apps/recommendation-wide-n-deep/README.md
```diff
@@ -0,0 +1,25 @@
...
+## Steps to run the notebook
+* Download Analytics Zoo and build it.
+* Run `export SPARK_HOME=the root directory of Spark`.
+* Run `export ANALYTICS_ZOO_HOME=the dist directory under the Analytics Zoo project`.
+* Run the following bash command to start the jupyter notebook. Change parameter settings as you need, ie `MASTER = local[physcial_core_number]`.
+```bash
+MASTER=local[4]
+bash ${ANALYTICS_ZOO_HOME}/bin/jupyter-with-zoo.sh \
+    --master ${MASTER} \
+    --driver-cores 4  \
```
shane-huang: 4 cores?

apps/recommendation-wide-n-deep/README.md
```diff
@@ -0,0 +1,25 @@
...
+    --driver-memory 22g  \
```
shane-huang: BTW, is 22g really needed in wide & deep? Is it possible to reduce its memory usage too?

songhappy: As in the email, this PR just separate wide and deep from ncf, has nothing to do with resource management. These resources are copied from the old one since except the local[4], because my local[*] will give 8 cores, and Jason suggested 4 is enough

PR #540, opened 2018-07-17
split recommendation into ncf and wnd

songhappy:


apps/recommendation-ncf/README.md
```diff
@@ -14,7 +14,7 @@ These two notebooks demonstrate how to build neural network recommendation syste
 * Run `export ANALYTICS_ZOO_HOME=the dist directory under the Analytics Zoo project`.
 * Run the following bash command to start the jupyter notebook. Change parameter settings as you need, ie `MASTER = local[physcial_core_number]`.
 ```bash
-MASTER=local[*]
+MASTER=local[4]
```
shane-huang: Why 4 cores instead of *? What if a user has only 2 cores? Will fail to run in this case?

jason-dai:
No need to make change to branch 0.1

PR #543, opened 2018-07-17
Fix creating sc when import bigdl.nn.keras

hkvision:
Fix https://github.com/intel-analytics/zoo/issues/275


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/798/

PR #551, opened 2018-07-18
thread-safe abstract inference model

xuex2017:


yangw1234:
Should we consider the scenario where `loadxxx` and `reload` is  called concurrently with `predict`.

zhichao-li:
There are lot of duplicated code in the loadxx methods, it would be good if we can avoid those.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,71 +16,100 @@
...
-
+import java.lang.Object;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.LinkedBlockingQueue;
 
 public abstract class AbstractInferenceModel implements Serializable {
-  private FloatInferenceModel model;
+  //private FloatInferenceModel model;
```
glorysdj: this line can be deleted

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,71 +16,100 @@
...
   private int supportedConcurrentNum = 1;
+  protected LinkedBlockingQueue<FloatInferenceModel> modelQueue;
 
   public AbstractInferenceModel() {
   }
 
   public AbstractInferenceModel(int supportedConcurrentNum) {
     this.supportedConcurrentNum = supportedConcurrentNum;
+
```
glorysdj: modelQueue = new LinkedBlockingQueue<FloatInferenceModel>(supportedConcurrentNum); 
can be put here.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,71 +16,100 @@
...
+      FloatInferenceModel model = InferenceModelFactory.loadFloatInferenceModelForCaffe(modelPath, weightPath);
+      modelQueue.offer(model);
+    }
+
   }
 
   public void loadTF(String modelPath) {
-    this.model = InferenceModelFactory.loadFloatInferenceModelForTF(modelPath,
-            1, 1, true);
+    modelQueue = new LinkedBlockingQueue<FloatInferenceModel>(supportedConcurrentNum);
```
glorysdj: call loadTF(modelPath, 1, 1, true)?

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,71 +16,100 @@
...
+      modelQueue.offer(model);
+    }
   }
 
   public void reload(String modelPath) {
     load(modelPath, null);
   }
 
   public void reload(String modelPath, String weightPath) {
-    this.model = InferenceModelFactory.loadFloatInferenceModel(modelPath, weightPath);
```
glorysdj: we can deprecated reload for now.
and if we put modelQueue = new LinkedBlockingQueue in constructor, we can clear the queue here.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,71 +16,100 @@
...
+    for (int i = 0; i < supportedConcurrentNum; i++){
+      FloatInferenceModel model = InferenceModelFactory.loadFloatInferenceModel(modelPath, weightPath);
+      modelQueue.offer(model);
+    }
+
   }
 
   @Deprecated
-  public List<Float> predict(List<Float> input, int... shape) {
+  public List<Float> predict(List<Float> input, int... shape) throws InterruptedException {
```
glorysdj: should catche the InterruptedException  and deal it in the method, since we @Deprecated this method, we should not change the definition of this method.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,71 +16,100 @@
...
   }
 
-  public List<List<JTensor>> predict(List<JTensor> inputs) {
-    return model.predict(inputs);
+  public List<List<JTensor>> predict(List<JTensor> inputs) throws InterruptedException{
+    FloatInferenceModel model = modelQueue.take();
+    List<List<JTensor>> result =  model.predict(inputs);
+    modelQueue.offer(model);
+    return result;
   }
```
glorysdj: and please also format the code

glorysdj:
the offer and take methods are safe, loadxxx, reload, predict call offer and take, they are basically safe.
imo, load and reload should be called before any prediction, this should be controlled by the user.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
yangw1234: What if reload is called before some predict requests has finished, which, I think, is not uncommon in web services,  some of the old model will be push back to the queue again.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,71 +16,104 @@
...
+	}
+
+	public List<List<JTensor>> predict(List<JTensor> inputs) {
+		FloatInferenceModel model = null;
+		try {
+			model = modelQueue.take();
+		} catch (InterruptedException e) {
+			throw new InferenceRuntimeException("no model available", e);
+		}
+		List<List<JTensor>> result = model.predict(inputs);
```
yangw1234: What if model.predict throws an exception?
How about
```
try {
    model.predict(inputs);
} finally {
    modelQueue.offer(model);
}
```

zhichao-li:
I guess we should separate the responsibilities: 1) load model 2) a thread-safe inference utility method. There's a good example for doing this: `BigDL/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala`.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
glorysdj: we catch the InterruptedException here is to keep the method compatible as before.
after we caught the exception, we wrapper it as runtimeexception and throw it.
if model.predict throw exception, it will throw as runtimeexception, same way.

xuex2017: how about this?
```
public List<List<JTensor>> predict(List<JTensor> inputs) {
		FloatInferenceModel model = null;
		List<List<JTensor>> result;
		try {
			model = modelQueue.take();
			result = model.predict(inputs);
		} catch (InterruptedException e) {
			throw new InferenceRuntimeException("no model available", e);
		} finally {
			modelQueue.offer(model);
		}
		return result;
	}
```

glorysdj: this does not catch any exception from model.predict, it only catch InterruptedException

yangw1234: @xuex2017 if InterruptedException is thrown from modelQueue, then the model is still null, modelQueue.offer() will put a null in the queue.

@glorysdj If there is no finally and model.predict throws an exception, the model taken from queue will be lost.

glorysdj:
passed Jenkins http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/826/

PR #554, opened 2018-07-19
Fix fit method

zhichao-li:


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -700,3 +740,27 @@ object Sequential extends KerasLayerSerializable {
     new Sequential[T]()
   }
 }
+
+private[zoo] class ZooLocalOptimizer[T: ClassTag] (
```
jason-dai: InternalLocalOptimizer?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -700,3 +740,27 @@ object Sequential extends KerasLayerSerializable {
...
+    model: Module[T],
+    ds: LocalDataSet[MiniBatch[T]],
+    criterion: Criterion[T]
+)(implicit ev: TensorNumeric[T]) extends LocalOptimizer[T](model, ds, criterion) {
+
+  def setTrainData(trainingDataSet: DataSet[MiniBatch[T]]): this.type = {
+     this.dataset = trainingDataSet
+    this
```
jason-dai: `endEpoch()`?

zhichao-li: `optimize()` is a blocking method. Seems like we don't need to manually invoke `endEpoch` for increasing the epoch number. And also if we invoke `setTrainData` multiple times then the `endEpoch` would be increased even if we haven't started to `optimize()` cc @yangw1234 what's the reason for adding `endEpoch`?

yangw1234: @zhichao-li endEpoch only increase epoch number when the current epoch is not finished. If current epoch is finished, endEpoch will do nothing.

zhichao-li:
Jenkins passed. #829

PR #557, opened 2018-07-20
add methods in InferenceModelFactory for cloning model

edieYoung:
added couple methods for cloning model. 
For usage, 
1. EmptyModel: you need to firstly load a full model, and use cloneModule() to get emptyModel, a copy of it. Then, clear the weights in emptyModel for saving memory usage, so this emptyModel can be copied multiple times for multiple threads. 
* InferenceModelFactory.clearWeightsBias(the model you want to clear its weightsBias)

2.  WeightsBias: Besides, use getWeightsBias to get the weights in the full model (this is not a memory copy, so do not modify the original full model). 

For multiple threads, each one gets one emptyModel, and share the weightsBias, so there will not be multiple copies of weights taking a lot of memory.
*   InferenceModelFactory.makeUpModel(emptyModel, weightsBias)


glorysdj:
we may need methods like loadFloatInferenceModel/loadFloatInferenceModelForCaffe/loadFloatInferenceModelForTF
with one more parameter: int supportedConcurrentNum
to get a list of models with shared weights

edieYoung:
Tests indicate that cloning models can both save time and the memory usage, especially when the number of models is large.
The detail is as follow:

1. original (Load directly 100 times) : 
* Time/ms: 7981, 8203, 8252
* %Memory: 6.2, 6.1, 6.1

2. new (Load once and clone 100 times) : 
* Time/ms:  3145, 3194, 3108
* %Memory:  2.7 , 2.7, 2.7

3. original (Load directly 1000 times) : 
* Time/ms: 47246, 46850, 47212
* %Memory: 22.9, 22.5, 22.8

4. new (Load once and clone 1000 times) : 
* Time/ms:  8528, 8805, 8427
* %Memory:  11.6 , 12.6, 13.3

glorysdj:
code are not latest? it seems that makeUpModel and supportedConcurrentNum are not used.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -16,107 +16,127 @@
...
+        this.supportedConcurrentNum = supportedConcurrentNum;
+        modelQueue = new LinkedBlockingQueue<>(supportedConcurrentNum);
+    }
+
+    public void load(String modelPath) {
+        load(modelPath, null);
+    }
+
+    public void load(String modelPath, String weightPath) {
+        FloatInferenceModel originModel = InferenceModelFactory.loadFloatInferenceModel(modelPath, weightPath, supportedConcurrentNum);
```
glorysdj: add a method loadFloatInferenceModel/loadFloatInferenceModelForCaffe/loadFloatInferenceModelForTF
with one more parameter: int supportedConcurrentNum to get a list of models with shared weights
and offer them to the queue

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModelFactory.scala
```diff
@@ -16,25 +16,30 @@
...
+  def loadFloatInferenceModel(modelPath: String, weightPath: String, supportedConcurrentNum : Int)
+  : FloatInferenceModel = {
     val model = ModelLoader.loadFloatModel(modelPath, weightPath)
     val predictor = LocalPredictor(model = model, batchPerCore = 1)
     model.evaluate()
     new FloatInferenceModel(model, predictor)
   }
 
-  def loadFloatInferenceModelForCaffe(modelPath: String, weightPath: String)
-    : FloatInferenceModel = {
```
glorysdj: supportedConcurrentNum are not used. change the return as a list of shared weight models

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModelFactory.scala
```diff
@@ -44,12 +49,52 @@ object InferenceModelFactory {
...
+    var i = 0
+    while (i < tensors.length) {
+      if (tensors(i) != null) {
+        tensors(i).set()
+      }
+      i += 1
+    }
+  }
+
+  def clearWeightsBias(model: Module[Float]): Unit = {
```
glorysdj: can be private method

edieYoung:
added these methods as follow in InferenceModelFactory:
1. cloneSharedWeightsModelsIntoArray(): returning an array of shared weights models.
2. loadFloatInferenceModelArrayWithSharedWeights()
3. loadFloatInferenceModelArrayWithSharedWeightsForCaffe()
4. loadFloatInferenceModelArrayWithSharedWeightsForTF()

and they are called in AbstractInferenceModel

PR #559, opened 2018-07-24
ncf goes public

songhappy:
Changed Recommender models(NCF and WideAndDeep) from private to public.

Currently, features of recommender models are well fixed according to the papers and blogs, customers might want to add different features in different formats. For example, NCF only uses userId and itemId as features, but jobs2career wants to have features of vectors. 
https://github.com/intel-analytics/Jobs2Careers/blob/master/src/main/scala/com/intel/analytics/bigdl/apps/recommendation/NCFJob2Career.scala

Public NCF/WideAndDeep gives customers flexibility to build different recommender models with different features.


jason-dai:
Please add description for the change (and why it is needed).

jason-dai:
LGTM

PR #570, opened 2018-07-27
add unit test for InferenceModel to load models with shared weights

edieYoung:
added a unit test for timing the loading time of an array of models with shared weights.
the predict part is depended on the model and input data.


edieYoung:
Output examples:
1. autograd: Total 100 models sharing weights have been loaded. Time elapsed:2757.
2. textClassification: Total 100 models sharing weights have been loaded. Time elapsed:3145.

PR #572, opened 2018-07-30
Set serializer to java in conf for TextClassification example

hkvision:
Since kryo serializer will cause problem


jason-dai:
Can we do it by default in init_nncontext?

hkvision:
@jason-dai You want to use java serializer everywhere? Then I think we can specify this in `spark-analytics-zoo.conf`?

jason-dai:
Yes, maybe we should do that.

hkvision:
Okay. I will create another PR to do this.

PR #575, opened 2018-07-30
Add java serializer to spark-analytics-zoo.conf

hkvision:
We will now have our own conf and won't rely on spark-bigdl.conf anymore.


hkvision:
@glorysdj Could you please have a review about pom change? Thanks.

hkvision:
Now in Scala, `initNNContext()` will include conf in our `spark-analytics-zoo.conf`.
In Python, `init_nncontext()` will do the same thing.
Removed `spark-bigdl.conf` from PYTHONPATH as we don't rely on it anymore.
This also gives flexibility in the future if we want to add more conf to our project.
Currently I add java serializer by default.
Have a review? @jason-dai @zhichao-li

hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/912/

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
```diff
@@ -178,5 +179,30 @@ object NNContext {
     initNNContext(null, null)
   }
 
+  /**
+   * Read spark conf values from spark-analytics-zoo.conf
+   */
+  private[zoo] def readConf: Seq[(String, String)] = {
```
zhichao-li: can we  depend on the method in BigDL directly instead of copying the code if no specific reason?

pyzoo/zoo/common/nncontext.py
```diff
@@ -102,23 +167,24 @@ def _get_bigdl_verion_conf():
...
                         content = str(content, 'latin-1')
-                    return load_conf(content)
+                    return load_conf(content, "=")
     raise RuntimeError("Error while locating file zoo-version-info.properties, " +
                        "please make sure the mvn generate-resources phase" +
                        " is executed and a zoo-version-info.properties file" +
                        " is located in zoo/target/extra-resources")
+
+
+def load_conf(conf_str, split_char=None):
```
zhichao-li: can we simply use  `load_conf` from BigDL?

pyzoo/zoo/common/nncontext.py
```diff
@@ -48,14 +48,79 @@ def getOrCreateSparkContext(conf=None, appName=None):
...
+        if existing_py_files:
+            sparkConf.set(key="spark.submit.pyFiles",
+                          value="%s,%s" % (python_lib, existing_py_files))
+        else:
+            sparkConf.set(key="spark.submit.pyFiles", value=python_lib)
+
+    return sparkConf
+
+
+def get_spark_context(conf=None):
```
zhichao-li: How about use `getOrCreateSparkContext` method or update it if need to add more features?

zoo/src/main/scala/com/intel/analytics/zoo/common/NNContext.scala
hkvision: This code in BigDL is hardcoded to be `spark-bigdl.conf`

pyzoo/zoo/common/nncontext.py
hkvision: This is already in our code. I just extend it with `split_char` to support both load conf and zoo version properties.

pyzoo/zoo/common/nncontext.py
hkvision: Deleted.

hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/913/

PR #582, opened 2018-08-02
model.predict() should set default batch size to -1

zhichao-li:
https://github.com/intel-analytics/zoo/issues/351


zhichao-li:
Jenkins passed. ZOO-PR-Validation/868/

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/utils/KerasUtils.scala
```diff
@@ -424,9 +424,17 @@ object KerasUtils {
     }
   }
 
+  def calBatchPerPartition(batchSize: Int, partitionNum: Int): Int = {
+    if (batchSize > 0) {
+      batchSize / partitionNum
```
jason-dai: what if batchSize < partitionNum?

PR #583, opened 2018-08-02
Fix LossFunction updateOutput

hkvision:
Due to the inconsistency of implementations of criterions in BigDL, directly assign the value of `updateOutput` to output in LossFunction instead of getting `loss.output`


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/873/

PR #590, opened 2018-08-06
Change batchSize to batchPerThread for `predict` API

zhichao-li:


zhichao-li:
@hhbyyh  Seems like nnframe only depend on `forward` method only, so I suppose this changes would not affect `nnframe` API.

zhichao-li:
Jenkins passed.

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -207,25 +207,23 @@ def convert_output(output):
...
         # Arguments
         x: Prediction data. A Numpy array or RDD of Sample or ImageSet.
-        batch_size: When distributed is True, the default value is 4 * rdd.getNumPartitions.
-                    When distributed is False the default value is 4 * numOfCores.
-                    If you want to tune the batch_size,
-                    you should make sure its value can be divisible by
-                    rdd.getNumPartitions(distributed=True) or numOfCores(distributed=False)
+        batch_per_partition:
+          When distributed is True,the default total batch size is 4 * rdd.getNumPartitions.
+          When distributed is False the default total batch size is 4 * numOfCores.
```
hhbyyh: 4 can be misleading here. does it mean `batch_per_partition` * `rdd.getNumPartitions`.

And for local mode where x is nparray, I don't think `numOfCores` is corresponding to `rdd.getNumPartitions`. Just using `batch_per_partition` as the batch size for local mode seems to be more intuitive for me.

zhichao-li: Currently, we  try to split the input data by numOfCores for processing the data in parallel, so the effective total_batch_size still batch_per_partition * numOfCores.
Maybe change the `batchPerPartition` to `batchPerThread` would be more straightforward.

zhichao-li:
ZOO-PR-Validation/906/

PR #593, opened 2018-08-06
add docs of Average layer

xiaxin1998:
add docs of Average Layer


docs/docs/Keras2StyleAPIGuide/Layers/merge.md
```diff
@@ -394,3 +394,228 @@ Output is
...
+```
+
+**Parameters:**
+
+* `inputs`: A list of input tensors (at least 2).
+* `**kwargs`: Standard layer keyword arguments.
+
+**Scala example:**
+```scala
+val input1 = Tensor[Float](3, 10).rand(0, 1)
```
zhichao-li: include the importing statement here?

PR #594, opened 2018-08-06
Dockerfile update

glorysdj:
This PR is to add more runtime args, user can specify the args for different apps/environments.
Also add back tensorflow

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/594)


yangw1234:
LGTM

PR #597, opened 2018-08-07
add a transformer to support 0-1 image normalize

dding3:
Fix https://github.com/intel-analytics/zoo/issues/355#issuecomment-410875333


jason-dai:
MinMaxScaler http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html?

jason-dai:
Any update?

pyzoo/zoo/feature/image/imagePreprocessing.py
```diff
@@ -85,6 +85,22 @@ def __init__(self, mean_r, mean_g, mean_b, std_r=1.0,
                                                     mean_b, std_r, std_g, std_b)
 
 
+class ImageNormalize(ImagePreprocessing):
+    """
+    Normalizes the norm or value range, similar to opencv::normalize
```
hhbyyh: Maybe add a little more to the comments.
```ImageNormalize normalizes scale and shift the input features. Various normalize methods are supported, E.g. NORM_INF, NORM_L1, NORM_L2 or NORM_MINMAX.```

File: pyzoo/zoo/feature/image/imagePreprocessing.py [hunk 2]
```diff
@@ -85,6 +85,22 @@ def __init__(self, mean_r, mean_g, mean_b, std_r=1.0,
...
+    https://docs.opencv.org/ref/master/d2/de8/group__core__array.html#ga87eef7ee3970f86906d69a92cbf064bd
+    :param min lower range boundary in case of the range normalization or
+    norm value to normalize
+    :param max upper range boundary in case of the range normalization.
+    It is not used for the norm normalization.
+    :param norm_type normalization type, see opencv:NormTypes.
+    https://docs.opencv.org/ref/master/d2/de8/group__core__array.html#gad12cefbcb5291cf958a85b4b67b6149f
+    Default Core.NORM_MINMAX
+    """
+    def __init__(self, min, max, norm_type=32, bigdl_type="float"):
```
hhbyyh: Can we add an extra param `norm` to avoid reusing `min`, to make it more clear?
also 32 --> Core.NORM_MINMAX?

zoo/src/main/scala/com/intel/analytics/zoo/feature/image/ImageChannelNormalize.scala
```diff
@@ -54,3 +56,35 @@ object ImageChannelNormalize {
...
+ * @param min lower range boundary in case of the range normalization or
+ *            norm value to normalize
+ * @param max upper range boundary in case of the range normalization;
+ *            it is not used for the norm normalization.
+ * @param normType normalization type, see opencv:NormTypes.
+ *           https://docs.opencv.org/ref/master/d2/de8/group__core__array.html
+ *           #gad12cefbcb5291cf958a85b4b67b6149f
+ *           Default Core.NORM_MINMAX
+ */
+class ImageNormalize(min: Double, max: Double, normType: Int = Core.NORM_MINMAX)
```
hhbyyh: Same as above, can we add norm?

zoo/src/main/scala/com/intel/analytics/zoo/feature/image/ImageSet.scala
```diff
@@ -121,34 +121,56 @@ object ImageSet {
    * create LocalImageSet
    * @param data array of ImageFeature
    */
-  def array(data: Array[ImageFeature]): LocalImageSet = {
+  private[zoo] def array(data: Array[ImageFeature]): LocalImageSet = {
```
hhbyyh: Since it's a public API already, shall we deprecate it first?

zoo/src/test/scala/com/intel/analytics/zoo/feature/ImageFeatureSpec.scala
```diff
@@ -1,48 +0,0 @@
...
-
-import com.intel.analytics.bigdl.nn.abstractnn.DataFormat
-import com.intel.analytics.bigdl.transform.vision.image.ImageFeature
-import com.intel.analytics.bigdl.tensor.Tensor
-import com.intel.analytics.zoo.feature.common.{BigDLAdapter, Preprocessing}
-import com.intel.analytics.zoo.feature.image.{ImageMatToTensor, ImageResize, ImageSet}
-import org.scalatest.{BeforeAndAfter, FlatSpec, Matchers}
-
-
-class ImageFeatureSpec extends FlatSpec with Matchers with BeforeAndAfter {
```
hhbyyh: Is this removed intentionally?

pyzoo/zoo/feature/image/imagePreprocessing.py
dding3: Updated

pyzoo/zoo/feature/image/imagePreprocessing.py
dding3: I followed OpenCV's behavior, I am thinking if we use the same interface with OpenCV, it may be friendly for OpenCV user?
And yes, Core.NORM_MINMAX is 32 in OpenCV

zoo/src/main/scala/com/intel/analytics/zoo/feature/image/ImageSet.scala
dding3: Since the api is still used in Zoo internal, and it's already used by some users, maybe I should change the scope back?

zoo/src/test/scala/com/intel/analytics/zoo/feature/ImageFeatureSpec.scala
dding3: I was trying to move the ut under ImageFeatureSpec to FeatureSpec but somehow missed the ut. Added the ut to FeatureSpec.

dding3:
I understand your concern. In our current implementation it's hard to gracefully do an overall normalization. We need get the global min, max value of each axis of train/test before normalization.

I checked tensorflow, there is a per_image_standarlize which permforms standardlize per image, same behavior as this pr but in different normalization way. And looks like it's used in Cifar10 example? For pytorch, looks like there is only a normalize method if we want a 0-1 normalize pixels. And this api is the same with our ImageChannelNormalize.

I searched if per image standardization will lead to a lower accuracy and find various opinions. @hhbyyh what do you think if we emphasize this transformer is a per image normalization and let user make a choice?

hhbyyh:
Sounds good. E.g. PerImageNormalization?

dding3:
cc @hhbyyh

dding3:
Jekins pass. http://10.239.47.211:18888/job/ZOO-PR-Validation/1135/

hhbyyh:
LGTM. Thanks for the update.

dding3:
Thank you for your review and suggestions.

PR #604, opened 2018-08-10
Load onnx model

zhichao-li:
This PR would focus on initializing the structure which support a Convolution2D only, and would add more support shortly.


zhichao-li:
Jenkins passed #948

PR #613, opened 2018-08-14
remove unnecessary require in Reshape

hhbyyh:
nonBatchInput could contain -1 in it and breaks the require constraint. The following code (same as in unit test) fails with the original code

```
from zoo.pipeline.api.keras.layers import *
from zoo.pipeline.api.keras.models import Model
from zoo.common.nncontext import *
import numpy as np

sparkConf = create_spark_conf().setMaster("local[1]").setAppName("testNNClassifer")
sc = init_nncontext(sparkConf)

a = np.random.random((2, 2, 3, 4))
i1 = Input(shape = (2, 3, 4))
s = Reshape((-1, 2, 12))(i1)
m = Model(i1, s)
y = m.predict(a, distributed=False)
print(y)
```

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/Reshape.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/Reshape.scala
@@ -75,8 +75,8 @@ class Reshape[T: ClassTag](
     if (infer) {
       val nElements = nonBatchInput.product
       val resizeElements = - targetShape.product
-      require(nElements > resizeElements && nElements % resizeElements == 0,
-        "Total size after reshape must be unchanged")
+      require(nElements % resizeElements == 0, s"Total size after reshape must be unchanged." +
+          s" inputShape: $inputShape, targetShape: ${targetShape.mkString(", ")}")
       targetShape(inferIndex) = nElements / resizeElements
     }
     else {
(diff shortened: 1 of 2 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/Reshape.scala
```diff
@@ -75,8 +75,8 @@ class Reshape[T: ClassTag](
     if (infer) {
       val nElements = nonBatchInput.product
       val resizeElements = - targetShape.product
-      require(nElements > resizeElements && nElements % resizeElements == 0,
-        "Total size after reshape must be unchanged")
+      require(nElements % resizeElements == 0, s"Total size after reshape must be unchanged." +
```
hkvision: I checked the example you provide. Non-batch input shape is (2, 3, 4) after Reshape (-1, 2, 12), the target non-batch shape ought to be (1, 2, 12).
Then what about changing `nElements > resizeElements` to `nElements >= resizeElements`? Then this exception won't be raised. I suppose we need this constraint, otherwise when we do `nElements / resizeElements` it will be less than 1.
And sorry for this bug I left.

hkvision: Ar, just reminded that `nElements % resizeElements == 0` can guarantee `nElements >= resizeElements`...
Ignore my previous comment...

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/962/ passed. Thanks for the review @hkvision

PR #629, opened 2018-08-29
tf training check inputs

yangw1234:


yangw1234:
jenkins fail due to unrelated error

PR #638, opened 2018-09-05
Add TFPredictor

yangw1234:


yangw1234:
pr validation pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1018/

pyzoo/zoo/pipeline/api/net.py
```diff
@@ -339,17 +367,39 @@ def optimize(self, end_trigger=None, batch_size=32):
 
 class TFDataset:
 
-    def __init__(self, rdd, names, shapes, types):
+    def __init__(self, rdd, names, shapes, types, batch_size=None, hard_code_batch_size=False):
+        _, core_num = get_node_and_core_number()
+        if batch_size is None:
```
jason-dai: we have different requirements of batch size for training and prediction:
1) For training, we ask user to set total batch size and it should be a multiple of node_number*core_number
2) For prediction, we ask user to set batch size per partition

jason-dai: we have different requirements of batch size for training and prediction:

For training, we ask user to set total batch size and it should be a multiple of node_number*core_number
For prediction, we ask user to set batch size per partition

PR #643, opened 2018-09-07
[WIP]Export TFNet should raise an error when inputs do not in the graph defined by outputs

LudviqIntel:
Raise error-info about nodes not exit in the model


LudviqIntel:
http://10.239.47.211:18888/job/ZOO-PR-Validation/1023/
pass

PR #645, opened 2018-09-10
Add get_uri to Python Imageset API.

hhbyyh:
Enable users to map the image, label, predict to URI, this will further allow setting labels to Images.

### DIFF
--- a/pyzoo/test/zoo/feature/image/test_image_set.py
+++ b/pyzoo/test/zoo/feature/image/test_image_set.py
@@ -56,6 +56,14 @@ class Test_Image_Set():
         image_set = ImageSet.read(self.image_path)
         image_set.get_label()
 
+    def test_get_uri_local(self):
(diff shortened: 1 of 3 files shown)


jason-dai:
Why Python only?

hhbyyh:
Scala API already exposed access for ImageFeature, user can get URI using map operation.

jason-dai:
We should scala and python APIs consistent as much as possible

jason-dai:
I don't think we need to expose Imagefeature at this moment. We should add new helper methods (like get_uri) to both Scala and Python APIs though.

dding3:
Synced with yuhao offline, it would be convenient if we can have a "get" method for both scala and python, which supports user get the value of imageset with provided field. Since localImageSet returns a array while distributedImageSet returns a rdd, we may need add "getList" and "getRdd" separately for localImageSet and distributedImageset.

PR #657, opened 2018-09-12
Export TFNet should raise an error when inputs do not in the graph defined by outputs And test code

LudviqIntel:
Export TFNet should raise an error when inputs do not in the graph de…


LudviqIntel:
pass
http://10.239.47.211:18888/job/ZOO-PR-Validation/1087/

yangw1234:
LGTM

PR #666, opened 2018-09-16
supports check point for nnframes

hhbyyh:


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/nnframes/NNEstimator.scala
```diff
@@ -111,6 +111,16 @@ private[nnframes] trait TrainingParams[@specialized(Float, Double) T] extends Pa
...
+
+  /**
+   * Set a check point saved at `path` triggered by `trigger`
+   * Default: not enabled
+   */
+  final val checkpointPath = new Param[String](this, "checkpointPath", "path for check points")
+  final val checkpointTrigger = new Param[Trigger](this, "checkpointTrigger",
+    "Trigger for check points")
+
+  def getCheckpoint: (String, Trigger) = ($(checkpointPath), $(checkpointTrigger))
```
dding3: Looks like Trigger is a trait which only has an apply method. After user get checkpointTrigger, they still don't know what trigger it is?

dding3:
If user use a dir as cp path and run it more than one time, there will be exception, do we need support overWriteCheckpoint? It's not a big issue, we can do it in another PR if needed.

hhbyyh:
Thanks for the suggestion. Just found that BigDL Optimizer has different API for Scala and Python. I'll add overwrite similar to python interface.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/nnframes/NNEstimator.scala
hhbyyh: Triggers like EveryEpoch are defined as classes in Python. I added test in the unit tests to confirm the return result.

dding3: Sounds good. Please also add test for scala to confirm the result.

hhbyyh: For Scala, the best solution is overwrite toString for Triggers defined in BigDL....

dding3: Sounds good to me

dding3:
LGTM

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1079/ passed. Thanks for the review. Merging to master.

PR #667, opened 2018-09-17
Imageset.read with resize keep the same channel with original image

dding3:
Bug fix for https://github.com/intel-analytics/zoo/issues/379


dding3:
If Dell want to use ImageSet to read a gray image and extend it to RGB image, I think we support it now, set imageCodec with ImageSet.read

hhbyyh:
I see. Then we can merge this.

PR #668, opened 2018-09-17
ImageSet.get_image should not change channel of original image

dding3:
Fix https://github.com/intel-analytics/zoo/issues/380


dding3:
cc @hhbyyh Update.

dding3:
Synced with yuhao offline. If we use ImageMatToTensor and ImageFeatureToTensor in get_image, there will be exception for some examples(eg. object detection) as ImageMatToTensor can be called only once. Also it may bring extra computation overhead.

dding3:
Thanks for your review. Jekins pass. http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1114/

PR #672, opened 2018-09-18
change inference model to support batch forward

glorysdj:
This PR is to change the execution of batch inference from while loop to batch forward.
Please help to review the logic.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/672)

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
@@ -82,10 +82,7 @@ trait InferenceSupportive {
   def transferBatchTensorToJListOfJListOfJTensor(batchTensor: Tensor[Float], batchSize: Int)
   : JList[JList[JTensor]] = {
     val batchTensorSize: Array[Int] = batchTensor.size()
-    val batchShape: Array[Int] = batchTensorSize.length match {
-      case 1 => 1 +: batchTensorSize
-      case _ => batchTensorSize
-    }
+    val batchShape: Array[Int] = batchTensorSize
     val outputLength = batchTensor.nElement() / batchSize
     val outputShape = batchShape.tail
     require(batchSize == batchShape.head, "batchSize should be same")
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
@@ -150,7 +150,7 @@ trait InferenceSupportive {
         transferTensorsToTensorOfBatch(tensors.toArray)
       case x =>
         val inputTable = T()
-        val tensorsArray = Array.range(0, x).map(i => ArrayBuffer[JTensor]())
+        val tensorsArray = (1 to x).map(i => ArrayBuffer[JTensor]())
         var i = 0
         while (i < batchSize) {
           val inputList = inputs.get(i)
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
@@ -85,9 +85,9 @@ trait InferenceSupportive {
     val batchShape: Array[Int] = batchTensorSize
     val outputLength = batchTensor.nElement() / batchSize
     val outputShape = batchShape.tail
-    require(batchSize == batchShape.head, "batchSize should be same")
+    require(batchSize == batchShape.head, "batchSize should be the same")
     require(outputLength == outputShape.reduce(_ * _),
-      "data length should equal to product of shape")
+      "data length should be equal to the product of shape")
     val outputs = new util.ArrayList[JList[JTensor]]()
     var i = 0
     while(i < batchSize) {
@@ -133,7 +133,7 @@ trait InferenceSupportive {
 
   def transferListOfActivityToActivityOfBatch(inputs: JList[JList[JTensor]], batchSize: Int)
   : Activity = {
-    require(batchSize == inputs.size, "batchSize should be same")
+    require(batchSize == inputs.size, "batchSize should be the same")
     val head = inputs.get(0)
     val headLength = head.size()
     headLength match {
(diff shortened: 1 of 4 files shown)


glorysdj:
PR validated
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1101/

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -82,6 +131,62 @@ trait InferenceSupportive {
     new JTensor(res, outputShape, false)
   }
 
+  def transferListOfActivityToActivityOfBatch(inputs: JList[JList[JTensor]], batchSize: Int)
+  : Activity = {
+    val head = inputs.get(0)
```
hkvision: Do we need to check batchSize == inputs.size() here?

hkvision: Actually do we need this parameter batchSize? As it will always equal inputs.size(). Also doubt about this for other methods.

glorysdj: yes, i will add this check

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -73,6 +79,49 @@ trait InferenceSupportive {
     result
   }
 
+  def transferBatchTensorToJListOfJListOfJTensor(batchTensor: Tensor[Float], batchSize: Int)
+  : JList[JList[JTensor]] = {
+    val batchTensorSize: Array[Int] = batchTensor.size()
+    val batchShape: Array[Int] = batchTensorSize
+    val outputLength = batchTensor.nElement() / batchSize
+    val outputShape = batchShape.tail
+    require(batchSize == batchShape.head, "batchSize should be same")
```
hkvision: the same

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -73,6 +79,49 @@ trait InferenceSupportive {
...
+    require(outputLength == outputShape.reduce(_ * _),
+      "data length should equal to product of shape")
```
hkvision: be equal to the product of

yangw1234:
LGTM

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -82,6 +131,62 @@ trait InferenceSupportive {
...
+    val data = ArrayBuffer[Float]()
+    var i = 0
+    while (i < batchSize) {
+      val tensor = tensors(i)
+      val tensorData = tensor.getData
+      data ++= tensorData
+      i += 1
+    }
+    require(data.length == shape.reduce(_ * _),
+      "data length should equal to product of shape")
```
hkvision: be equal to the product

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
glorysdj: in predict method, we only need to call input.size once to get batchSize, then use it several times in different method calls to save time.

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModelSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+    val tensorOfBatch = transferTensorsToTensorOfBatch(inputTensorsArray)
+    val data1 = inputTensorsArray.map(_.getData).flatten
+    val data2 = tensorOfBatch.storage().array()
+    data1 should be (data2)
+    tensorOfBatch.size() should be (Array(3, 3, 5, 5))
+  }
+
+  "transferListOfActivityToActivityOfBatch" should "work" in {
+    val tensorOfBatch = transferListOfActivityToActivityOfBatch(inputTensorList,
+      inputTensorList.size()).toTensor[Float]
```
hkvision: Do we need to input batchSize as an argument? As it is directly from inputTensorList and seems duplicate

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModelSpec.scala
```diff
@@ -98,10 +98,15 @@ class InferenceModelSpec extends FlatSpec with Matchers with BeforeAndAfter
...
     val inputTensor = Tensor[Float](3, 5, 5).rand()
     val input = new util.ArrayList[JTensor]()
-    input.add(transferTensorToJTensor(inputTensor))
+    val inputJTensor = transferTensorToJTensor(inputTensor)
+    input.add(inputJTensor)
     val inputs = new util.ArrayList[util.List[JTensor]]()
     inputs.add(input)
-    val result = fModels(0).predict(inputs).get(0).get(0)
+    inputs.add(input)
+    inputs.add(input)
```
hkvision: Is adding two same input to avoid the problem we met yesterday about the error from model with batch=1?

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/inference/FloatInferenceModelSpec.scala
glorysdj: this is a test to the method, in the call stack of transferListOfActivityToActivityOfBatch, before it we get the batchSize once and will pass it directly in several method calls later to save time.

hkvision: But you are still do the checking inside the method and it will still cause the calculation of the list size?

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModelSpec.scala
glorysdj: yes, to test batch. some legacy layer from bigdl dont have batch when it is 1.

PR #674, opened 2018-09-18
[ONNX] add Dropout, MatMul and test with mnist model

zhichao-li:


zhichao-li:
- Build # 1098 - Successful

PR #683, opened 2018-09-21
reduce messi notebook size to 6M

dding3:


dding3:
Jekins passed. http://10.239.47.211:18888/job/ZOO-PR-Validation/1118/

PR #686, opened 2018-09-25
Make WideAndDeepExample support float/long continuosCols.

bethunebtj:
See this: https://github.com/intel-analytics/analytics-zoo-internal/issues/1226


jason-dai:
Thanks for the patch. @songhappy please take a look.

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/Utils.scala
```diff
@@ -124,22 +124,38 @@ object Utils {
...
+    */
+  implicit def AnyValToFloat(x:AnyVal):Float = {
+    x match {
+      case i: Int => i.toFloat
+      case l: Long => l.toFloat
+      case f: Float => f
+      case d: Double => d.toFloat
+      case _ => 0.0f
+    }
+  }
```
songhappy: This implicit function would like to change any value if getAs[AnyVal] to float, it might cause some unwanted issue. 
Can you just use a regular function?

bethunebtj: I think you are right. Users may feel confused if they just use string or char or something else casually and the program just executes without any hint .

I will move the switch code in this section: ` (0 to deepColumns2.length - 1).map {` and throw an exception if a user passes an `anyval` of non integers , longs or doubles.

Is that ok?

PR #705, opened 2018-10-03
replace bytes with mat in nnframes

hhbyyh:
Since ImageSet.read already creates Mat from image bytes, NNImageReader now can just save the converted mat in DataFrame, and original files bytes are no longer useful and is removed from generated imageFeature.


dding3:
LGTM

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1170/ passed. Thanks for the review. Merging to master.

Issue #708, opened 2018-10-06
ChannelNormalize bug?


hhbyyh:
I was just checking the code in ImageChannelNormalize, and in BigDL 
https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/transform/vision/image/augmentation/ChannelNormalize.scala#L52

I find the following code:
```
  def apply(meanR: Float, meanG: Float, meanB: Float,
    stdR: Float = 1, stdG: Float = 1, stdB: Float = 1): ChannelNormalize = {
    new ChannelNormalize(Array(meanB, meanG, meanR), Array(stdR, stdG, stdB))
  }
```

Notice that  `new ChannelNormalize(Array(meanB, meanG, meanR), Array(stdR, stdG, stdB))`, mean and std have inconsistent order. This looks like a bug.

PR #710, opened 2018-10-08
bug fix for imagechannelnormalize

dding3:
Fix https://github.com/intel-analytics/analytics-zoo/issues/708


dding3:
Yes, existing examples are correct. It would be incorrect if user call imagechannelnormalize themselves. I think it's not harmful if we merge the fix now and after BigDL fix this issue, no changes are needed from zoo side :)

dding3:
Thanks for your review. Jekins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/1192/.

PR #713, opened 2018-10-09
[WIP] use tf config proto to configure session

yangw1234:


yangw1234:
does not work on bigdl-0.6.0

PR #719, opened 2018-10-10
Update using_variational_autoencoder_to_generate_digital_numbers.ipynb

yunyanliluo:


apps/variational-autoencoder/using_variational_autoencoder_to_generate_digital_numbers.ipynb
```diff
@@ -60,6 +60,7 @@
     "# a bit of setup\n",
     "import numpy as np\n",
     "from bigdl.dataset import mnist\n",
+    "from bigdl.nn.criterion import *\n",
```
hkvision: Seems that if you directly edit the ipynb page instead of coding, it will have a problem in recognizing the code: https://github.com/intel-analytics/analytics-zoo/blob/b595886bfda0c0e5dd7710b1b0b91296c1eb1bb7/apps/variational-autoencoder/using_variational_autoencoder_to_generate_digital_numbers.ipynb
See the place you add and there's error in rendering the notebook.

hkvision:
Also @glorysdj need to investigate why missing this import can pass the jenkins and previous release manual test?

PR #720, opened 2018-10-10
Update run-app-tests.sh

yunyanliluo:


hkvision:
What's the purpose of this PR?

glorysdj:
this PR is to update the run-apps script to accept different arguments to run different parts of the apps to split the one long jenkins job to several parts

hkvision:
Okay. Splitting into several jobs is a good choice, but doing it in this way is bit awkward...

glorysdj:
by default, if no arguments is passed to it, it remains the old way, run all of the apps.

Issue #723, opened 2018-10-11
zoo.keras.LSTM extend from zoo.keras.Recurrent?


dding3:
Looks like com.intel.analytics.zoo.pipeline.api.keras.layers.Recurrent is not used in anywhere? and zoo.keras.layers.LSTM is extend from bigdl.nn.keras.Recurrent? If I update LSTM to extend from zoo.keras.Recurrent, there will be exception：

### Caused by: com.intel.analytics.bigdl.nn.abstractnn.InvalidLayer: Do not mix Sequential[6ed52d6b](isKerasStyle=true) with Layer
                           (isKerasStyle=false):
         InternalRecurrent[f454c033]ArrayBuffer(TimeDistributed[5379272d]Linear[53d69303](12 -> 128), LSTM(12, 32, 0.0))
	at com.intel.analytics.bigdl.nn.abstractnn.InferShape$class.excludeInvalidLayers(InferShape.scala:98)
	at com.intel.analytics.bigdl.nn.abstractnn.AbstractModule.excludeInvalidLayers(AbstractModule.scala:58)
	at com.intel.analytics.bigdl.nn.abstractnn.InferShape$class.validateInput(InferShape.scala:108)
	at com.intel.analytics.bigdl.nn.abstractnn.AbstractModule.validateInput(AbstractModule.scala:58)
	at com.intel.analytics.bigdl.nn.keras.Sequential.add(Topology.scala:299)
	at com.intel.analytics.zoo.pipeline.api.keras.layers.Recurrent.doBuild(Recurrent.scala:41)
	at com.intel.analytics.bigdl.nn.keras.KerasLayer.build(KerasLayer.scala:225)

And I agreed we do need a zoo.keras.Recurrent extends from bigdl.keras.Recurrent so we can add more functions.

hkvision:
This is the bug from https://github.com/intel-analytics/analytics-zoo/pull/518 @zhichao-li 
1. First of all, `zoo Recurrent` is added and for those recurrent layers they still extend original `bigdl.nn,keras.Recurrent`, this is wrong.
2. When adding `zoo Recurrent`, TSequential and Select are wrongly imported and thus the error of mix torch and keras style layers occur.

PR #731, opened 2018-10-11
Fix Keras Recurrent layers bug

hkvision:
Fix https://github.com/intel-analytics/analytics-zoo/issues/723


zhichao-li:
There's serialization test within `InternalRecurrentSpec`

PR #743, opened 2018-10-15
[WIP] try to change ut to avoid batchsize=1 predictor error

glorysdj:
this pr is to change ut by adding one more pic to avoid bigdl predictor bug of batchsize=1

Edit: This bug is from some layer in the pretrained image model.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/743)


hkvision:
Can we change to a smaller image? I notice all other image are around ~40K while the one you added is ~200K

glorysdj:
sure, then i need to reopen a new pr, since this pic is committed and will be stored in the git history

PR #788, opened 2018-10-17
Refactor python to scala functions

yangw1234:
Make all python to scala converter classes to subclass one base class (PythonZoo) and PythonZoo subclass BigDL's class.

This way we can override the method in BigDL's class.


zhichao-li:
LGTM

yangw1234:
pr validation pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1210/

PR #795, opened 2018-10-17
fix tfnet cannot handle scalar

yangw1234:


pyzoo/zoo/pipeline/api/net.py
```diff
@@ -200,6 +202,20 @@ def load_keras(json_path=None, hdf5_path=None, by_name=False):
         return BModel.load_keras(json_path, hdf5_path, by_name)
 
 
+def to_sample_rdd(x, y, sc, num_slices=None):
```
hkvision: this is already in bigdl.util.common?

yangw1234: implementation is different, here we use the new Sample

hkvision: okay

PR #810, opened 2018-10-18
Add KNRM model

hkvision:
- Support `weights` and `trainable` in `Embedding` (Fix https://github.com/intel-analytics/zoo/issues/311)
- Add KNRM model in Scala and Python
MatchZoo KNRM model: https://github.com/NTMC-Community/MatchZoo/blob/master/matchzoo/models/knrm.py

This PR only aims at adding the model first. Will investigate/add the QA Ranker example including dataset preprocessing in following PRs.


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/1260/
@shane-huang This PR is almost finished.

pyzoo/zoo/pipeline/api/keras/layers/embeddings.py
```diff
@@ -45,15 +45,21 @@ class Embedding(ZooKerasLayer):
...
     >>> embedding = Embedding(1000, 32, input_length=10, name="embedding1")
     creating: createZooKerasEmbedding
+
+    >>> import numpy as np
+    >>> embedding = Embedding(10, 200, weights=np.random.random([10, 200]), input_length=10)
+    creating: createZooKerasEmbedding
     """
-    def __init__(self, input_dim, output_dim, init="uniform", input_length=None,
-                 W_regularizer=None, input_shape=None, **kwargs):
+    def __init__(self, input_dim, output_dim, init="uniform", weights=None, trainable=True,
```
shane-huang: Make sure the overriding between argument "init" and "weights" is correct.  And we need to add docs for new parameters in high level interface, even if it's following Keras 1.2.2 interface.

shane-huang: And I don't think Keras embedding layer has the parameter "trainable". We need to explain it to users.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/math.scala
```diff
@@ -57,17 +57,17 @@ object AutoGrad {
    * Sum of the values in a variable, alongside the specified axis.
    * @param x A variable.
    * @param axis axis to compute the mean. 0-based indexed.
-   * @param keepdims A boolean, whether to keep the dimensions or not.
```
shane-huang: If I understand it right, this change (together with a few others) only renames the `keepdims` param? Shall we separate it from this big pr? It will be cleaner for looking up in future.

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/EmbeddingSpec.scala
```diff
@@ -24,10 +24,12 @@ import com.intel.analytics.zoo.pipeline.api.keras.serializer.ModuleSerialization
 class EmbeddingSpec extends KerasBaseSpec {
 
   // Compared results with Keras on Python side
-  "Embedding" should "work properly" in {
```
shane-huang: Are you changing the original test to add weights? I think we'd better add a new test instead of change the original one, since people could also use the embedding layer for other purposes which might need lecun_uniform or other tantalizers - we need to test them as well.

shane-huang:
Generally LGTM. This PR only includes interface, model implementation and unit tests. Maybe we need examples. I think the original QA Ranking might be too heavy for an example. We can make it an app and use jupyter for demonstration. For examples we can use a simplified version.

pyzoo/zoo/pipeline/api/keras/layers/embeddings.py
hkvision: Sorry I forgot to update the Python docs...

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/math.scala
hkvision: Ok. I will revert first :)

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/EmbeddingSpec.scala
hkvision: Originally the unit test just runs forward and backward and doesn't do any correctness verification. Comparing with Keras is done on Python side. Since now we can specify weights, then I can add correctness test that forward results should match weights.

hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/1277/
Thanks for your review and comments @shane-huang

PR #830, opened 2018-10-22
Fix Accuracy for binary classification

hkvision:
A bug for zero-based label Accuracy.
In the case of binary classification, zero-based label shouldn't take effect.
Only to add 1 if it is a multiclass problem.


yangw1234:
LGTM

hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/1252/
Thanks for your review @yangw1234

PR #840, opened 2018-10-23
app/Sentiment

xiaxin1998:


xiaxin1998:
@hkvision  Jenkins passed view/ZOO-PR/job/ZOO-PR-Pip-Apps-Integration/89/

apps/run-app-tests-pip.sh
```diff
@@ -37,6 +37,56 @@ now=$(date "+%s")
...
+start=$(date "+%s")
+
+# Conversion to py file and data preparation
+${ANALYTICS_ZOO_HOME}/apps/ipynb2py.sh ${ANALYTICS_ZOO_HOME}/apps/sentiment-analysis/sentiment
+
+# Run the example
+export SPARK_DRIVER_MEMORY=12g
+python ${ANALYTICS_ZOO_HOME}/apps/sentiment-analysis/sentiment.py
+
+FILENAME="/tmp/.bigdl/dataset/glove.6B.zip"
```
hkvision: data should be put before the command

apps/recommendation-wide-n-deep/wide_n_deep.ipynb
```diff
@@ -319,7 +319,7 @@
    ],
    "source": [
     "# Create an Optimizer\n",
-    "batch_size = 8000\n",
+    "batch_size = 5600\n",
```
hkvision: don't modify the app. use `sed` to change in the script if necessary.

PR #861, opened 2018-10-24
Update run-app-tests-pip for object detection and image similarity 

cynthia-liu:


apps/run-app-tests-pip.sh
```diff
@@ -37,6 +37,142 @@ now=$(date "+%s")
...
+
+# Run the example 
+export SPARK_DRIVER_MEMORY=12g
+python ${ANALYTICS_ZOO_HOME}/apps/object-detection/object-detection.py
+
+exit_status=$?
+if [ $exit_status -ne 0 ];
+then
+    clear_up
+    echo "anomaly-detection failed"
```
hkvision: object detection?

cynthia-liu:
fixed @hkvision

apps/run-app-tests-pip.sh
cynthia-liu: I have fixed this.

Issue #724, opened 2018-10-11
Missing Convlstm related keras layer


dding3:
Since we are trying to use keras api for seq2seq pr, it would be great if we can have convlstm(2D/3D) keras layers.

hkvision:
We have ConvLSTM2D: https://github.com/intel-analytics/analytics-zoo/blob/master/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
I will add ConvLSTM3D soon.
Also feel free to tell me if there's any argument not exposed in Keras API but needed in seq2seq.

hkvision:
Code in ConvLSTM2D is not well written currently. I will refactor the code and add ConvLSTM3D after the fix of Recurrent.

dding3:
Yes, please extend ConvLSTM2D from zoo.keras.Recurrent other than BigDLConvLSTM2D

dding3:
On a second thought, currently ConvLSTM2D cannot be used, maybe we can fix it by extend it from bigdl.keras.Recurrent before zoo.Recurrent is fixed?

hkvision:
OK. Then I will fix it today.

hkvision:
It can be used (i.e. added to a Sequential). The problem is doesn't extend Recurrent and is not treated as a Recurrent layer. Will refactor it afterwards.

anpunjabi:
Any updates on the status of ConvLSTM3D?

hkvision:
Hi, @anpunjabi Thanks for your reminder. We will add ConvLSTM3D very soon.

anpunjabi:
Thank you very much, and let me know when the code is pushed. I have a 3D recurrent architecture that will greatly benefit from this addition.

hkvision:
Hi, @anpunjabi we have added ConvLSTM3D into master, which currently supports 'same' padding and 'channel_first' data format.
See here for Python API: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/pipeline/api/keras/layers/convolutional_recurrent.py#L92
See here for Scala API: https://github.com/intel-analytics/analytics-zoo/blob/master/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM3D.scala
Hope that it can meet your needs.

hkvision:
I will close this issue first. If you have further requests about ConvLSTM layers, feel free to raise another issue :).

PR #877, opened 2018-10-25
ConvLSTM2D extend Recurrent

hkvision:
enhancement for #724 

ConvLSTM2D should extend Recurrent instead of KerasLayer.

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
@@ -30,10 +30,10 @@ import scala.reflect.ClassTag
 
 /**
  * Convolutional LSTM.
- * Data format currently supported for this layer is 'CHANNEL_FIRST' (dimOrdering='th').
- * Border mode currently supported for this layer is 'same'.
- * The convolution kernel for this layer is a square kernel with equal strides 'subsample'.
- * The input of this layer should be 5D.
+ * Note that currently only 'same' padding is supported.
+ * The convolution kernel for this layer is a square kernel with equal strides.
+ * The input of this layer should be 5D, i.e. (samples, time, channels, rows, cols), and
+ * CHANNEL_FIRST' (dimOrdering='th') is expected.
  *
  * When using this layer as the first layer in a model, you need to provide the argument
  * inputShape (a Single Shape, does not include the batch dimension).
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
@@ -16,7 +16,9 @@
 
 package com.intel.analytics.zoo.pipeline.api.keras.layers
 
-import com.intel.analytics.bigdl.nn.keras.{KerasLayer, ConvLSTM2D => BigDLConvLSTM2D}
+import com.intel.analytics.bigdl.nn.{Cell, ConvLSTMPeephole}
+import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
+import com.intel.analytics.bigdl.nn.keras.KerasLayer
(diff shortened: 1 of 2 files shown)


hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/1281/

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
```diff
@@ -54,28 +56,55 @@ import scala.reflect.ClassTag
...
+   val subsample: Int = 1,
+   var wRegularizer: Regularizer[T] = null,
+   var uRegularizer: Regularizer[T] = null,
+   var bRegularizer: Regularizer[T] = null,
+   var returnSeq: Boolean = false,
+   var goBackward: Boolean = false,
+   var mInputShape: Shape = null)(implicit ev: TensorNumeric[T])
+  extends Recurrent[T](outputDimension, returnSeq, goBackward, mInputShape) with Net {
+
+  require(dimOrdering.toLowerCase() == "channel_first", s"ConvLSTM2D currently only supports " +
```
dding3: If we only support "channel_first", maybe we should not expose `dimOrdering`, if we need have the same parameters with keras, we should support other dimOrdering. 

BTW, do we need keep the same parameter name with keras? Looks like they're different. https://keras.io/layers/recurrent/#convlstm2d

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
```diff
@@ -54,28 +56,55 @@ import scala.reflect.ClassTag
...
-  extends BigDLConvLSTM2D[T](nbFilter, nbKernel, activation, innerActivation, dimOrdering,
-                              subsample, wRegularizer, uRegularizer, bRegularizer, returnSequences,
-                              goBackwards, inputShape) with Net {}
+   var outputDimension: Int,
+   val nbKernel: Int,
+   val activation: KerasLayer[Tensor[T], Tensor[T], T] = null,
+   val innerActivation: KerasLayer[Tensor[T], Tensor[T], T] = null,
+   val dimOrdering: String = "CHANNEL_FIRST",
+   val subsample: Int = 1,
+   var wRegularizer: Regularizer[T] = null,
```
dding3: miss padding?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
hkvision: What you are looking is Keras 2.x, here we refer to Keras 1.2.2

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
hkvision: In Keras, it uses `border_mode` = 'same'/'valid' for padding. Seems we only support 'same' and I ignore this argument. Alternatively, we can add a border_mode argument with only one available value like dimOrdering lol if you wish...

hkvision:
Let's continue on this PR cc @dding3

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/ConvLSTM2D.scala
dding3: Let's keep current implementation. After support `valid` padding, we will support user config padding way.

dding3:
Generally looks good to me. Please stress we only support "same" padding currently.

hkvision:
Highlighted only same padding is currently supported. We can add "borderMode" arg after support same padding. @dding3

hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/1848/
Thanks for your review @dding3

PR #884, opened 2018-10-25
Fix RegistryMap key not found bug

yangw1234:
if we fist test if the key is in the map then retrieve the key, there is a chance that the key got garbage collected right after the test, thus result in a KeyNotFound exception.


hkvision:
LGTM

yangw1234:
jenkins pass
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1288/
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1290/

PR #886, opened 2018-10-26
update tenserflow example test

cynthia-liu:


cynthia-liu:
@glorysdj jenkins has passed

PR #857, opened 2018-10-24
Add nnframes for run-example-test

Deliadd:

PR #891, opened 2018-10-26
Fix pip example integration

hkvision:
Error brought by #857


hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Pip-Example-Integration/47/

PR #924, opened 2018-10-31
[WIP]Zoo Optimizer 

songhappy:
This is a wrapper of BigDL optimizer. It splits the train rdd randomly into an array of rdd, with a length of nSplits, and train the model with a sub rdd at step by step.


zoo/src/main/scala/com/intel/analytics/zoo/optim/ZooOptimizer.scala
```diff
@@ -0,0 +1,384 @@
...
+
+import scala.reflect.ClassTag
+
+
+class ZooOptimizer[T: ClassTag, D](optimzer: Optimizer[T, D],
+                                   model: Module[T],
+                                   trainRdd: RDD[Sample[T]],
+                                   batchSize: Int,
+                                   nEpochs: Int,
+                                   nRdds: Int = 1) {
```
qiuxin2012: Why not extend DistriOptimizer and override `optimize()`?

PR #936, opened 2018-11-07
Character embedding layer

junyji:


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/charEmbedding.scala
```diff
@@ -0,0 +1,103 @@
...
+}
+
+object CharEmbedding {
+  def apply[@specialized(Float, Double) T: ClassTag](
+      inputDim: Int,
+      outputDim: Int,
+      charEmbedDim: Int,
+      inputLength: Int,
+      kernelRow: Int = 2,
+      inputShape: Shape = null)(implicit ev: TensorNumeric[T]): CharEmbedding[T] = {
```
hkvision: Change this to inputLength

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/charEmbeddingSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+package com.intel.analytics.zoo.pipeline.api.keras.layers
+
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.utils.Shape
+import com.intel.analytics.zoo.pipeline.api.keras.models.Sequential
+import com.intel.analytics.zoo.pipeline.api.keras.serializer.ModuleSerializationTest
+
+class CharEmbeddingSpec extends KerasBaseSpec {
+
+  // Compared results with Keras on Python side
```
hkvision: Have you compared the results?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/charEmbedding.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  def doBuild(inputShape: Shape): AbstractModule[Tensor[T], Tensor[T], T] = {
+    val input = inputShape.toSingle().toArray
+    val model = Sequential[T]()
+    val layer = Embedding(
+      inputDim = inputDim,
+      outputDim = charEmbedDim,
+      inputLength = input(1))
+    model.add(layer)
+    model.add(com.intel.analytics.bigdl.nn.keras.Reshape(Array(1, input(1), charEmbedDim)))
```
hkvision: For all these layers, please use `com.intel.analytics.zoo.pipeline.api.keras.layers`

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/CharEmbedding.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  def doBuild(inputShape: Shape): AbstractModule[Tensor[T], Tensor[T], T] = {
+    val input = inputShape.toSingle().toArray
+    val model = Sequential[T]()
+    val layer = com.intel.analytics.zoo.pipeline.api.keras.layers.Embedding(
+      inputDim = inputDim,
+      outputDim = charEmbedDim,
+      inputLength = input(1))
+    model.add(layer)
+    model.add(com.intel.analytics.zoo.pipeline.api.keras.layers.Reshape(Array(1, input(1), charEmbedDim)))
```
hkvision: As these layers are already in the same package, just use Reshape()?

hkvision:
@shane-huang His implementation is referred to Kim's paper. Shall we use Kim's or Zhang's?
https://github.com/mhjabreel/CharCnn_Keras/tree/master/models

hkvision:
Some general comments about the notebook: cc @shane-huang 
- If we use the new notebook you write, then we might need to delete the original one to avoid duplication.
- As for the example, may be we can have a Scala one since there's no Scala example about sentiment?
- Can we use TextSet related API? Probably I may need to make some extensions to support.
- `Prediction accuracy on validation set is:  0.8756` This seems even lower than the original version: `
Prediction accuracy on validation set is:  0.89312`?
- Please remark the time of training for each epoch.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/CharEmbedding.scala
```diff
@@ -0,0 +1,102 @@
...
+  * @param inputShape A Single Shape, does not include the batch dimension.
+  * @tparam T Numeric type of parameter(e.g. weight, bias). Only support float/double now.
+  */
+class CharEmbedding[T: ClassTag](
+    val inputDim: Int,
+    val outputDim: Int,
+    val charEmbedDim: Int,
+    val inputLength: Int,
+    val kernelRow: Int,
+    val inputShape: Shape = null)(implicit ev: TensorNumeric[T])
```
hkvision: Wondering is inputLength duplicate with inputShape?

apps/sentiment-analysis/sentiment_char_embedding.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    }
+   ],
+   "source": [
+    "from zoo.pipeline.api.keras.layers import *\n",
+    "from zoo.pipeline.api.keras.models import *\n",
+    "\n",
+    "word_input = Input(shape=(500,))\n",
+    "char_input = Input(shape=(500, char_input_length))\n",
+    "\n",
+    "char_emb = TimeDistributed(CharEmbedding(input_dim=char_input_dim, output_dim=char_output_dim, \\\n",
```
hkvision: Maybe we can think of how to avoid TimeDistributed.

apps/sentiment-analysis/sentiment_char_embedding.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    "                                         char_embed_dim=char_dim, input_length=char_input_length))(char_input)\n",
+    "# word_emb = Embedding(embedding_file=\"/tmp/.bigdl/dataset/glove.6B.100d.txt\", word_index=word_idx , trainable=False)(word_input)\n",
+    "word_emb = Embedding(input_dim=max_words, output_dim=word_embedding_dim, input_length=sequence_len, )\n",
+    "emb = word_emb(word_input)\n",
+    "word_emb.set_weights(np.array([w2v]))\n",
+    "word_emb.trainable = False\n",
```
hkvision: Do we support this syntax?

hkvision:
One question is is there any __existing code__ for model topology that you refer to for char embedding and fusion of char & word embeddings? Or you directly refer to the paper?

hkvision:
I find in this implementation: https://github.com/mhjabreel/CharCnn_Keras/blob/master/data_utils.py#L70
For each sentence, it is preprocessed as a sequence of character tokens of a fixed length. Then in this way we don't need TimeDistributed.
In your implementation, you perform on each word instead of each sentence.

JunyangJiang:
> One question is is there any **existing code** for model topology that you refer to for char embedding and fusion of char & word embeddings? Or you directly refer to the paper?

Please refer to https://github.com/yoonkim/lstm-char-cnn (LUA, original implementation from Kim)  and https://github.com/jarfo/kchar (Keras).

> For each sentence, it is preprocessed as a sequence of character tokens of a fixed length. Then in this way we don't need TimeDistributed. In your implementation, you perform on each word instead of each sentence.

Actually, it is very easy to apply similar Char-CNN on sentence structure. However, the reason for implementing word-level Char-CNN other than sentence-level Char-CNN is I think the restriction input is a sentence is too strong for there may be other NLP models like word/document classification. Besides the input is a word and the output is an embedding is more like a **char-embedding**  (at least for me).

hkvision:
I looked at the Keras implementation you provide. Analytics Zoo and Keras 1.2.2 only supports 2D input for Embedding, while later Keras versions support arbitrary input shape: https://github.com/keras-team/keras/pull/6392
Thus in this case, seems we need to use TimeDistributed.

hkvision:
Word-level or sentence-level char-embedding are both fine for me. Just for an example I think the latter is also ok if the result is good? As for the latter we don't need to do tokenization and we don't need TimeDistributed...

PR #941, opened 2018-11-07
Upgrade BigDL dependency to 0.7.1 and update Scala doc

hkvision:


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/1338/

PR #944, opened 2018-11-08
Fix pip install release issue in setup.py

hkvision:
http://10.239.47.211:18888/view/ZOO-NB/job/ZOO-NB-Deploy-PYTHON/238/console

Previously get the version need to execute the python file, which will throw the exception that bigdl hasn't been installed.
Now just read the lines and get the version.


hkvision:
Deploy jenkins passed http://10.239.47.211:18888/view/ZOO-NB/job/ZOO-NB-Deploy-PYTHON/240/

PR #947, opened 2018-11-09
Restrict pyspark version <2.4 for pip dependency

hkvision:
Newly released pyspark 2.4.0 will cause problem in our example.


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/1344/

PR #949, opened 2018-11-09
Change download URL to use "dist.apache.org"

engapa:
Spark 2.4.0


jason-dai:
@glorysdj please take a look

glorysdj:
both url works, i think maybe the new one is more stable. so LGTM.

PR #952, opened 2018-11-12
Fix TextClassifier methods and add setEvaluateStatus

hkvision:
Previously, `self.model.compile/fit/predict`, this will cause an error if the TextClassifier is directly loaded instead of initialized. Wrap the method from Scala side in this PR.

Add more unit tests for predict after model loading.

Add `setEvaluateStatus()` in KerasNet/ZooModel, which will be clearer than the previous `evaluate()` and also for Python use. Added unit tests.

### DIFF
--- a/pyzoo/test/zoo/feature/text/test_text_set.py
+++ b/pyzoo/test/zoo/feature/text/test_text_set.py
@@ -17,10 +17,12 @@
 import pytest
 import shutil
 
+from bigdl.optim.optimizer import SGD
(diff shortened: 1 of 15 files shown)


hkvision:
Rerun TextClassification examples (Python and Scala) manually.

hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/1354/

PR #969, opened 2018-11-23
Change the order of shape sequence and word2idx

hkvision:
First transform from word2index. Those words not in wordDict will be aborted.
Then shape sequence on indices.


hkvision:
jenkins passed http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/1375/

PR #971, opened 2018-11-27
simplify predictor model shutdown [#964 follow up]

yangw1234:


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/Predictor.scala
```diff
@@ -170,14 +136,16 @@ object Predictor {
...
       val localToBatch = toBatchBroad.value._1.cloneTransformer()
-      val batchedIter = partition.grouped(localBatchPerPartition)
-      def shutdownHook() = {
-        localModel.release()
-      }
-      FlatMapIteratorWithShutdown(batchedIter, shutdownHook) { imageFeatures =>
-        Predictor.predictImageBatch[T](localModel, imageFeatures, outputLayer, predictKey,
-          localToBatch, shareBuffer)
-        imageFeatures
+      val batchedIter = partition.grouped(localBatchPerPartition) ++ Array(null)
```
jason-dai: this is guaranteed to be append, yes?

yangw1234: yes, it is append

PR #974, opened 2018-11-29
add expand

dding3:
Seprate expand part from https://github.com/intel-analytics/analytics-zoo/pull/966/


dding3:
CC @zhichao-li

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/InternalExpandSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+import com.intel.analytics.bigdl.utils.Shape
+import com.intel.analytics.zoo.pipeline.api.keras.layers.internal.InternalExpand
+import com.intel.analytics.zoo.pipeline.api.keras.serializer.ModuleSerializationTest
+
+
+class InternalExpandSpec extends KerasBaseSpec {
+  "InternalExpand" should "generate correct output" in {
+    val tgtSizes = Array(5, 4, 3)
+    val layer = InternalExpand[Float](tgtSizes)
+    val input = Tensor[Float](5, 4, 1).rand()
```
zhichao-li: - add a unitest for `includeBatch=true`? i.e `tgtSizes = Array(5, 4, 3), input = Tensor[Float](1, 4, 3).rand()`
- Not sure if it's good that the dimension of `tgtSizes` doesn't respect the value of `includedBatch`, any comments?

dding3: Thanks for your suggestions. Added the unit test with `includeBatch=true`
`includeBatch`  is used to config if we need expand the batch dim, as when we build the model, the batch dim is -1. Rename it to `expandBatchDim` to bring less confusing.

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/InternalExpandSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.utils.Shape
+import com.intel.analytics.zoo.pipeline.api.keras.layers.internal.InternalExpand
+import com.intel.analytics.zoo.pipeline.api.keras.serializer.ModuleSerializationTest
+
+
+class InternalExpandSpec extends KerasBaseSpec {
+  "InternalExpand" should "generate correct output" in {
+    val tgtSizes = Array(5, 4, 3)
```
zhichao-li: What if user input `val tgtSizes = Array(50000, 4, 3) val input = Tensor[Float](5, 4, 1).rand()`  when `includedBatch=false` ? Wondering should we change the tgtSizes to Array(4, 3) in this case ?

dding3:
Synced with @zhichao-li offline and we agreed to use -1 to mask dim which we don't want to expand.

zhichao-li:
LGTM

dding3:
Jekins pass. http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1593/

Thanks for the review.

PR #978, opened 2018-11-30
[ONNX] concat （axis= 0）

Deliadd:


Deliadd:
Jenkins passed. ID is 1427.

pyzoo/zoo/pipeline/api/onnx/mapper/concat.py
```diff
@@ -23,12 +23,8 @@ def __init__(self, node, _params, _all_tensors):
         super(ConcatMapper, self).__init__(node, _params, _all_tensors)
 
     def _to_tensor(self):
-        axis = 0
+        axis = -1
```
zhichao-li: the default axis is 0?

zhichao-li: assert 'axis' in self.onnx_attr, "axis is a required attribute"

Deliadd:
Jenkins passed. ID is 1429

PR #983, opened 2018-12-04
[WIP][ONNX] Split 

Deliadd:
unable to support list of output


Deliadd:
Currently, multiple outputs are not supported.

PR #985, opened 2018-12-05
[ONNX] ReduceMean operator

Deliadd:


pyzoo/test/zoo/pipeline/onnx/test_model_loading.py
```diff
@@ -1050,3 +1050,53 @@ def test_slice3_default_axes(self):
...
+        class ReduceMean(torch.nn.Module):
+            def __init__(self, *parameter):
+                super(ReduceMean, self).__init__()
+                self.dim = parameter[0]
+                self.keepdim = parameter[1]
+
+            def forward(self, x):
+                return torch.mean(x, dim=self.dim, keepdim=self.keepdim)
+
+        pytorch_model = ReduceMean(1, True)
```
zhichao-li: What if we put `False` here? can we support that?

Deliadd:
Jenkins passed. ID is 1440.

pyzoo/test/zoo/pipeline/onnx/test_model_loading.py
Deliadd: Yes, we support both True and False.

PR #1012, opened 2018-12-24
Add keras support doc, test and evaluation example

yangw1234:

PR #1016, opened 2018-12-25
[WIP] nlp-architect intent extraction using TFDataSet and TFOptimizer

hkvision:
Due to some error, only enable the intent extraction part of MultiTaskIntentModel with the following modifications / constraints:
- Use keras (2.1.6) instead of tf.keras
- Use zero-based label instead of one-hot especially for validation
- nlp-architect only supports Python 3.5+


yangw1234:
#1012 has fixed those error. I tested the full model without metrics, it works fine

hkvision:
@yangw1234 Wow! Great! Thanks!

hkvision:
Close this PR first. Will open another one to update the code.

PR #1028, opened 2018-12-28
add keras docs and fix GaussianSampler newly

yunyanliluo:


yunyanliluo:
fix GaussianSampler from yesterday's bug

add keras docs:

./GlobalMaxPooling2D.scala
./SpatialDropout2D.scala
./GaussianSampler.scala
./Exp.scala
./Square.scala
./AtrousConvolution1D.scala
./BatchNormalization.scala
./UpSampling3D.scala

yunyanliluo:
jenkins has passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/1582/

hkvision:
@hhbyyh Could you please have a review?

hkvision:
@yangw1234 Please have a look at GaussianSampler. Especially, how many inputs to this layer and also is the Keras wrapping correct.

hkvision:
@yunyanliluo Then could you please test the scala code, modify and confirm this?
We also need to specify the input size in inputShape

hkvision:
Also @yangw1234 is it ok to use KerasLayerWrapper for this layer? Seems it won't check the input size, and if the input shape is not correct, it may throw some error that is not obvious.

yangw1234:
@hkvision not really sure about the Wrapper. GaussianSampler take two tensors of the same size and output a tensor of that size. It seems quite straightforward to do the check.

hkvision:
@yangw1234 Then we can just make it extend KerasLayer and override computeOutputShape instead of using KerasLayerWrapper? Maybe we can do this in a later stage if we have labor.

docs/docs/KerasStyleAPIGuide/Layers/core.md
```diff
@@ -1397,3 +1397,367 @@ array([[[ 0.06560206,  0.38629526, -0.23159817],
...
+GaussianSampler(inputShape = null)
+```
+**Python:**
+```python
+GaussianSampler(input_shape=None, name=None)
+```
+
+**Parameters:**
+
+* `inputShape`: Only need to specify this argument when you use this layer as the first layer of a model. For Scala API, it should be a [`MultiShape`](../keras-api-scala/#multishape) object that consists of two identical SingleShape. For Python API, it should be a list of two identical shape tuple. Batch dimension should be excluded.
```
hkvision: is there #multishape section in the page?

hkvision:
Since there is code change, you need to run a jenkins job.

docs/docs/KerasStyleAPIGuide/Layers/core.md
hkvision: change back to #shape

yunyanliluo:
I run a jenkins job successfully after code change, the link is http://10.239.47.211:18888/job/ZOO-PR-Validation/1637/

PR #1059, opened 2019-01-10
add play-scala serve example for recommendation model inference

tmacraft:


songhappy:
@glorysdj @tmacraft, let's have a chat first

tmacraft:
Sure, let me know when you want to talk.

Thanks

Lu

apps/web-service-sample/recommendation-inference/src/main/play-scala-inference/app/controllers/ModelController.scala
```diff
@@ -0,0 +1,122 @@
...
+
+          val predict = rnnParams.recModel match {
+            case Some(_) =>
+              Map(
+                "0" -> Map(
+                  "atc" -> "true",
+                  "heading" -> "Office Depot top sellers",
+                  "name" -> "content2_json",
+                  "type" -> "json",
+                  "sku" -> LocalPredictor(rnnParams.recModel.get).predict(inputSample)
```
jenniew: Can we persist local predictor to avoid creating this object and clone model every time if the model is the same

apps/web-service-sample/recommendation-inference/src/main/play-scala-inference/app/controllers/UpdateModelController.scala
```diff
@@ -0,0 +1,56 @@
...
+class UpdateModelController @Inject()(cc: ControllerComponents) extends AbstractController(cc) with LoadModel {
+
+  /**
+    * Create an Action to return @ModelController status.
+    * The configuration in the `routes` file means that this method
+    * will be called when the application receives a `POST` request with
+    * a path of `/update`.
+    */
+
+  def update: Action[JsValue] = Action(parse.json) { request =>
```
jenniew: When to call update to refresh? should the model be refreshed after each click?

tmacraft: Hi Jennie, when to update is determined by the user. Whenever user has a new model ready to deploy, user can either manually call this update api or automate this process by using something like a cron scheduled job.

apps/web-service-sample/recommendation-inference/src/main/play-scala-inference/app/controllers/ModelController.scala
tmacraft: Thanks for pointing it out! I have fixed it with a new commit. Now I can see significant improvement with inference time versus previous version:)

songhappy:
To give a clear example, a lot of files should be deleted and code to be reorganized.
1. move play-scala-infernce to web-service-sample as a separate module.
2. only one or two controllers should stay. 
3. delete utilities and move the file into models
4. only application.conf and routes should stay.
5. keep README.md, add scripts of entrypoint.sh and explain in README.md, samething for dockerfile. 
6. delete project
7. delete target
8. delete tmp
9. delete Jenkinsfile, Dockfile, .gitignore
10. call we from wechat if confused.

PR #1061, opened 2019-01-11
Support save/load seq2seq related layers

dding3:
1. Support save/load seq2seq related layers
2. Add related ut


pyzoo/zoo/models/seq2seq/seq2seq.py
```diff
@@ -183,3 +183,27 @@ def __init__(self, encoder, decoder, input_shape, output_shape, bridge=None,
...
+
+        # Arguments
+        path: The path for the pre-defined model.
+              Local file system, HDFS and Amazon S3 are supported.
+              HDFS path should be like 'hdfs://[host]:[port]/xxx'.
+              Amazon S3 path should be like 's3a://bucket/xxx'.
+        weight_path: The path for pre-trained weights if any. Default is None.
+        """
+        jmodel = callBigDlFunc(bigdl_type, "loadSeq2seq", path, weight_path)
+        model = ZooModel._do_load(jmodel, bigdl_type)
```
zhichao-li: not sure if we can use a unified loading method here rather than using `loadseq2seq`?

zhichao-li: I guess it's not an issue specified to this PR as the other loading methods are all written like these instead of `ZooModel.load(xxx)`

zoo/src/main/scala/com/intel/analytics/zoo/models/seq2seq/RNNDecoder.scala
```diff
@@ -148,4 +156,57 @@ object RNNDecoder {
...
   }
+
+  override def doLoadModule[T: ClassTag](context : DeserializeContext)
+    (implicit ev: TensorNumeric[T]) : AbstractModule[Activity, Activity, T] = {
+
+    val attrMap = context.bigdlModule.getAttrMap
+
+    val rnns = DataConverter.getAttributeValue(context, attrMap.get("rnns")).
+      asInstanceOf[Array[AbstractModule[_, _, T]]].map(_.asInstanceOf[Recurrent[T]])
+    rnns.map(_.labor = null)
```
zhichao-li: If we discard the `labor` here, would it be possible that the original "weights" is loss? or somehow it magically OK as you can pass the unitest?

pyzoo/zoo/models/seq2seq/seq2seq.py
hkvision: use ZooModel.load can only load as a ZooModel, but we want a Seq2seq object, thus every model has its load method to load the type of the exact model.

zoo/src/main/scala/com/intel/analytics/zoo/models/seq2seq/RNNDecoder.scala
zhichao-li: Actually I'm confuse here.  Logically, I was supposing that the `weights` is brand new here as the the `decoder` is rebuilt via `decoder.build(KerasUtils.addBatch(shape))` which construed by `embedding` and `rnn` and the `weights`(labor) is re-initiated as the labor is discard.

dding3: Yes, `decoder` is rebuilt here using `embedding` and `rnn`, and `embedding` and `rnn` are not brand new. They are loaded from serialized layers so `weights` are the same with layers before serialization which makes `decoder` weights also correct?

dding3: For `KerasLayer`, not only it will serialize the constructor params, but also will serialize `subModules` including torch layers used to build keras layer. https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala#L417. So it's able to rebuilt the layer with the same `weights`

dding3:
Jekins passed. http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1730/

PR #1065, opened 2019-01-11
add keras learning phase support

yangw1234:
@hkvision This should fix the problem you encountered. Could you help testing this?

1. fix keras mtl model does not converge
2. support setting sessionConfig in python

If performance is batch size sensitive, using
```
TFOptimizer.from_keras(model, tfdataset, config=tf.ConfigProto())
```
and setting spark.master to `local[1]` would be faster. In this case, we are using a single tensorflow session to run the whole batch.


pyzoo/zoo/pipeline/api/net.py
```diff
@@ -362,7 +366,8 @@ class TFOptimizer:
 
     def __init__(self, loss, optim_method, sess=None, dataset=None, inputs=None,
                  grads=None, variables=None, graph=None,
-                 val_outputs=None, val_labels=None, val_method=None, val_split=0.0):
+                 val_outputs=None, val_labels=None, val_method=None, val_split=0.0,
+                 tensors_with_value=None, session_config=None):
```
jason-dai: need to update the python doc and documents

PR #1076, opened 2019-01-16
load keras with nnframes

hhbyyh:
right now loading NNModel with keras model will report an error, since Net registered some serializer and we need to include that.

Also added a unit test for loading test.


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/nnframes/NNEstimator.scala
```diff
@@ -735,6 +736,7 @@ object NNModel extends MLReadable[NNModel[_]] {
   }
 
   private[nnframes] def getMetaAndModel(path: String, sc: SparkContext) = {
+    Net // this is necessary to load Net and register the serializer
```
dding3: If we call `NNModel.load` twice, not sure if it will raise exception about register more than once.

hhbyyh: Thanks for comments. Every time we call `Net.load`, the initialize will be executed. I don't think multiple invoke will bring an exception, but I can update the unit test to cover it.

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1751/ passed.

PR #1097, opened 2019-01-22
Fix default value of zeroBasedId in Embedding

hkvision:
An error brought by PR #1062 
In the factory method Scala, default should be true. @dding3


hkvision:
Scala tests passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/1795/
Thanks for your review @dding3

PR #1107, opened 2019-01-23
fix app test1

yangw1234:


yangw1234:
pass the case: merge it first
http://10.239.47.211:18888/view/ZOO-NB/job/ZOO-NB-Python-AppTests-part1/111/consoleFull

PR #1109, opened 2019-01-23
fix bug for anomaly_detection

glorysdj:
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/1109)


hkvision:
I have fixed in #1103

PR #1110, opened 2019-01-23
Add a hooker to offset the pmem

zhichao-li:


zhichao-li:
We might still need to figure out how to offset the allocation if the process is terminate by "kill -9" rather than restart the system.

PR #1111, opened 2019-01-23
Fix pip

yangw1234:


yangw1234:
pip app test pass:
http://10.239.47.211:18888/view/ZOO-NB/job/ZOO-NB-Pip-Apps-Integration/98/console

PR #1127, opened 2019-01-25
fix inceptionV1 converge slow

qiuxin2012:


zhichao-li:
In theory, there's no difference before and after the changes as we only use " isInOrder: Boolean = false, groupSize: Int = 1" for now. Anyway, let's check the result first :)

qiuxin2012:
It's a bug of bigdl, see [bigdl#2734](https://github.com/intel-analytics/BigDL/issues/2734)

qiuxin2012:
close this one.

PR #1137, opened 2019-02-11
add recommenders in Keras style

songhappy:
Add ncf and widendeep in keras apis
Add KerasZooModel to use apis of KerasNet


zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
```diff
@@ -58,135 +61,140 @@ case class ColumnFeatureInfo(wideBaseCols: Array[String] = Array[String](),
...
-        throw new IllegalArgumentException("unknown type")
+    (1 to hiddenLayers.length - 1).map(i => {
+      LinearTail = Dense(hiddenLayers(i), activation = "relu").inputs(LinearTail)
+    })
+
+    Dense(numClasses).inputs(LinearTail)
+
+  }
+
+  def deepMerge(inputInd: ModuleNode[T], inputEmb: ModuleNode[T], inputCunt: ModuleNode[T]) = {
```
tmacraft: typo, inputCunt to inputCount

zoo/src/main/scala/com/intel/analytics/zoo/examples/recommendation/NeuralCFexample.scala
```diff
@@ -96,20 +97,15 @@ object NeuralCFexample {
...
     val optimMethod = new Adam[Float](
       learningRate = param.learningRate,
       learningRateDecay = param.learningRateDecay)
 
-    optimizer
-      .setOptimMethod(optimMethod)
-      .setEndWhen(Trigger.maxEpoch(param.nEpochs))
-      .optimize()
+    ncf.compile(optimizer = optimMethod,
+      loss = ClassNLLCriterion[Float](),
```
zhichao-li: Maybe we should change this to `SparseCategoricalCrossEntropy` as `ClassNLLCriterion` is a convention from Torch.

zoo/src/test/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeepSpec.scala
```diff
@@ -61,121 +61,89 @@ class WideAndDeepSpec extends ZooSpecHelper {
...
 
-    val data = datain
-      .rdd.map(r => Utils.getDeepTensor(r, columnInfo))
-    data.map{ input =>
-      val feature : Tensor[Float] = input.reshape(Array(1, input.size(1)))
-      val output = model.forward(feature)
-      val gradInput = model.backward(feature, output)
-    }.count()
-  }
+    val optimizer = Optimizer(
```
zhichao-li: still using Optimizer?

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/NeuralCF.scala
```diff
@@ -52,52 +55,63 @@ class NeuralCF[T: ClassTag](
...
-    val mlpEmbeddedLayer = Concat[T](2)
-      .add(Sequential[T]().add(Select(2, 1)).add(mlpUserTable))
-      .add(Sequential[T]().add(Select(2, 2)).add(mlpItemTable))
-    val mlpModel = Sequential[T]()
-    mlpModel.add(mlpEmbeddedLayer)
-    val linear1 = Linear[T](itemEmbed + userEmbed, hiddenLayers(0))
-    mlpModel.add(linear1).add(ReLU())
+    val input = Input[T](inputShape = Shape(2))
+
+    val userSelect = Select[T](1, 0).inputs(input)
```
zhichao-li: why do we still need to select the input here? should we just use a separated input?

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/NeuralCF.scala
```diff
@@ -52,52 +55,63 @@ class NeuralCF[T: ClassTag](
...
+    val userFlat = Flatten().inputs(userSelect)
+    val itemFlat = Flatten().inputs(itemSelect)
+
+    val mlpUserTable = Embedding[T](userCount + 1, userEmbed)
+    val mlpItemTable = Embedding[T](itemCount + 1, itemEmbed)
+
+    val mlpUserLatent: ModuleNode[T] = Flatten().inputs(mlpUserTable.inputs(userFlat))
+    val mlpItemLatent: ModuleNode[T] = Flatten().inputs(mlpItemTable.inputs(itemFlat))
+
+    mlpUserTable.setWeightsBias(Array(Tensor[T](userCount + 1, userEmbed).randn(0, 0.1)))
```
zhichao-li: What's the reason for setting the parameter in such range or is there any available initializer can be used in `Embedding`?

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
```diff
@@ -58,135 +61,140 @@ case class ColumnFeatureInfo(wideBaseCols: Array[String] = Array[String](),
...
 
-    // add hidden layers
-    deepModel.add(deepColumn)
-      .add(Linear[T](indicatorWidth + embedOutputDimSum + continuousWidth, hiddenLayers(0)))
-      .add(ReLU())
-    for (i <- 1 to hiddenLayers.length - 1) {
-      deepModel.add(Linear[T](hiddenLayers(i - 1), hiddenLayers(i))).add(ReLU())
-    }
-    deepModel.add(Linear[T](hiddenLayers.last, numClasses))
+    var LinearTail = Dense(hiddenLayers(0), activation = "relu").inputs(merge1)
```
zhichao-li: `linearTail`?

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
zhichao-li: Maybe we can just merge what need to be merged here as that's what this method name implies?
```
    val merge1 = Merge.merge(mergeList, "concat")
```

hkvision:
Can we make our KerasNet a trait so that it can be extended by both Sequential/Model and some ZooModels? In this case, we only need to maintain compile, fit methods in one place @zhichao-li

zhichao-li:
If we take a step back, we can find that the problem is more about `ZooModel` is not a "Model" as the name implies or the same way as `Seq` or `Model` does which should inherit from `KerasNet`.

hkvision:
But actually KerasNet, Seq, Model are not Model as well, they are also just wrappers, which extends KerasLayer and it is the labor that does the work. The implementation of ZooModel follows KerasLayer.

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/NeuralCF.scala
songhappy: the input is one tensor which contains two elements, so select is still needed.

songhappy:
By design, ZooModel  supports models of both torch style and Keras style, it should be used in optimizer and nnestimater. KerasNet defines "fit, compile " etc. To leverage both ZooModel and Keras style, KerasZooModel is defined to bridge keras apis like fit compile for ZooModel. 
There are two possible ways to do this:
1. define a trait of KerasZooModel, then we have to pass the model, set up the model, this process is tedious.  
2. define an abstract class of KerasZooModel which is used in here.

zoo/src/main/scala/com/intel/analytics/zoo/models/common/KerasZooModel.scala
```diff
@@ -0,0 +1,54 @@
...
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.zoo.pipeline.api.keras.models.KerasNet
+import org.apache.spark.rdd.RDD
+
+import scala.reflect.ClassTag
+
+abstract class KerasZooModel[A <: Activity : ClassTag, B <: Activity : ClassTag, T: ClassTag]
+(implicit ev: TensorNumeric[T]) extends ZooModel[A, B, T] {
+
+  def compile(optimizer: OptimMethod[T],
```
shane-huang: Since this is an important interface/base API, we'd better add more api docs to this.

zoo/src/main/scala/com/intel/analytics/zoo/models/common/KerasZooModel.scala
```diff
@@ -0,0 +1,54 @@
...
+import com.intel.analytics.bigdl.dataset.Sample
+import com.intel.analytics.bigdl.nn.abstractnn.Activity
+import com.intel.analytics.bigdl.optim.{OptimMethod, ValidationMethod, ValidationResult}
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.zoo.pipeline.api.keras.models.KerasNet
+import org.apache.spark.rdd.RDD
+
+import scala.reflect.ClassTag
+
+abstract class KerasZooModel[A <: Activity : ClassTag, B <: Activity : ClassTag, T: ClassTag]
```
shane-huang: Since this is an important interface/base API, we'd better add more api docs to this file

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/DeepFM.scala
```diff
@@ -0,0 +1,130 @@
...
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.nn._
+import com.intel.analytics.bigdl.nn.abstractnn.AbstractModule
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.zoo.models.common.ZooModel
+
+import scala.reflect.ClassTag
+
+/**
+ * The neural collaborative filtering model used for recommendation.
```
shane-huang: I think this is Factorization Machine? Why api doc still says NCF? And do we need some examples or more documentations on deepFM ? e.g. what paper or implementation we refers to?

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/NeuralCF.scala
songhappy: changed

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
songhappy: Merge.merge is the right way here, Merge.apply does not take ModuleNode

zoo/src/main/scala/com/intel/analytics/zoo/models/common/KerasZooModel.scala
songhappy: This needs a bit more discussion, I will add more apis or anybody can add more apis when needed after we make the decision to go with it

zhichao-li:
It looks good to me for the model definition. Just not sure if we need to maintain that much model alternatives in one file or shall we just focus on a clean and optimized W&D.

zoo/src/main/scala/com/intel/analytics/zoo/examples/recommendation/NeuralCFexample.scala
```diff
@@ -96,20 +96,16 @@ object NeuralCFexample {
...
     val optimMethod = new Adam[Float](
       learningRate = param.learningRate,
       learningRateDecay = param.learningRateDecay)
 
-    optimizer
-      .setOptimMethod(optimMethod)
-      .setEndWhen(Trigger.maxEpoch(param.nEpochs))
-      .optimize()
+    ncf.compile(optimizer = optimMethod,
+      loss = SparseCategoricalCrossEntropy[Float](zeroBasedLabel = false, logProbAsInput = true),
```
hkvision: Can we use 'softmax' as the last layer instead of 'log_softmax' and set logProbAsInput=false (by default), which is more commonly used by keras

jason-dai:
> If we take a step back, we can find that the problem is more about `ZooModel` is not a "Model" as the name implies or the same way as `Seq` or `Model` does which should inherit from `KerasNet`.

`ZooModel` is not necessary a Keras style model, so a `KerasZooModel` makes sense.

hkvision:
Is this PR ready so that KNRM can also use these methods?

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/Utils.scala
```diff
@@ -125,9 +125,63 @@ object Utils {
...
-    Tensor.sparse(Array(indices), values, shape)
+    val x: Array[Array[Int]] = Array(indices)
+
+    val out = Tensor.sparse(Array(indices), values, shape)
+
+    out
   }
 
+  def getDeepTensors(r: Row, columnInfo: ColumnFeatureInfo): Array[Tensor[Float]] = {
+
```
shane-huang: If this is an external interface (which user needs to use), we need more code docs to explain what it does. If this is an internal interface, we may change add "_" in front of the func name.

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
```diff
@@ -77,98 +95,106 @@ case class ColumnFeatureInfo(wideBaseCols: Array[String] = Array[String](),
...
+      linear = Dense(hiddenLayers(i), activation = "relu").inputs(linear)
+    })
+
+    Dense(numClasses).inputs(linear)
+
+  }
+
+  def deepMerge(ind: ModuleNode[T], emb: ModuleNode[T], cont: ModuleNode[T]) :
+  (Array[ModuleNode[T]], List[ModuleNode[T]]) = {
+
```
shane-huang: Is it a public interface?

shane-huang: if public interface we need more code docs to explain what it does and the paramters

pyzoo/test/zoo/models/recommendation/test_wideanddeep.py
```diff
@@ -54,38 +52,45 @@ def setup_method(self, method):
             continuous_cols=["age"])
 
     def test_wide_forward_backward(self):
-        column_info = ColumnFeatureInfo(
```
hkvision: Is this case class still used? If not, maybe we can remove it from wide_and_deep.py

hkvision:
Since you have modified quite much related to WideAndDeep, add a description for those changes?

hkvision:
Also, in a following PR, don't forget to update website docs if API has changed.

pyzoo/zoo/models/common/zoo_model.py
```diff
@@ -98,7 +98,7 @@ class KerasZooModel(ZooModel):
     """
     # For the following method, please see documentation of KerasNet for details
     def compile(self, optimizer, loss, metrics=None):
-        self.model.compile(optimizer, loss, metrics=None)
+        self.model.compile(optimizer, loss, metrics)
```
hkvision: I think this PR is too large to merge in a rush before release 0.5.0
What about separating these bug fixing modifications in another PR to merge it first?

songhappy:
will update doc in next PR
Jenkins pass
http://172.168.2.101:8080/job/ZOO-PR-Validation/2275/

pyzoo/test/zoo/models/recommendation/test_wideanddeep.py
songhappy: in use

pyzoo/test/zoo/models/recommendation/test_neuralcf.py
```diff
@@ -31,7 +30,6 @@ class TestNeuralCF(ZooTestCase):
 
     def test_forward_backward_without_mf(self):
         model = NeuralCF(30, 30, 2, include_mf=False)
-        model.summary()
```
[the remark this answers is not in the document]
songhappy: there are other places to have summary()

apps/recommendation-wide-n-deep/README.md
```diff
@@ -6,7 +6,7 @@ This notebook demonstrates how to build a neural network recommendation system (
 * Python 2.7/3.5/3.6
 * JDK 8
 * Spark 1.6.0/2.1.1/2.1.2/2.2.0(This version needs to be same with the version you use to build Analytics Zoo)
-* Analytics Zoo 0.3.0
+* Analytics Zoo 0.4.0
```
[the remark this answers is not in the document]
songhappy: this will be updated

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
songhappy: will update it as protected in next PR

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/Utils.scala
songhappy: it is the same as the getDeepTensor

pyzoo/zoo/models/recommendation/wide_and_deep.py
```diff
@@ -101,22 +102,120 @@ class WideAndDeep(Recommender):
...
+        linear = Dense(self.hidden_layers[0], activation="relu")(merged)
+
+        for ilayer in range(1, len(self.hidden_layers)):
+            linear_mid = Dense(self.hidden_layers[ilayer], activation="relu")(linear)
+            linear = linear_mid
+        last = Dense(self.class_num, activation="relu")(linear)
+        return last
+
+    def deep_merge(self, input_ind, input_emb, input_cont):
+        embed_width = 0
```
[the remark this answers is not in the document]
songhappy: doc will go another PR

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
```diff
@@ -77,98 +95,106 @@ case class ColumnFeatureInfo(wideBaseCols: Array[String] = Array[String](),
...
 
   override def buildModel(): AbstractModule[Tensor[T], Tensor[T], T] = {
 
-    val fullModel = Sequential[T]()
+    val inputWide: ModuleNode[T] = Input(inputShape = Shape(wideBaseDims.sum + wideCrossDims.sum))
+    val inputInd: ModuleNode[T] = Input(inputShape = Shape(indicatorDims.sum))
+    val inputEmb = Input(inputShape = Shape(embedInDims.length))
+    val inputCon = Input(inputShape = Shape(continuousCols.length))
 
-    val wideModel = Sequential().setName("widePart")
```
qiuxin2012: Please retain this `.setName("widePart")`, or will throw an exception in Census example.

zoo/src/main/scala/com/intel/analytics/zoo/models/recommendation/WideAndDeep.scala
```diff
@@ -77,98 +95,106 @@ case class ColumnFeatureInfo(wideBaseCols: Array[String] = Array[String](),
...
-      wideBaseDims.sum + wideCrossDims.sum, numClasses).setInitMethod(Zeros))
-      .add(CAdd(Array(numClasses)))
+    modelType match {
+      case "wide" =>
+        val out = Activation("softmax").inputs(wideLinear)
+        val model: Model[T] = Model(Array(inputWide), out)
+          model.asInstanceOf[AbstractModule[Tensor[T], Tensor[T], T]]
 
-    // deep model
-    val deepModel = Sequential[T]().setName("deepPart")
```
qiuxin2012: Please retain this `.setName("deepPart")`, or will throw an exception in Census example.

PR #1138, opened 2019-02-12
change input name to align with tensorflow object detection model convention

yangw1234:
…vention


yangw1234:
jenkins pass http://172.168.2.101:8080/job/ZOO-PR-Validation-Other/1828/

PR #1139, opened 2019-02-12
Add Keras ConvLSTM3D

### LINKED ISSUE #724: Missing Convlstm related keras layer

hkvision:
Per user's request. #724 

Only channel_first is supported.
Cubic kernel with equal strides.
Same padding. (valid padding has error for both 2D and 3D)


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/1843/

PR #1094, opened 2019-01-21
add tfpark model api

yangw1234:

PR #1166, opened 2019-02-22
[WIP] Support tf.keras.metrics

yangw1234:
This PR attempts to support tf.keras.metrics and custom designed metrics.
 
In tf.keras, there are two kinds of metrics, namely **stateful metrics** and **non-stateful metrics** (the official keras seems only support this).

1. A **non-stateful metrics** is basically a function (variable free tf graph) that maps two tensor to a single scalar and the final metric across all batches is simply averaged. 

    In this case, we can simple fetch the final metric tensor and average them.

2. A **stateful metrics** is an object (tf graph contains variable) that can keep updating this state (variables) using  two tensor inputs and get the final metrics using the variables after iterating all the batches. 

    In this case, we can first compute all the predictions and labels (presumably not very large) and collect them to driver, then iterate through those to calculate the final metric.


yangw1234:
@jason-dai @hkvision any advises to this approach?

jason-dai:
I think we should first implement the Keras model support in tfpark (https://github.com/intel-analytics/analytics-zoo/pull/1094) before adding the metrics support

yangw1234:
done in other pr

PR #1182, opened 2019-02-28
Update tfpark API

hkvision:
- Modify predict result if input is ndarray to keep aligned with Keras,
Output now should be (batch, ...) [single output] or [(batch, ...), (batch, ...), ...] [multiple outputs]
- Some minor modifications: change get_spark_context() in bigdl to getOrCreateSparkContext() in Zoo.

Will add some ut to cover multiple output case.

### DIFF
--- a/pyzoo/zoo/pipeline/api/net.py
+++ b/pyzoo/zoo/pipeline/api/net.py
@@ -18,7 +18,6 @@ import sys
 import tempfile
 import warnings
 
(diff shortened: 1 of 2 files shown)


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/1927/

yangw1234:
LGTM

PR #1190, opened 2019-03-04
 change tensorflow runtime to tf v1.13.1 with mkl

yangw1234:


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFNet.scala
```diff
@@ -578,6 +578,89 @@ class TFNet(private val graphDef: TFGraphHolder,
 
 object TFNet {
 
+  val iomp5name = "libiomp5.so"
```
jason-dai: shall we move this to core?

yangw1234: moved to core

pyzoo/dev/prepare_env.sh
```diff
@@ -37,3 +37,8 @@ echo "ANALYTICS_ZOO_CLASSPATH": $ANALYTICS_ZOO_CLASSPATH
 
 export BIGDL_CLASSPATH=$ANALYTICS_ZOO_CLASSPATH
 echo "BIGDL_CLASSPATH": $BIGDL_CLASSPATH
+
+export KMP_BLOCKTIME=0
```
yangw1234: tried to use jni to call setenv in c++ from java side, but it does not work, not sure about why

yangw1234: anywhere else I should export those environment variables, like in pip install mode? @hkvision

hkvision: can we just export these in __init__ of zoo?

yangw1234: they need to be set before jvm are started

jason-dai: where is this script used? does it work in the cluster? how is this handled in BigDL MKLDNN support?

yangw1234: the script is used in python testing
In BigDL, it is handled by calling a few native methods like omp_set_num_threads, but in our case tensorflow will override that

zoo/src/main/resources/spark-analytics-zoo.conf
```diff
@@ -31,4 +31,8 @@ spark.shuffle.reduceLocality.enabled                false
 spark.shuffle.blockTransferService                  nio
 spark.scheduler.minRegisteredResourcesRatio         1.0
 spark.speculation                                   false
-spark.serializer                                    org.apache.spark.serializer.JavaSerializer
\ No newline at end of file
+spark.serializer                                    org.apache.spark.serializer.JavaSerializer
+spark.executorEnv.KMP_BLOCKTIME                     0
+spark.executorEnv.KMP_AFFINITY                      granularity=fine,verbose,compact,1,0
+spark.executorEnv.KMP_SETTINGS                      1
+spark.executorEnv.OMP_NUM_THREADS                   1
```
yangw1234: for cluster to work @jason-dai

pyzoo/dev/prepare_env.sh
yangw1234: in cluster, the environment varible is exported using analytics-zoo-spark.conf

yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1942/

PR #1191, opened 2019-03-05
Dataset imdb

Litchilitchy:
call bigdl method to download imdb, (not copying keras file), add parameter support 

(the old pull request did not separate one old commit, this is the new one)


Litchilitchy:
re-implementation finished, datasets are all cached into `/tmp/.zoo/dataset/`

jason-dai:
LGTM

PR #1192, opened 2019-03-05
Dataset mnist

Litchilitchy:
call bigdl method to download mnist dataset, this one is imported from bigdl

(the old pull request did not separate one old commit, this is the new one)


Litchilitchy:
this dataset occasionally cause error when calling `fit` method, debugging now

Litchilitchy:
fixed to run `fit`

PR #1198, opened 2019-03-09
support constant variable

dding3:
Currently only support scala version, will add python version after this PR is merged.

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/KerasParameter.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/KerasParameter.scala
@@ -17,10 +17,10 @@
 
 package com.intel.analytics.zoo.pipeline.api.autograd
 
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, Initializable, TensorModule}
+import com.intel.analytics.bigdl.nn.abstractnn._
 import com.intel.analytics.bigdl.nn.keras.KerasLayer
 import com.intel.analytics.bigdl.nn.tf.InternalWithoutInput
-import com.intel.analytics.bigdl.nn.{InitializationMethod, RandomUniform, VariableFormat, Input => TInput}
+import com.intel.analytics.bigdl.nn.{InitializationMethod, RandomUniform, VariableFormat}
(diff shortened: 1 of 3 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/KerasParameter.scala
```diff
@@ -158,3 +158,45 @@ private[zoo] class InternalParameter[T: ClassTag](
...
+
+  override def updateOutput(input: Tensor[T]): Tensor[T] = {
+    output = data
+    output
+  }
+
+  override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
+    gradInput = gradOutput
+    gradInput
+  }
```
jason-dai: Can we remove the copy in `InternalParameter.updateOutput` and `InternalParameter.updateGradInput` too?

dding3: Removed.

dding3:
jekins passed. http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/1960/

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/KerasParameter.scala
zhichao-li: Maybe we should just return null or 0 for the grad input. I recall that tf would return null in this case in respect of performance but most people would expect 0.

jason-dai: returning `gradOutput`, null or empty tensor is fine - it is not supposed to be used anyway, no?

zhichao-li: yes. In theory, null is ok as it's suppose to be a leaf.

dding3:
jekins pass. http://10.239.47.211:18888/job/ZOO-PR-Validation/1973/

zhichao-li:
LGTM

dding3:
Thanks for your review.

PR #1211, opened 2019-03-13
Keras-api evaluation method return modification

Litchilitchy:
fix evaluation return to correspond with keras, created a new method `processEvaluateResultRDD`, not removing the original method

add a new `python` package in test module, now python-keras api could directly be tested under `PythonZooKerasSpec`


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -408,13 +408,33 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
    * @param batchSize Number of samples per batch.
    */
+
+  /*
   def evaluate(
       x: RDD[Sample[T]],
       batchSize: Int)
       (implicit ev: TensorNumeric[T]): Array[(ValidationResult, ValidationMethod[T])] = {
     require(this.vMethods != null, "Evaluation metrics haven't been set yet")
     this.evaluate(x, this.vMethods, Some(batchSize))
```
hkvision: Can we directly add Loss to vMethods here and call this.evaluate only once?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -408,13 +408,33 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
   }
+  */
```
jason-dai: delete the codes if you don't need them

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -408,13 +408,33 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
+  def evaluate(
+      x: RDD[Sample[T]],
+      batchSize: Int)
+      (implicit ev: TensorNumeric[T]): Array[(ValidationResult, ValidationMethod[T])] = {
+    // require(this.vMethods != null, "Evaluation metrics haven't been set yet")
+
+    val loss = this.evaluate(x, Array(new Loss(this.criterion)), Some(batchSize))
```
jason-dai: just add loss to `this.vMethods` at compile time

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKeras.scala
```diff
@@ -147,12 +147,19 @@ class PythonZooKeras[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
     }.toList.asJava
   }
 
+  private def processEvaluateResultRDD(
```
jason-dai: just `processEvaluateResult`, not RDD

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKeras.scala
```diff
@@ -147,12 +147,19 @@ class PythonZooKeras[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+  }
+
   def zooEvaluate(
       module: KerasNet[T],
       x: JavaRDD[Sample],
-      batchSize: Int = 32): JList[EvaluatedResult] = {
+      batchSize: Int = 32): JList[Float] = {
     val resultArray = module.evaluate(toJSample(x), batchSize)
-    processEvaluateResult(resultArray)
+    processEvaluateResultRDD(resultArray)
```
jason-dai: we should do this for both scala and python version of `evaluate`

zhichao-li: You should change `processEvaluateResult` directly rather than creating a new one.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
zhichao-li: Make you condition logic more clear. In this case, what you want to do is to assemble the correct vMethods  i.e
```
this.vMethods = if xxx else xxx
return this.evaluate(vMethods)
```

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKerasSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.intel.analytics.bigdl.utils.Shape
+import com.intel.analytics.zoo.common.NNContext
+import com.intel.analytics.zoo.pipeline.api.keras.layers.{Dense, Flatten}
+import com.intel.analytics.zoo.pipeline.api.keras.models.Sequential
+import org.apache.spark.{SparkConf, SparkContext}
+import org.apache.spark.rdd.RDD
+import org.scalatest.{BeforeAndAfter, FlatSpec, Matchers}
+import com.intel.analytics.zoo.pipeline.api.keras.python.PythonZooKeras
+
+class PythonZooKerasSpec extends FlatSpec with Matchers with BeforeAndAfter {
```
hkvision: put it into `TrainingSpec` or `TopologySpec`?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: no, if we want to achieve we have to modify bigdl code

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: loss does not belong to subclass of ValidationMethods

Litchilitchy:
@zhichao-li @jason-dai @hkvision I think you prefer modifying instead of adding the new one. I would just do the modification and would not keep the old code

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKerasSpec.scala
Litchilitchy: this is the first time to call python api and I think it would be more convenient if we just want to test python api?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: @zhichao-li @hkvision and if we want to assemble them in once operation we have to make `loss` into `vMethod`? Is there any side-effect? I am not sure because `vMethod` is originally designed to take `ValidationMethod` subclass

Litchilitchy: @zhichao-li @jason-dai ok I think we could make it. Perhaps I remembered something wrong before

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: noted, we could make it

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKerasSpec.scala
hkvision: We don't need to test python in scala. Here we put unit test to cover the change in scala. And in python, you add another unit test probably in `test_simple_integration.py`.

Litchilitchy: this could skip the `make-dist.sh` process and I think this is the advantage?

zhichao-li: I think the more the better in respect of unitest if no side affect, but you should also add some python unitests to conver the python path.

hkvision: I mean no need to create a separate file for this ut? as the function generateData seems already in TrainingSpec and you can put this ut there for convenience.

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKerasSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+    val api = new PythonZooKeras[Float]()
+    val bigdlApi = sc.broadcast(new PythonBigDL[Float]())
+
+    /** python api require to take no type Sample
+      * and it takes JavaRDD as input
+      *
+      * use toPySample to convert to no type
+      * use toJavaRDD to convert to JavaRDD
+    **/
+    val jd = trainingData.map(j => bigdlApi.value.toPySample(j)).toJavaRDD()
```
hkvision: It will be better to cover the ut in Python in this case. PythonZooKeras is not supposed to be tested in Scala.

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKerasSpec.scala
Litchilitchy: does function `generateData` only exist in `TrainingSpec`?

Litchilitchy: @zhichao-li actually there is no new methods created in this pr. So hasn't any old test already exist for testing this pr?

Litchilitchy: @hkvision modified

PR #1213, opened 2019-03-13
Add sequence tagger in tfpark

hkvision:
- [x] Sequence Tagger
- [x] Inline doc
- [x] Unit tests

### DIFF
--- a/pyzoo/zoo/tfpark/text/pos_tagging.py
+++ b/pyzoo/zoo/tfpark/text/pos_tagging.py
@@ -1,0 +1,69 @@
+#
+# Copyright 2018 Analytics Zoo Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


pyzoo/zoo/tfpark/text/pos_tagging.py
```diff
@@ -0,0 +1,69 @@
...
+#
+
+from nlp_architect.models import chunker
+from zoo.tfpark.text import TextKerasModel
+
+
+class SequenceTagger(TextKerasModel):
+    """
+    The model used as POS-tagger and chunker for sentence tagging, which contains three
+    Bidirectional LSTM layers.
```
hkvision: This model is for both pos and chunk, if we want only pos, then we may need to take one output and construct a new model as nlp-architect does here: https://github.com/NervanaSystems/nlp-architect/blob/master/nlp_architect/models/chunker.py#L233

hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/1998/

PR #1221, opened 2019-03-15
add text_generation keras example

Bob-Chou:
an error when run model.fit:
: java.lang.IllegalArgumentException: Node 'gradients/lstm/while/Exit_2_grad/b_exit' expects to be colocated with unknown node 'lstm/while/Exit_2'


pyzoo/zoo/examples/tensorflow/tfpark/text_generation.ipynb
```diff
@@ -0,0 +1,1269 @@
...
+   "execution_count": 0,
+   "metadata": {
+    "colab": {},
+    "colab_type": "code",
+    "id": "UK-hmKjYVoll"
+   },
+   "outputs": [],
+   "source": [
+    "keras_model = KerasModel(model)\n",
+    "history = keras_model.fit(input_example, target_example, \n",
```
hkvision: In our implementation currently fit doesn't return anything, so remove `history = `

PR #1226, opened 2019-03-18
Tensorboard fix

Litchilitchy:


pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -74,6 +74,33 @@ def set_tensorboard(self, log_dir, app_name):
...
+                             self.value, tag, "Train")
+
+    def get_validation_summary(self, tag=None):
+        """
+        Get the scalar from model validation summary
+        Return 2-D array like object which could be converted
+        by nd.array()
+        # Arguments
+        tag: The string variable represents the scalar wanted
+        """
```
jason-dai: check tag?

Litchilitchy: @jason-dai fixed

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -74,6 +74,40 @@ def set_tensorboard(self, log_dir, app_name):
...
+    def get_train_summary(self, tag=None):
+        """
+        Get the scalar from model train summary
+        Return 2-D array like object which could be converted
+        by nd.array()
+        # Arguments
+        tag: The string variable represents the scalar wanted
+        """
+        # exception handle
+        if tag != "Loss" and tag != "LearningRate" and tag != "Throughput":
```
hkvision: convert to lower case and check?

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -74,6 +74,40 @@ def set_tensorboard(self, log_dir, app_name):
...
+                            + 'are supported in train summary')
+
+        return callBigDlFunc(self.bigdl_type, "zooGetScalarFromSummary",
+                             self.value, tag, "Train")
+
+    def get_validation_summary(self, tag=None):
+        """
+        Get the scalar from model validation summary
+        Return 2-D array like object which could be converted
+        by nd.array()
```
hkvision: nd.array or nd_array()?

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -74,6 +74,40 @@ def set_tensorboard(self, log_dir, app_name):
...
+        # Arguments
+        tag: The string variable represents the scalar wanted
+        """
+        validation_set = set(('AUC', 'Accuracy', 'BinaryAccuracy', 'CategoricalAccuracy',
+                              'HitRatio', 'Loss', 'MAE', 'NDCG', 'SparseCategoricalAccuracy',
+                              'TFValidationMethod', 'Top1Accuracy',
+                              'Top5Accuracy', 'TreeNNAccuracy'))
+        if tag not in validation_set:
+            raise TypeError('Only subclasses of ValidationMethod are supported,'
```
hkvision: ValidationMethod is not exposed in Python and thus this error msg may be confusing to users?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -184,12 +187,38 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
 
+  /**
+   * To get the scalar like "Loss", "LearningRate" from train summary
+   * Return is a Array of 3-tuples
+   *
+   * @param tag The string variable represents the parameter you want to return
+   *            supported tags are "LearningRate", "Loss", "Throughput"
+   */
+  def getTrainSummary(tag: String): Array[(Long, Float, Double)] = {
+    this.trainSummary.readScalar(tag)
```
hkvision: will Scala do the tag check?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -184,12 +187,38 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
+  }
+
+  /**
+   * To get the scalar like "Loss", "Top1Accuracy" from validation summary
+   * Return is a Array of 3-tuples
+   *
+   * @param tag The string variable represents the parameter you want to return
+   *            supported tags are "Loss", "Top1Accuracy", etc based on
+   *            validation metric. Use a method to_bigdl_metric to convert
```
hkvision: this description seems to be not clear enough for users to understand?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/python/PythonZooKeras.scala
```diff
@@ -177,6 +177,26 @@ class PythonZooKeras[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
 
+  def zooGetScalarFromSummary(
+      module: KerasNet[T],
+      tag: String,
+      target: String): JList[JList[Any]] = {
+
+    var scalarArray: Array[(Long, Float, Double)] = null
+    require(target == "Train" || target == "Validation",
+      "Invalid target, must be Train or Validation.")
+    if (target == "Train") {
```
hkvision: change to the following?
```
val scalarArray = if ... else ...
```

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/models/TrainingSpec.scala
```diff
@@ -224,6 +224,28 @@ class TrainingSpec extends FlatSpec with Matchers with BeforeAndAfter {
     res
   }
 
+  "tensorboard api" should "work" in {
+    val trainingData = generateData(Array(12, 12), 1, 100)
```
hkvision: need to add some unit test in Python as well.

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/keras/models/TrainingSpec.scala
```diff
@@ -224,6 +224,28 @@ class TrainingSpec extends FlatSpec with Matchers with BeforeAndAfter {
...
+
+    model.add(Dense[Float](8, activation = "relu", inputShape = Shape(12, 12)))
+    model.add(Flatten[Float]())
+    model.add(Dense[Float](2, activation = "softmax"))
+
+    model.compile(optimizer = "sgd", loss = "sparse_categorical_crossentropy",
+      metrics = List("accuracy"))
+    val api = new PythonZooKeras[Float]()
+
+    model.setTensorBoard("./", "testTensorBoard")
```
hkvision: better to remove the temporary test files?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
zhichao-li: If you put the checking in scala, python can also take advantage of it.

pyzoo/zoo/pipeline/api/keras/engine/topology.py
Litchilitchy: typo..

pyzoo/zoo/pipeline/api/keras/engine/topology.py
Litchilitchy: so we output supported str (line 107)

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: I did not find check in scala. And currently most api do not check but few of them do. So I guess there is no general rule of whether python or scala to check?

jason-dai: error messages from scala are not user friendly to python users?

hkvision: Zhichao means if we put the check in Scala then we do not need to check in Python as they will eventually go to Scala.

hkvision: If we put `require tag==..` in scala with error msg, then it should be clear to Python users as well.

Litchilitchy: according to this then we should support both scala and python check, thus all check should be done in scala because scala should reflect to python but it could not reverse. So why there exists some check in python?..

pyzoo/zoo/pipeline/api/keras/engine/topology.py
Litchilitchy: or directly use lower case in python would be more user friendly?

hkvision: something like:
```
tag = tag.lower()
if tag != 'loss' ...
```??

File: pyzoo/test/zoo/pipeline/api/keras/test_simple_integration.py [hunk 10]
```diff
@@ -77,7 +77,11 @@ def test_training_with_tensorboard_checkpoint_gradientclipping(self):
         model.fit(X_train, y_train, batch_size=112, nb_epoch=2, validation_data=(X_test, y_test))
         model.set_gradient_clipping_by_l2_norm(0.2)
         model.fit(X_train, y_train, batch_size=112, nb_epoch=2, validation_data=(X_test, y_test))
-        model.evaluate(X_test, y_test, batch_size=112)
+        train_loss = model.get_train_summary("Loss")
+        val_loss = model.get_validation_summary("Loss")
+        np.array(train_loss)
+        np.array(val_loss)
+        eval = model.evaluate(X_test, y_test, batch_size=112)
```
Litchilitchy: python unit test directly added here

pyzoo/zoo/pipeline/api/keras/engine/topology.py
Litchilitchy: so case not sensitive? support sth like "LoSs"... may we directly force all params to be lower case? since other params in python seems all to be lower case.

Litchilitchy: and if so, the validation method becomes sth like `sparse_categorical_entropy` using lower case and  `_`

PR #1232, opened 2019-03-20
MAE fixed bug of bigdl (zoo called bigdl in old version)

Litchilitchy:
made a new `MAE` for zoo, old code called `bigdl.optim.optimizer.MAE`, which is not correct

I had test the new code in this pull request in out notebook tutorial, it gave the correct result. The BigDL code got a wrong logic, computing 2-d array of data with 1-d array of label.

There exists nearly no usage of MAE in both BigDL and Zoo previously.

ps: issue checked with @hkvision @qiuxin2012


pyzoo/zoo/pipeline/api/keras/utils.py
```diff
@@ -88,8 +88,7 @@ def to_bigdl_metric(metric, loss):
     elif metric == "top5accuracy" or metric == "top5acc":
         return metrics.Top5Accuracy()
     elif metric == "mae":
-        from bigdl.optim.optimizer import MAE
-        return MAE()
+        return metrics.MAE()
```
Litchilitchy: @songhappy issue [here](https://github.com/intel-analytics/analytics-zoo/pull/1075/files) modified here

zhichao-li: `bigdl.optim.optimizer.MAE` is suppose to be a good choice, but seems like it has different behavior with Keras as it require the target to be a Scalar type:
 https://github.com/intel-analytics/BigDL/blob/d8ad65511d0030744e412b4cb070607a3a0bbe37/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala#L505
It would be good if you can add some comment here to clarify the difference between this version and bigdl version.

Litchilitchy: @zhichao-li using a MAE to calculate a softmax? and get index of class label as return? then the error is the err = class_label1 - class_label2... what is this for? continuous class label? referring to the unit test [here](https://github.com/intel-analytics/BigDL/blob/04dbaea17aa8f756a8c987a403f808a57367f6fd/spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/ValidationSpec.scala#L224)

Litchilitchy: @zhichao-li by the way this seems not a conflict with previous modification of @songhappy . Previous one is add MAE to metric otherwise may be a mismatch of criterion and metric? And this one is fixing its calculation step and change an interface

hkvision: I think what @songhappy changes is due to the shaded import for MAE, she just add the import locally and otherwise the MAE will be a loss not a metrics.

PR #1234, opened 2019-03-20
support TFEstimator for model without parameters

Bob-Chou:
Adjusted the place to create `tf.train.Saver()`. There would be no parameters when using `tf.import_graph_def` to restore the model to evaluate or predict. In such a case, a `ValueError` would occur because there are no parameters to save or restore when creating `tf.train.Saver()`


yangw1234:
LGTM

hkvision:
jenkins passed: http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2079/

PR #1241, opened 2019-03-22
adjust tf version and local[*]

glorysdj:
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/1241)

### DIFF
--- a/docker/install-python.sh
+++ b/docker/install-python.sh
@@ -12,7 +12,7 @@ apt-get install -y build-essential ${PYTHON} ${PYTHON}-setuptools ${PYTHON}-dev 
 ${PIP} install --upgrade setuptools
 ${PIP} install numpy scipy
 ${PIP} install --no-binary pandas -I pandas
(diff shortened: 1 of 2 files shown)


docker/install-python.sh
```diff
@@ -12,7 +12,7 @@ apt-get install -y build-essential ${PYTHON} ${PYTHON}-setuptools ${PYTHON}-dev
 ${PIP} install --upgrade setuptools
 ${PIP} install numpy scipy
 ${PIP} install --no-binary pandas -I pandas
-${PIP} install scikit-learn matplotlib seaborn jupyter wordcloud moviepy requests h5py opencv-python tensorflow
+${PIP} install scikit-learn matplotlib seaborn jupyter wordcloud moviepy requests h5py opencv-python tensorflow=1.10.0
```
yangw1234: tensorflow==1.10.0

glorysdj: ok

PR #1244, opened 2019-03-25
Val split

Litchilitchy:


Litchilitchy:
add `validation_split` params support in `fit` method of keras api

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -164,7 +164,8 @@ def set_evaluate_status(self):
                       self.value)
         return self
 
-    def fit(self, x, y=None, batch_size=32, nb_epoch=10, validation_data=None, distributed=True):
+    def fit(self, x, y=None, batch_size=32, nb_epoch=10,
+            validation_split=0, validation_data=None, distributed=True):
```
hkvision: Is this supported in Scala?

pyzoo/zoo/pipeline/api/keras/metrics.py
```diff
@@ -37,6 +37,18 @@ def __init__(self, threshold_num=200, bigdl_type="float"):
         JavaValue.__init__(self, None, bigdl_type, threshold_num)
 
 
+class MAE(ZooKerasCreator, JavaValue):
```
hkvision: Why is MAE in this PR?

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -179,11 +180,19 @@ def fit(self, x, y=None, batch_size=32, nb_epoch=10, validation_data=None, distr
         distributed: Boolean. Whether to train the model in distributed mode or local mode.
                      Default is True. In local mode, x and y must both be Numpy arrays.
         """
+
         if distributed:
             if isinstance(x, np.ndarray) and isinstance(y, np.ndarray):
-                training_data = to_sample_rdd(x, y)
                 if validation_data:
                     validation_data = to_sample_rdd(*validation_data)
+                elif validation_split != 0:
```
jason-dai: what if user specify both  validation_data and validation_split?

Litchilitchy: as Keras behavior, if `validation_data` is not None, `validation_split` would be ignored

Litchilitchy: a little difference from Keras, if keras `validation_split` is taken, and value is not valid (>1 or <0) it would just skip validation, but we raise an error here

pyzoo/zoo/pipeline/api/keras/engine/topology.py
jason-dai: This can be only supported for LocalDataSet

Litchilitchy: This api is just added for python support (specifically in our notebook writing), it makes no difference to original internal scala behavior, not affected by Local of Distributed?

PR #1245, opened 2019-03-25
using latest tensorflow's freeze funtion, fix keras embedding freeze bug

### LINKED ISSUE #1228: tfpark KerasModel has error when specifying batch_input_shape with RNN
```
model = tf.keras.Sequential([
      tf.keras.layers.Embedding(vocab_size, embedding_dim,
                                batch_input_shape=[BATCH_SIZE, None]),
      tf.keras.layers.GRU(rnn_units,
                          return_sequences=True,
                          recurrent_initializer='glorot_uniform',
                          stateful=True),
      tf.keras.layers.Dense(vocab_size)
    ])
```
The above code will throw the following Scala error:
```
py4j.protocol.Py4JJavaError: An error occurred while calling o50.createTFTrainingHelper.
: java.lang.IllegalArgumentException: Node 'gradients/lstm/while/Exit_2_grad/b_exit' expects to be colocated with unknown node 'lstm/while/Exit_2'
```

Have spotted the problem is that we specify batch_size in input_shape, since a stateful RNN needs batch_size.
It can work if we change batch_input_shape to input_shape and use a stateless RNN.

yangw1234:
should fix #1228 
do not need to use variable_creator_scope anymore
modified lasted tensorflow version of graph_util (https://github.com/tensorflow/tensorflow/commit/cebce4a2b5e33a06a1c92772008082895568f10a), and fixed bug in it:
1. treat colocated operation as input operation to prevent it being lost
2. for every operation that reads from a variable resource, use dfs to find the variable it reads and record all operations in the path. Then make those operations accept a regular tensor instead of a resource tensor (because variable will become a const after freeze).

### DIFF
--- a/pyzoo/dev/check-license
+++ b/pyzoo/dev/check-license
@@ -23,7 +23,7 @@ header="Copyright 2018 Analytics Zoo Authors"
 
 echo "Checking Analytics Zoo license for Python files..."
 
-find $DL_PYTHON_HOME ! -name "pep8-1.7.0.py" -name "*.py" -print0 | while read -d $'\0' file
+find $DL_PYTHON_HOME ! -name "pep8-1.7.0.py" ! -name "tf_graph_util.py" -name "*.py" -print0 | while read -d $'\0' file
 do
     if ! (grep -qF "$header" "$file");
     then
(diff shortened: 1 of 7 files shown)


yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2099/

PR #1246, opened 2019-03-26
unpersist rdd after fit method

Litchilitchy:
due to bigdl issue, this `unpersisit` has to be called twice. One for rdd, one for transformers & indices.

issue has been referenced in bigdl, the DistributedDataSet unpersist method in bigdl lack the unpersisting of rdd


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -391,8 +391,16 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
       featurePaddingParam: PaddingParam[T] = null,
       labelPaddingParam: PaddingParam[T] = null)(implicit ev: TensorNumeric[T]): Unit = {
     KerasUtils.validateBatchSize(batchSize)
-    this.fit(toDataSet(x, batchSize, featurePaddingParam, labelPaddingParam),
-      nbEpoch, toDataSet(validationData, batchSize, featurePaddingParam, labelPaddingParam))
+    val trainData = toDataSet(x, batchSize, featurePaddingParam, labelPaddingParam)
+    val valData = toDataSet(validationData, batchSize, featurePaddingParam, labelPaddingParam)
+    this.fit(trainData, nbEpoch, valData)
+    // due to bigdl issue, need to call 2 times
```
zhichao-li: How about detail the issue here?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -391,8 +391,16 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
+    trainData.toDistributed().unpersist()
+    trainData.toDistributed().originRDD().unpersist()
+    if (valData != null) {
+      valData.toDistributed().unpersist()
```
zhichao-li: Extract these two lines to be a private method as there are multiple places require this function?

zhichao-li:
It looks good except that we should also update the other `fit` methods as well.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: issue has been referenced, will bigdl fix this?

Litchilitchy: if they do then we should modify this once again

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -592,7 +600,16 @@ class Model[T: ClassTag] private (private val _inputs : Seq[ModuleNode[T]],
...
-    new Model[T](_inputs, nodes(Seq(output)).map(_.removeNextEdges()))
+    new Model[T](_inputs, nodes(Seq(output)).map(_.removeNextEdges())) {
+      /**
+        * Train a model for a fixed number of epochs on a DataSet.
+        *
+        * @param x              Training dataset. If x is an instance of LocalDataSet, train in local mode.
+        * @param nbEpoch        Number of epochs to train.
+        * @param validationData Dataset for validation, or null if validation is not configured.
+        */
+      override def fit(x: DataSet[MiniBatch[T]], nbEpoch: Int, validationData: DataSet[MiniBatch[T]])(implicit ev: TensorNumeric[T]): Unit = super.fit(x, nbEpoch, validationData)
```
Litchilitchy: why this pr has this thing?...

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -409,7 +427,16 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
       nbEpoch: Int,
       validationData: ImageSet)(implicit ev: TensorNumeric[T]): Unit = {
     KerasUtils.validateBatchSize(batchSize)
+    val trainData = toDataSet(x, batchSize)
```
zhichao-li: What's the point for adding `trainData` here while no one actually using it?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -409,7 +427,16 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
+    val valData = toDataSet(validationData, batchSize)
     this.fit(toDataSet(x, batchSize), nbEpoch, toDataSet(validationData, batchSize))
+
+    if (x.isDistributed()) {
```
zhichao-li: We don't need the checking here or can it be put into `releaseDataSet`?

zhichao-li:
There's still other `fit` method which should be relied on `releaseDataSet` as well. 
```
  def fit(
      x: TextSet,
      batchSize: Int,
      nbEpoch: Int,
      validationData: TextSet)(implicit ev: TensorNumeric[T]): Unit = {
    KerasUtils.validateBatchSize(batchSize)
    val dataset = x.toDataSet
    this.fit((dataset -> SampleToMiniBatch[Float](batchSize)).asInstanceOf[DataSet[MiniBatch[T]]],
      nbEpoch, toDataSet(validationData, batchSize))
    if (x.isDistributed) {
      dataset.toDistributed().unpersist()
    }
  }

```

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: @zhichao-li The TextSet has different operation already done by @qiuxin2012 , it has `SampleToMiniBatch` and would not go into that `CachedDistriDataSet` which has wrong release method.

I don't know how it would be if we use `LocalImageSet`, but I suppose if the type is RDD, then it would be distributed, otherwise not sure, and just as the `TextSet` code has this check so that I also add one.

Litchilitchy: Above also refers to what you are concerning on `TextSet` should rely on `releaseDataSet`

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
Litchilitchy: this is a mistake, will add another pr to fix it

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
zhichao-li: This is more about the style that the method `releaseDataSet(dataset)` should handle the checking for local or distributed.

Litchilitchy: now we want to use `releaseDataSet` method to handle all of `RDD, ImageSet, TextSet`, `RDD` itself is distributed and it does not have to check thus it does not has `isDistributed`. All we need to check is just `ImageSet, TextSet`.

Then we may need to write something like `if (x isInstance of RDD || (some other check) && x.isDistributed() )`, would it seem somehow worse?

zhichao-li: No.... you only need to take care of`DataSet`.

PR #1249, opened 2019-03-26
support tfdataset without batch

yangw1234:
support tfdataset without batch:
if user does not specify batch_size in TFDataset, then TFDataset.tensors will return the a tensor without batch dim.

### DIFF
--- a/pyzoo/zoo/tfpark/estimator.py
+++ b/pyzoo/zoo/tfpark/estimator.py
@@ -180,12 +180,9 @@ class TFEstimator(object):
                     if result.batch_per_thread < 0:
                         batch_size = result.batch_size
                     else:
-                        batch_size = result.batch_per_thread * result.total_core_num
+                        batch_size = result.batch_per_thread * result.rdd.getNumPartitions()
 
                     eval_methods = [self._to_bigdl_metric(m) for m in eval_methods]
-                    print("batch_size is %s " % batch_size)
-                    print("result.batch_size is %s " % result.batch_size)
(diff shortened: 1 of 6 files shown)


pyzoo/zoo/tfpark/estimator.py
```diff
@@ -205,7 +205,7 @@ def predict(self, input_fn, checkpoint_path=None):
                         saver.restore(sess, checkpoint_path)
                     else:
                         sess.run(tf.global_variables_initializer())
-                    inputs = nest.flatten(result.feature_tensors)
+                    inputs = nest.flatten(result._original_tensors[0])
```
jason-dai: I think we should require the dataset has batch when used for training

pyzoo/zoo/tfpark/estimator.py
```diff
@@ -168,7 +168,7 @@ def evaluate(self, input_fn, eval_methods, steps=None, checkpoint_path=None):
...
```
jason-dai: Do we allow `evaluate` without batch? Do we have unit test for it?

jason-dai:
I think we should require the dataset has batch when used for training (and maybe also for KerasModel)?

pyzoo/zoo/tfpark/estimator.py
yangw1234: Being without batch is disallowed in KerasModel and estimator training. Estimator evaluate and predict is allowed. Unit tests are added.

yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2109/

PR #1252, opened 2019-03-27
Refining unpersist style and test distributed data source before converting into DataSet

Litchilitchy:


Litchilitchy:
@zhichao-li would this one be better? BTW I am also testing if we could skip the `isDistributed` check

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -391,9 +391,13 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
...
    * @param dataSet Target DataSet to release
    */
-  def releaseDataSet(dataSet: DataSet[MiniBatch[T]]): Unit = {
-    dataSet.toDistributed().unpersist()
-    dataSet.toDistributed().originRDD().unpersist()
+  def releaseDataSets(dataSets: Array[DataSet[MiniBatch[T]]]): Unit = {
+    for (ds <- dataSets) {
+      if (ds != null) {
+        ds.toDistributed().unpersist()
+        ds.toDistributed().originRDD().unpersist()
```
zhichao-li: filter and foreach can be apply here, but it's ok for not depending on those.

zhichao-li:
Choose a better title for this PR?

zhichao-li:
is there any `fit` method that need to be updated as well.

Litchilitchy:
Not sure, would check after test is finished

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -391,9 +391,13 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
    *
    * @param dataSet Target DataSet to release
    */
-  def releaseDataSet(dataSet: DataSet[MiniBatch[T]]): Unit = {
-    dataSet.toDistributed().unpersist()
-    dataSet.toDistributed().originRDD().unpersist()
+  def releaseDataSets(dataSets: Array[DataSet[MiniBatch[T]]]): Unit = {
+    for (ds <- dataSets) {
+      if (ds != null) {
+        ds.toDistributed().unpersist()
```
zhichao-li: Check if `isDistributed` here?  Actually, both `Local` or `Distributed` DataSet should implement the `unpersist` method then we don't need to write such a dummy code here.

Litchilitchy: use `isInstanceOf[DistributedDataSet]`? 

I checked the code and there seems to be 5 categories of handles. `RDD, Local & DistributedImageSet, Local & DistributedTextSet`, the distributed one all called `Dataset.rdd`, seems this check would work.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -391,9 +391,13 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
    *
    * @param dataSet Target DataSet to release
    */
-  def releaseDataSet(dataSet: DataSet[MiniBatch[T]]): Unit = {
-    dataSet.toDistributed().unpersist()
-    dataSet.toDistributed().originRDD().unpersist()
+  def releaseDataSets(dataSets: Array[DataSet[MiniBatch[T]]]): Unit = {
+    for (ds <- dataSets) {
+      if (ds != null && ds.isInstanceOf[DistributedDataSet[T]]) {
```
Litchilitchy: I think this should work..

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -416,10 +420,7 @@ abstract class KerasNet[T](implicit val tag: ClassTag[T], implicit val ev: Tenso
     val valData = toDataSet(validationData, batchSize, featurePaddingParam, labelPaddingParam)
     this.fit(trainData, nbEpoch, valData)
 
-    releaseDataSet(trainData)
-    if (valData != null) {
-      releaseDataSet(valData)
-    }
+    releaseDataSets(Array(trainData, valData))
```
jason-dai: why do we release the dataset unconditionally here?

Litchilitchy: this check has been moved into `releaseDataSet` method

PR #1255, opened 2019-03-29
add auto-scaling to InferenceModel

glorysdj:
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/1255)


glorysdj:
jenkins passed.
http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2120/

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -25,11 +25,11 @@
 public abstract class AbstractInferenceModel extends InferenceModel implements Serializable {
 
   public AbstractInferenceModel() {
-    super(1, null, null);
+    super(true, 1, null, null);
   }
 
   public AbstractInferenceModel(int supportedConcurrentNum) {
-    super(supportedConcurrentNum, null, null);
+    super(false, supportedConcurrentNum, null, null);
```
qiyuangong: The default value of `autoScalingEnabled ` is not constant in initializations. It is `true` by default, and is `false` when `supportedConcurrentNum` is given.  I think we should add comments for it, or make them the same.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -240,40 +251,57 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
...
+  }
+
+  private def predict(inputActivity: Activity): Activity = {
+    val model: AbstractModel = retrieveModel()
     try {
-      val result = model.predict(inputActivity)
-      result
+      model.predict(inputActivity)
     } finally {
       modelQueue.offer(model)
```
qiyuangong: I think `retrieveModel` method fulfills scale-up. But, the max number of model copies is unlimited , which may lead to OOM. I suggest add `maxNum` parameter when auto-scaling is enabled.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
glorysdj: ok

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: ok

glorysdj:
http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2123/
passed

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -240,40 +274,57 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
...
+          model = modelQueue.take
+        } catch {
+          case e: InterruptedException =>
+            throw new InferenceRuntimeException("no model available", e);
+        }
+      case true =>
+        // if auto-scaling is enabled, will poll a model, no waiting but scale 1 model if necessary.
+        model = modelQueue.poll()
+        model match {
+          case null => model = this.originalModel.copy(1)(0)
```
jason-dai: what if it exceeds maxConcurrentNum?

jason-dai: The `copy` method is very inefficient, especially if `num ==1`

glorysdj: for now, it will create a shared weights model and do the prediction, but can not be offered to the queue. 
modelQueue.offer(model) will return false and do nothing. I will add model.release() when modelQueue.offer(model) return false.
Another solution is to use an AtomicInteger to strictly control the number of running models. if exceeds, user will wait until get a model from the queue rather than create a new one.

glorysdj: fixed copy method in FloatModel.scala
will clone model from a empty one and with shared weights.

glorysdj:
jenkins passed: http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2127

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
zhichao-li: wondering do we need to put the copied model back to the queue?

zhichao-li: oh...you would put it back while the prediction completed, but it indeed would be possible that the total number would exceed the maxConcurrentNum.

glorysdj: with a queue, user can reuse the models instead of copying them again and again

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -26,12 +26,46 @@ import com.intel.analytics.zoo.pipeline.inference.DeviceType.DeviceTypeEnumVal
...
+   * @param maxConcurrentNum       the max concurrentNum for the LinkedBlockingQueue
+   * @return a specified InferenceModel
+   */
+  def this(autoScalingEnabled: Boolean, supportedConcurrentNum: Int, maxConcurrentNum: Int) =
+    this(autoScalingEnabled, supportedConcurrentNum, null, null, maxConcurrentNum)
+
+  this.modelQueue = autoScalingEnabled match {
+    // Creates a LinkedBlockingQueue with a capacity of maxConcurrentNum
+    case true => new LinkedBlockingQueue[AbstractModel](maxConcurrentNum)
+    case false => new LinkedBlockingQueue[AbstractModel](supportedConcurrentNum)
```
jason-dai: can we just combine `maxConcurrentNum` and `supportedConcurrentNum` into one argument?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -240,40 +274,61 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
...
+
+  private def predict(inputActivity: Activity): Activity = {
+    val model: AbstractModel = retrieveModel()
     try {
-      val result = model.predict(inputActivity)
-      result
+      model.predict(inputActivity)
     } finally {
-      modelQueue.offer(model)
+      val ifSuccessed = modelQueue.offer(model)
```
jason-dai: `success = modelQueue.offer(model)`

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -291,6 +346,6 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
   def getOriginalModel: AbstractModel = originalModel
```
jason-dai: `offerModelQueue` should check if autoScalingEnabled is true?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: ok

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: for now, for auto-scaling disabled InferenceModel, supportedConcurrentNum is used to limit the number of model copies, for auto-scaling enabled InferenceModel, supportedConcurrentNum is used to initialize model copies when get started, and maxConcurrentNum is only used for  auto-scaling enabled InferenceModel to limi the max number of model copies.
it's OK to use only one concurrentNum to represent both, when auto-scaling disabled, it limits the copies, when  auto-scaling enabled, it limits the max number of copies, but always starts with 1 copy.
so for the api, 
InferenceModel() will create an auto-scaling enabled one, 
InferenceModel(concurrentNum) will create an auto-scaling disabled one, 
InferenceModel(autoScalingEnabled, concurrentNum) will provide the flexibility.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: If we use supportedConcurrentNum to initialize the InferenceModel with fixed size of model copies, it does not matter if autoScaling is enabled or not. If we combine two parameters to one concurrentNum and if the initialization of autoScaling enabled and disabled are different, we should check it.
if ture, offer only one.
if false, offer concurrentNum ones.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -240,40 +274,61 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
...
     try {
-      val result = model.predict(inputActivity)
-      result
+      model.predict(inputActivity)
     } finally {
-      modelQueue.offer(model)
+      val ifSuccessed = modelQueue.offer(model)
+      ifSuccessed match {
+        case true =>
+        case false => model.release()
```
jason-dai: Need to make sure the model weight is emptied before releasing

glorysdj: ok. now add clearWeightBias and releaseWeightBias,
to get a empty model, will clone and releaseWeightBias
to relase a copied model which share weights, will clearWeightBias  before model.release

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -189,6 +192,63 @@ trait InferenceSupportive {
...
+      if (tensors(i) != null) {
+        tensors(i).set()
+      }
+      i += 1
+    }
+  }
+
+  def makeUpModel(metaModel: Module[Float], weightBias: Array[Tensor[Float]]):
+  AbstractModule[Activity, Activity, Float] = {
+    val cloned = metaModel.cloneModule()
```
jason-dai: why clone again?

glorysdj: from one original FloatModel, each copied one will take reference to the same metaModel.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -240,40 +269,61 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
...
+
+  private def predict(inputActivity: Activity): Activity = {
+    val model: AbstractModel = retrieveModel()
     try {
-      val result = model.predict(inputActivity)
-      result
+      model.predict(inputActivity)
     } finally {
-      modelQueue.offer(model)
+      val success = modelQueue.offer(model)
```
jason-dai: what will be the value for `model` if there is an exception?

glorysdj: will model.predict(inputActivity) change model value? I think if there is no side effect, model will be the same when it is retrieved.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: I think here need to add more logic. the original model's weights and native resource also need to be released.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -189,6 +192,63 @@ trait InferenceSupportive {
...
+
+  def clearWeightBias(model: Module[Float]): Unit = {
+    clearTensors(model.parameters()._1)
+    clearTensors(model.parameters()._2)
+  }
+
+  private def clearTensors[T: ClassTag](tensors: Array[Tensor[T]])
+                                       (implicit ev: TensorNumeric[T]) = {
+    var i = 0
+    while (i < tensors.length) {
```
jason-dai: will this work? I don't think this will change the value of weight tensors in the model.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
jason-dai: what if the exception is in model retrieval?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
jason-dai: Think the user will need to explicitly release the original model (maybe provide a release method for InferenceModel)

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
jason-dai: I think you have already cloned the model before calling makeUpModel

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
glorysdj: clearTensors is to set reference to null, not change the value of weights since the original and other copied ones are still using the shared weights.
releaseTensors is to set tensors(i) to size 0, this will change the value of weights, so this is only used by release a original model after all copied ones are released.
I have tested them in the unit tests.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: ok, i think we need check if model is null before offer it back.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
glorysdj: I have changed the logic, the metaModel is passed to the constructor. the previous implementation is not efficient as you pointed out.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: InferenceModel already has one clearModelQueue, which is used to release all the models and clear the queue.
any load and reload function will call it first.
I will also add a release method to InferenceModel to be called explicitly if user still need one.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatModel.scala
```diff
@@ -28,7 +28,9 @@ import com.intel.analytics.bigdl.utils.Table
 import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 
-class FloatModel(var model: AbstractModule[Activity, Activity, Float])
+class FloatModel(var model: AbstractModule[Activity, Activity, Float],
+                 var metaModel: AbstractModule[Activity, Activity, Float],
+                 var isOriginal: Boolean)
```
zhichao-li: What's `isOriginal` ?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -157,8 +188,8 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
    * @param weightPath the path of openvino ir bin file
    */
   def doLoadOpenVINO(modelPath: String, weightPath: String): Unit = {
-    if (supportedConcurrentNum > 1) {
-      InferenceSupportive.logger.warn(s"supportedConcurrentNum is $supportedConcurrentNum > 1, " +
+    if (concurrentNum > 1) {
+      InferenceSupportive.logger.warn(s"concurrentNum is $concurrentNum > 1, " +
```
zhichao-li: should we throw an exception here? or rephrase the warning message that concurrent prediction would consume lots of memory as  openvino does not support shared weights model copies

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -240,40 +271,80 @@ class InferenceModel(private var supportedConcurrentNum: Int = 1,
...
+
+  /**
+   * release original model and all the cloned ones in the queue
+   */
+  def release() = {
+    clearModelQueue()
+  }
+
+  private def predict(inputActivity: Activity): Activity = {
+    val model: AbstractModel = retrieveModel()
```
zhichao-li: Shall we suppose that `retrieveModel` would always return a valid model as it would block there if not enough resource? then we don't need to check the returning result.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
zhichao-li:
```
array = [model.tensor1, model.tensor2]
array(0)=null
array(1)=null
```
would only alter the array itself other than modify the reference of `model.tensor1`?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatModel.scala
glorysdj: to indicate if the model is cloned from others, if is cloned will not release its weights, otherwise will release its weights.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: it will not clone models when it is an openvinomodel, as the warning message said: not supported.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/FloatModel.scala
zhichao-li: releaseWeights? hhhh

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
zhichao-li: ok. should we fail fast here as the input parameter is invalid?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: if auto-scaling enabled, it may return null, or throw exception in the retrieveModel  method, so we do need check if it is null.

zhichao-li: Throw an exception is good, but in which case that "auto-scale" would return `null` as it would always copy a new one if not enough in the queue?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
glorysdj: this is wrong, i will refine the logic.

glorysdj: finally fixed. please check.
or give me advice if you have better ways.
Thanks.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
```diff
@@ -200,20 +200,26 @@ trait InferenceSupportive {
...
-    clearTensors(model.parameters()._2)
-  }
-
-  private def clearTensors[T: ClassTag](tensors: Array[Tensor[T]])
-                                       (implicit ev: TensorNumeric[T]) = {
+    model.reset()
+    val weightBias = model.parameters()._1
+    val clonedweightBias = model.parameters()._1.map(tensor => {
+      val newTensor = Tensor[Float]().resizeAs(tensor)
+      newTensor.copy(tensor)
```
jason-dai: This is expensive, and does not work for quantized tensor.

Let's ignore this problem for now (just directly release the model), which does not support quantized model in this version.

glorysdj: ok

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: yes, you are right, logically retrieveModel always return a valid model or throw an exception.  predict to check if it is null is just in case since the model is initialized as null in the retrieveModel.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
glorysdj: I think warning is enough here, it can work to support concurrent requests.

glorysdj:
jenkins passed
http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2133/
there is style error caused by commit 1164, I have asked Jennie to correct it.

if this pr is ok, i will merge it and add openvino image classification based on this pr.

--- scalastyle-maven-plugin:0.8.0:check (default) @ analytics-zoo-bigdl_0.8.0-spark_1.6.3 ---
error file=/opt/work/jenkins/workspace/ZOO-PR-Scala-StyleCheck/zoo/src/main/scala/com/intel/analytics/zoo/examples/objectdetection/finetune/ssd/Train.scala message=File line length exceeds 100 characters line=171
error file=/opt/work/jenkins/workspace/ZOO-PR-Scala-StyleCheck/zoo/src/test/scala/com/intel/analytics/zoo/models/image/objectdetection/ssd/SSDSpec.scala message=File line length exceeds 100 characters line=97
Saving to outputFile=/opt/work/jenkins/workspace/ZOO-PR-Scala-StyleCheck/zoo/target/stylecheck/scalastyle-output.xml

glorysdj:
in order to add openvino image classification, will now merge this pr first.
please open issues to me if still need any change

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSupportive.scala
jason-dai: Line 84 ~ 85:
```
   val clonedModel = metaModel.cloneModule()
   val clonedModelWithWeightsBias = makeUpModel(clonedModel, weightBias)
```

glorysdj: fixed in InferenceSupportive.scala
 def makeUpModel(clonedModel: Module[Float], weightBias: Array[Tensor[Float]]):
  AbstractModule[Activity, Activity, Float] = {
    putWeightBias(clonedModel, weightBias)
    clonedModel.evaluate()
    clonedModel
  }

glorysdj:
http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2143/

PR #1256, opened 2019-03-29
support init dataset from ndarrays

yangw1234:


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2115/ jenkins pass

pyzoo/zoo/tfpark/estimator.py
```diff
@@ -62,15 +62,13 @@ def add_train_op(model_fn, features, labels, mode, params, config, optimizer):
                 " that do not support gradients, between variables %s and loss %s." %
                 ([str(v) for _, v in grads_and_vars], spec.loss))
 
-        train_op = spec.optimizer.apply_gradients(grads_and_vars,
-                                                  global_step=tf.train.get_global_step())
+        train_op = optimizer.apply_gradients(grads_and_vars,
+                                             global_step=tf.train.get_global_step())
```
jason-dai: should be `train_op = optimizer.minimize(spec.loss, global_step=tf.train.get_global_step())`, no?

pyzoo/test/zoo/tfpark/test_tfpark_estimator.py
```diff
@@ -56,6 +56,11 @@ def input_fn(mode):
                                              features=(tf.float32, [10]),
                                              labels=(tf.int32, []),
                                              batch_size=4)
+            elif mode == tf.estimator.ModeKeys.EVAL:
```
jason-dai: this condition is redundant - the if statement above already checks whether it is EVAL mode

yangw1234: yeah, I should have deleted the "or mode == tf.estimator.ModeKeys.Eval" above

pyzoo/zoo/tfpark/estimator.py
yangw1234: `minimize(loss)` is equivalent to first `compute_gradient(loss)` then `apply_gradients`. It got to this way because I used to want to record the gradient.  I'll change to the `minimize` version for simpilicity.

PR #1258, opened 2019-04-01
add inference examples for flink and spark, batch and streaming

glorysdj:
Still need to change the project structure, clean code, add readme and add more example including openvino, and auto-scaling,

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/1258)


apps/model-inference-examples/model-inference-spark/src/main/scala/com/intel/analytics/zoo/apps/model/inference/spark/SparkBatchExample.scala
```diff
@@ -0,0 +1,78 @@
+package com.intel.analytics.zoo.apps.model.inference.spark
```
jason-dai: why do we need this?

glorysdj: I will delete spark examples

PR #1260, opened 2019-04-01
Seq2seq doc fix for numLayers parameter

abdolence:
A small doc update for `numLayers` parameter in Seq2Seq (`numLayer` doesn't exist)


hkvision:
Thanks for fixing this typo!

PR #1264, opened 2019-04-04
fix tf graph util bug

yangw1234:


jason-dai:
Do we have unite tests for these bugs?

yangw1234:
@jason-dai I will add the unit tests, but currently I still do not know how to reproduce the bug. I found 
 it by inspecting the error message and code. @Bob-Chou could you provide a few lines of code independet of the ocr repo that can reproduce your problem? We will add it to the unit tests.

jason-dai:
In addition, we need to separate the TF implementation update and our fixes into different commits/PRs

yangw1234:
separated to two commits, first check in tensorflow latest update, then check in our fix

yangw1234:
jenkin pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2132/
I'll merge it first. Unit tests will be added soon.

PR #1269, opened 2019-04-10
Support creating a ImageSet from a folder of images

yangw1234:


yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2146/?auto_refresh=true

dding3:
Seems like `fromImageFolder` doesn't support read from hdfs folder, I think we need support that. 

More importantly, seems like there are lots of duplicate code of `fromImageFolder` with `ImageSet.read` and the difference is `fromImagaeFolder` supports read `label`. And it may be bring confusion. Can we add this feature to `ImageSet.read`? eg. add a sth like `readLabel` param in `ImageSet.read` to support read label?

yangw1234:
Per discussion with @dding3 offline, I will change `fromImageFolder` to `readWithLabel` and hdfs support will be added in the next PR.

jason-dai:
I think we should add this functionality to `ImageSetr.read`

yangw1234:
@jason-dai as discussed with @dding3 , `readWithLabel` should return a `LabelMap` to tell the called which class maps to which label (for example we have folder cats and folder dogs, label map tells user whether cats is 1 and dogs is 2 or dogs is 1 and cats is 2). This will change the signature of the ImageSet.read, thus we decided to add another method.

yangw1234:
@jason-dai @dding3 I have a another idea. Add a `labelMap` field in ImageSet and pass in the labelMap to ImageSet through constructor. Is that ok?

yangw1234:
1. Moved this functionality to `ImageSet.read`
2. Added a `labelMap` field in ImageSet.
3. Added HDFS support

yangw1234:
test hdfs manually

yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2149/

yangw1234:
I'll merge it first as there are little public api changes and my next pr depends on this. If further changes are required, please leave a mesage.

PR #1274, opened 2019-04-11
Update object detection predict example location

jenniew:
Update robot and doc inference for the package change


jenniew:
Jenkins pass: http://172.168.2.101:8080/job/ZOO-PR-Validation/2147/

PR #1280, opened 2019-04-15
Make ImageSet and TextSet transform return a new ImageSet/TextSet

### LINKED ISSUE #1279: ImageSet and TextSet transform method in python should return a new Set
I think `ImageSet.transform` and `TextSet.transform` in python should return a new ImageSet or TextSet instead modifying the current one as they do in scala api.

yangw1234:
fix #1279


yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2158/

dding3:
LGTM

Issue #1277, opened 2019-04-12
Keras-style API model load/save issue for resume training


shane-huang:
1. Neither of ```save``` or ```saveModel``` work with ```Net.load```
     *  ```model.save``` + ```Net.load``` will lead to NullPointerException from bigdl.utils.serializer.ModuleLoader. The error message does not make any sense to the user. 
     *  save a model using```model.saveModel``` and then load model using ```Net.load```, then call ```model.fit``` will lead to an error says "'ZooKerasLayer' object has no attribute 'fit'"

1. We should add test case and example for resume training as it's a very common request from customers.  Also, model ```save``` and ```saveModel``` has no clear documents and user don't know which they should use. It easily cause confusion. We may just keep one of them.

shane-huang:
related to issue #1004

zhichao-li:
Any example code for reproducing this?
I tried with the following snippet, and it works fine:
```
  input = ZLayer.Input(shape=(5,))
        output = ZLayer.Dense(10)(input)
        zmodel = ZModel(input, output, name="graph1")
        tmp_path = "/tmp/aaa" 
        # zmodel.saveModel(tmp_path, None, True)
        model_reloaded = Net.load(tmp_path)
```

hkvision:
Probably Net.load in Python doesn't gives an instance of KerasNet and thus can't call fit?

hkvision:
Emmm... Just checked that Net.load returns a KerasNet and should have fit method...

zhichao-li:
fixed.

PR #1282, opened 2019-04-15
Fix Net.load

zhichao-li:
1)  User should use `model.saveModel` and exception would be raised if they use `model.save` accidentally.
2) `Net.load` should return a correct python class(Model or Sequential) rather than `KerasLayer` as we cannot use `fit` or `compile` for this type.


zhichao-li:
https://github.com/intel-analytics/analytics-zoo/issues/1277

zhichao-li:
http://10.239.47.211:18888/job/ZOO-PR-Validation/2160/

zhichao-li:
cc @yangw1234 any concerns for the changes regarding to the `GraphNet` staff?

zhichao-li:
Merging this for now. Let's reopen it if come up with any further thoughts.

PR #1286, opened 2019-04-16
KerasZooModel

songhappy:
Separate KerasZooModel from recommendation PR

### DIFF
--- a/pyzoo/zoo/examples/anomalydetection/anomaly_detection.py
+++ b/pyzoo/zoo/examples/anomalydetection/anomaly_detection.py
@@ -53,8 +53,7 @@ if __name__ == "__main__":
     model = AnomalyDetector(feature_shape=(int(options.unroll_len), feature_size),
                             hidden_layers=[8, 32, 15], dropouts=[0.2, 0.2, 0.2])
     model.compile(loss='mse', optimizer='rmsprop', metrics=['mae'])
-    model.fit(train, batch_size=int(options.batch_size), nb_epoch=int(options.nb_epoch),
-              validation_data=test)
+    model.fit(train, batch_size=int(options.batch_size), nb_epoch=int(options.nb_epoch))
     test.cache()
     y_predict = model.predict(test).map(lambda x: float(x[0]))
     y_truth = test.map(lambda x: float(x.label.to_ndarray()[0]))
--- a/pyzoo/zoo/models/anomalydetection/anomaly_detector.py
+++ b/pyzoo/zoo/models/anomalydetection/anomaly_detector.py
@@ -86,8 +86,7 @@ class AnomalyDetector(KerasZooModel):
         weight_path: The path for pre-trained weights if any. Default is None.
         """
         jmodel = callBigDlFunc(bigdl_type, "loadAnomalyDetector", path, weight_path)
-        model = ZooModel._do_load(jmodel, bigdl_type)
-        model.model = KerasZooModel._do_load(jmodel, bigdl_type)
+        model = KerasZooModel._do_load(jmodel, bigdl_type)
         model.__class__ = AnomalyDetector
         return model
 
--- a/pyzoo/zoo/models/anomalydetection/anomaly_detector.py
+++ b/pyzoo/zoo/models/anomalydetection/anomaly_detector.py
@@ -18,7 +18,7 @@ import warnings
 
 from zoo.pipeline.api.keras.models import Sequential
 from zoo.pipeline.api.keras.layers import *
-from zoo.models.common import ZooModel
+from zoo.models.common import *
 from bigdl.util.common import callBigDlFunc
 from bigdl.util.common import Sample
 
@@ -27,7 +27,7 @@ if sys.version >= '3':
     unicode = str
 
 
-class AnomalyDetector(ZooModel):
+class AnomalyDetector(KerasZooModel):
     """
     The anomaly detector model for sequence data based on LSTM.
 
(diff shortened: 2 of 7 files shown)


pyzoo/zoo/models/anomalydetection/anomaly_detector.py
```diff
@@ -87,55 +87,10 @@ def load_model(path, weight_path=None, bigdl_type="float"):
         """
         jmodel = callBigDlFunc(bigdl_type, "loadAnomalyDetector", path, weight_path)
         model = ZooModel._do_load(jmodel, bigdl_type)
+        model.model = KerasZooModel._do_load(jmodel, bigdl_type)
```
hkvision: Since you are still wrapping methods from Scala, is this still needed?
We discussed this because we originally want to call model.compile.

pyzoo/zoo/models/common/zoo_model.py
```diff
@@ -26,13 +27,16 @@ class ZooModelCreator(JavaValue):
     def jvm_class_constructor(self):
         name = "createZoo" + self.__class__.__name__
         print("creating: " + name)
+
+
```
hkvision: ??

pyzoo/zoo/models/common/zoo_model.py
```diff
@@ -89,3 +93,32 @@ def _do_load(jmodel, bigdl_type="float"):
...
+class KerasZooModel(ZooModel):
+    """
+    The base class for Keras models in Analytics Zoo.
+    """
+    # leverage compile fit of kerasNet and so on
+    def compile(self, optimizer, loss, metrics=None):
+        self.model.compile(optimizer, loss, metrics=None)
+
+    def fit(self, x, y=None, batch_size=32, nb_epoch=10, validation_data=None, distributed=True):
+        self.model.fit(x, y, batch_size, nb_epoch, validation_data, distributed)
```
hkvision: oops, is it workable now?

songhappy:
PR-validation passed
http://172.168.2.101:8080/job/ZOO-PR-Validation/2166/

pyzoo/zoo/examples/anomalydetection/anomaly_detection.py
```diff
@@ -53,8 +53,7 @@ def load_and_scale(input_path):
     model = AnomalyDetector(feature_shape=(int(options.unroll_len), feature_size),
                             hidden_layers=[8, 32, 15], dropouts=[0.2, 0.2, 0.2])
     model.compile(loss='mse', optimizer='rmsprop', metrics=['mae'])
-    model.fit(train, batch_size=int(options.batch_size), nb_epoch=int(options.nb_epoch),
-              validation_data=test)
+    model.fit(train, batch_size=int(options.batch_size), nb_epoch=int(options.nb_epoch))
```
hkvision: Why changing this?

hkvision:
Want to clarify the following:
- Seems you have fixed the bug you previously mentioned and using the idea of our first discussion last week?
- For other models (text classifier, seq2seq), shall each owner change the inheritance accordingly?
- Any unit test added to cover these changes? Especially call these methods after model loading.

Probably we can have a quick chat tomorrow.

pyzoo/zoo/examples/anomalydetection/anomaly_detection.py
songhappy: minor, just want it to look cleaner, it might be confused for somebody that why using test for validation

songhappy:
This PR is to port KerasNet APIs to KerasZooModel. It works to leverage KerasNet and by wrapping a model inside KerasZooModel given both scala and python have a model from build_model. The other methodology is to wrap APIs like kerasZooComple in scala and callBigDlFunc(self.bigdl_type, "kerasZooCompile" from python. Both approaches are better than current implementation since each ZooModel has its own compile and fit and so on, it is redundant; the first approach is better than the second one since it leverages old code and involves less new implementation, plus it is consistent for both scala and python.
 
Each owner of Keras style ZooModel has a choice to use KerasZooModel other than owning a very similar piece of "compile, fit, setTensorBoard...." 

No need to have separate unit tests since all are covered by KerasNet.

pyzoo/zoo/examples/anomalydetection/anomaly_detection.py
songhappy: minor, just want it to look cleaner, it might be confused for somebody who might ask why using test for validation

pyzoo/zoo/models/anomalydetection/anomaly_detector.py
jason-dai: Why do we need to load the model twice?

pyzoo/zoo/models/common/zoo_model.py
```diff
@@ -89,3 +90,54 @@ def _do_load(jmodel, bigdl_type="float"):
...
+    def predict(self, x, batch_per_thread=4, distributed=True):
+        self.model.predict(x, batch_per_thread, distributed)
+
+    @staticmethod
+    def _do_load(jmodel, bigdl_type="float"):
+        model = Layer(jvalue=jmodel, bigdl_type=bigdl_type)
+        model.value = jmodel
+        labor_model = callBigDlFunc(bigdl_type, "getModule", jmodel)
+        model.model = KerasNet(labor_model)
+        return model
```
jason-dai: `model` is a BigDL `Layer`, and `model.model` is a `KerasNet`; however, you are assigning `model` to `AnomalyDetector.model` above - is this correct? do we have a unit test for that?

hkvision: In a specific model, for example Anomaly Detection, the model type will be AnomalyDector, and model.model will be KerasNet.

hkvision:
I don't think the unit tests about KerasNet covers your case. In other words, those unit tests only make sure a KerasNet has compile fit method, but doesn't make sure AnomalyDetector has compile fit.

pyzoo/zoo/models/anomalydetection/anomaly_detector.py
hkvision: No need ZooModel._do_load here as what it does is actually done in KerasZooModel

zoo/src/main/scala/com/intel/analytics/zoo/models/common/KerasZooModel.scala
```diff
@@ -0,0 +1,183 @@
...
+import org.apache.spark.rdd.RDD
+
+import scala.reflect.ClassTag
+
+abstract class KerasZooModel[A <: Activity : ClassTag, B <: Activity : ClassTag, T: ClassTag]
+(implicit ev: TensorNumeric[T]) extends ZooModel[A, B, T] {
+
+  // This class is defined to leverage APIs of KerasNet
+  // For the following methods, please refer to KerasNet for documentation.
+
```
hkvision: add a require model to be KerasNet[T]?

pyzoo/zoo/models/common/zoo_model.py
jason-dai: In `AnomalyDetector.load_model`, we have 

```
model.model = KerasZooModel._do_load(jmodel, bigdl_type)
```

Is this correct?

hkvision: Not really. I think the correct should be the following:
1. In KerasZooModel._do_load:
   - Call ZooModel._do_load to load the Container model `model`
   - callBigDLFunc getModule to load the inner labor jmodel and assign KerasNet(jmodel) as `model.model`
 2. In AnomalyDetecor._do_load:
    - Call KerasZooModel._do_load
    - Cast the class to be AnomalyDector.

songhappy: it makes sense to have only  model = KerasZooModel._do_load(jmodel, bigdl_type).
updated

zoo/src/main/scala/com/intel/analytics/zoo/models/common/KerasZooModel.scala
songhappy: not necessary

zhichao-li:
Would it be possible that if we remove the `ZooModel` and change the hierarchy like this:
 ```
class KerasZooModel extends KerasNet {
 def buildModel()
 def save()
  ....
}

```
The benefits are we don't need to duplicate tons of these methods: `compile, fit, evaluate ...` and  `KerasZooModel` is fairly maintainable if we want to add/refactor functions for `KerasNet`.

It's just one of my rough idea and which of the two you prefer might be partly a matter of taste. :)

songhappy:
@zhichao-li ZooModel is defined to all models in the Zoo, and KerasZooModel is for Keras style models, there are a number of models using torch style, so ZooModel is still needed

hkvision:
LGTM.

zhichao-li:
@songhappy  Got the idea. Maybe we can pick up this later when we have a free slot.
Actually, `KerasNet` can accept any "Model" including "torch model"  by overring the function `labor` although the naming is confusing here.

PR #1289, opened 2019-04-18
update gelu in bert

dding3:
`gelu` in bert is a slight different with transformer `gelu`, override the method in bert.


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/BERT.scala
```diff
@@ -68,6 +69,13 @@ class BERT[T: ClassTag] (
     new Dense(outputSize, init = RandomNormal(0.0, initializerRange))
   }
 
+  override def gelu(x: Variable[T]): Variable[T] = {
+    val y = x / math.sqrt(2.0)
+    val e = new KerasLayerWrapper[T](new InternalERF[T]()
```
jason-dai: why not make `erf` an operator in autograd?

dding3:
jekins pass. http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2179/

PR #1290, opened 2019-04-18
Support bert save/load

dding3:
This PR supports load from a pretrained bert model and use it to inference for input data with different seq length. It includes several changes:
1. Bug fix for `Parameter` save/load
2. Add bert save/load api

This PR changed original interfaces and it is an initial version to collect feedback about the change.


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/KerasParameter.scala
```diff
@@ -127,12 +128,12 @@ private[zoo] class InternalParameter[T: ClassTag](
   }
 
   override def updateOutput(input: Tensor[T]): Tensor[T] = {
-    output = weight
+    output = weight.clone()
```
jason-dai: why clone?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/KerasParameter.scala
```diff
@@ -127,12 +128,12 @@ private[zoo] class InternalParameter[T: ClassTag](
...
     output
   }
 
   override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
-    gradInput = gradOutput
+    gradInput = gradOutput.clone()
```
jason-dai: why clone?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/autograd/KerasParameter.scala
dding3: `output` will be cleared right before saving module. If `output` is point to `weight`, `weight` will also be cleared and save an empty tensor

jason-dai: Can we override the behaviors wrt module saving? otherwise clone will introduce additional overheads in forward & backward

dding3: Have tried that, however some following layers will update `output` in place and `weight` will also be changed and saved value will not be correct.

How about we add an additional parameter to enable user config if `output ` reuse `weight`?

jason-dai: Which layers will update output inplace? we need to fix those layers.

dding3: Seems there are many layers that support user inplace udpate `input`, eg, `AddConstant`, `CAddTable`, `MulConstant` etc. If we use these layers with `Parameter save`, the result will not be correct, seems not easy to forbid user do this.

zhichao-li: A simple trick would be you can clone the `weight` outside the `updateOutput` method and reuse it whenever you want.

jason-dai: I think all these layers have an `inplace` argument which default to false?

dding3: Yes, but user is able to set `inplace` to true and use it together with `Parameter`?

jason-dai: In our keras layers, it is always set to false?

dding3: Yes, removed `weight.clone`

dding3:
cc @jason-dai

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/layers/BERT.scala
```diff
@@ -169,4 +187,196 @@ object BERT {
...
+   * Load an existing model (with weights).
+   *
+   * @param path The path for the pre-defined model.
+   *             Local file system, HDFS and Amazon S3 are supported.
+   *             HDFS path should be like "hdfs://[host]:[port]/xxx".
+   *             Amazon S3 path should be like "s3a://bucket/xxx".
+   * @param weightPath The path for pre-trained weights if any. Default is null.
+   * @tparam T Numeric type of parameter(e.g. weight, bias). Only support float/double now.
+   */
+  def loadModel[T: ClassTag](path: String, weightPath: String = null, inputSeqLen: Int = -1,
```
jason-dai: add scala doc for the arguments

jason-dai: instead of using `loadModel`, you can actually add an argument for the model path to the factory method, e.g.,

```
 def apply[@specialized(Float, Double) T: ClassTag](
    vocab: Int = 40990,
    hiddenSize: Int = 768,
    ...,
    inputSeqLen: Int = -1
    modePath: String = null, 
    weightPath: String = null
    ) BERT[T] = ...
```

dding3: In that case, user has to know `vocab`, `hiddenSize` etc of the loaded model, or the parameter size will be different and will cause exception

jason-dai: Or just directly change this method to a factory method?

dding3: Updated.

dding3:
tried jekins several  times and each time different job failed due to unstable jekins env. I think we can assume jekins pass.
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2201/
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2202/
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2203/
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2204/

PR #1295, opened 2019-04-22
Add norm option to estimator.

qiuxin2012:
Add l2norm and contant norm to estimator API.


jason-dai:
Shall we follow Keras style API instead, e.g., using `setConstantGradientClipping` and `setGradientClippingByL2Norm`?

qiuxin2012:
jenkins passed

PR #1301, opened 2019-04-29
Split net.py into multiple files

yangw1234:
net.py is too large, split it into multiple files for better code structure.

No changes to the public api.


yangw1234:
jenkins pass with two random fail http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2226/

PR #1317, opened 2019-05-10
Make bert classifier ut cost less memory

hkvision:
Use smaller data
Now need 5g.
Will investigate more why so much is used.


hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2240/

PR #1318, opened 2019-05-10
Change RandomCropper to CenterCrop in ResNet50 preprocessor

hkvision:
- Run evaluation on ImageNet val data and got same results.
Top1Accuracy is Accuracy(correct: 37912, count: 50000, accuracy: 0.75824)
Top5Accuracy is Accuracy(correct: 46332, count: 50000, accuracy: 0.92664)
- Latency for one image preprocessing is around ~2.3 ms and the overall end2end latency is <10ms.

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/models/image/imageclassification/ImageClassificationConfig.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/models/image/imageclassification/ImageClassificationConfig.scala
@@ -18,7 +18,6 @@ package com.intel.analytics.zoo.models.image.imageclassification
 
 import java.net.URL
 
-import com.intel.analytics.bigdl.dataset.image.CropCenter
 import com.intel.analytics.bigdl.numeric.NumericFloat
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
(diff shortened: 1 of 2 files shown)


jason-dai:
LGTM

hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2243/

zoo/src/main/scala/com/intel/analytics/zoo/models/image/imageclassification/ImageClassificationConfig.scala
```diff
@@ -137,7 +136,7 @@ object ImagenetConfig {
   // Preprocessor for ResNet50 pre-trained in BigDL
   def bigdlResNetPreprocessor(): Preprocessing[ImageFeature, ImageFeature] = {
       ImageRandomResize(256, 256) ->
```
jason-dai: Can we just use resize?

hkvision: Tested that using resize will get almost compatible results and we can change to that. But for performance probably they are more or less the same (I will test further).

PR #1324, opened 2019-05-15
Bug fix for Resize and change imagecodec in vnni predict

hkvision:
- In ImageResize, resizeMode and useScaleFactor don't take effect.
- Change back to RandomResize... since @wzhongyuan said there's accuracy difference.
- Change Imagecodec to LOAD_IMAGE_COLOR so that both RGB and gray images can be supported by the model.


hkvision:
jenkins passed: http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2262/

PR #1333, opened 2019-05-17
Add sc to VNNI predict example

hkvision:
If use localMode, batch size is always 4, which is too small.
Opened an issue in BigDL: https://github.com/intel-analytics/BigDL/issues/2820


hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2276/

PR #1336, opened 2019-05-17
improve adamweightdecay perf

dding3:


hkvision:
Can we merge this PR?

dding3:
I run text_classifier.py with this PR once and seems there is no decrease in the accuracy. If this PR blocks your development, I think it should be fine to merge this PR :) . Otherwise I would like to do more test on this PR.

hkvision:
Not blocking my development but it will be good if it can be merged. :)
I ran classifier example and the performance boosted a bit; but not as fast as Adam.

PR #1339, opened 2019-05-21
Pytorch inference

hhbyyh:
I'll create the Pytorch branch after branch-0.5 is cut.

I compressed all files in `libtorch/lib` and the native lib we built `libpytorch-engine` to `resources/pytorch/lib.zip`, which has a size of 77.0MB. The original size of `libtorch/lib` is 292.8 MB. While this reduces the size of the package, it does add some overhead (about 1.7 second) when TorchNet is first loaded on each executor. It should not have impact on model inference speed as the native libs are only loaded once for each JVM.


zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -0,0 +1,105 @@
...
+    std::shared_ptr<torch::jit::script::Module> model_ptr = torch::jit::load(p_model_path);
+    assert(model_ptr != nullptr);
+
+    mtx.lock();
+    handles.push_back(model_ptr);
+    int id = handles.size() - 1;
+    mtx.unlock();
+
+    jenv->ReleaseStringUTFChars(jmodel_path, p_model_path);
+    return id;
```
jason-dai: `id` is an unsigned integer?

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -0,0 +1,83 @@
...
+    TorchNet wraps a TorchScript model as a single layer, thus the Pytorch model can be used for
+    distributed inference or training.
+    :param path: path to the TorchScript model.
+    """
+
+    def __init__(self, path, bigdl_type="float"):
+        super(TorchNet, self).__init__(None, bigdl_type, path)
+
+    @staticmethod
+    def from_pytorch(module, dummpy_input):
```
jason-dai: change to `from_pytorch(module, input_shape)`, so that we can automatically generate the dummy input in our lib

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/TARCompress.java
```diff
@@ -0,0 +1,35 @@
+package com.intel.analytics.zoo.pipeline.api.net;
```
jason-dai: why in `pipeline.api.net`?

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -0,0 +1,105 @@
+#include <cstring>
```
jason-dai: license header

jason-dai:
Several improvements needed (probably in a follow-up PR):
1) Instead of holding and cloning the byte buffer `modelbytes` for every `TorchNet`, we need to make sure there is only one copy per JVM, similar to `TFNet.graphDef` and `TFGraphHolder`
2) Instead of creating a temp model file, can we use 'torch::jit::load(std::istream &in, ..)' to load the model instead?
3) Also add a simple example

pyzoo/zoo/pipeline/api/net/torch_net.py
hhbyyh: sure

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: It should be an int. Do you mean we should update the return type of the function?

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/TARCompress.java
hhbyyh: moved the function to Scala.

hhbyyh:
> Instead of holding and cloning the byte buffer modelbytes for every TorchNet, we need to make sure there is only one copy per JVM, similar to TFNet.graphDef and TFGraphHolder

Thanks. I saw the model registry but not quite sure on some details. I'll talk to Yang and update it.

> Instead of creating a temp model file, can we use 'torch::jit::load(std::istream &in, ..)' to load the model instead?

Tried that but I haven't found a good way to convert from jbytearray to istream. Will log an issue for a follow-up improvement.

> Also add a simple example

sure

Meanwhile, I'm still working on the building with lower version of gcc for cluster compatibility.

hhbyyh:
Use static link to add libstdc and resolved the version conflict.

hhbyyh:
Hi @yangw1234 Do you have a moment to check the modelHolder part? Thanks

yangw1234:
the modelHolder part looks good to me

qiuxin2012:
zoo/src/main/resources/pytorch/lib.zip is 73.9MB...

qiuxin2012:
@jason-dai lib.zip contains a lots of shared object files and statics libraries, so share we move them to zoo-core?

jason-dai:
yes, we should move it to zoo-core

qiuxin2012:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2399/  jenkins passed

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -0,0 +1,84 @@
...
+                                    data_rdd,
+                                    batch_per_thread)
+            return results.map(lambda result: Layer.convert_output(result))
+        else:
+            if isinstance(x, np.ndarray) or isinstance(x, list):
+                results = callBigDlFunc(self.bigdl_type, "zooPredict",
+                                        self.value,
+                                        self._to_jtensors(x),
+                                        batch_per_thread)
+                return [Layer.convert_output(result) for result in results]
```
qiuxin2012: @hhbyyh  I'm confused here, if the input ndarray's size is `[batchsize, data]`, the output should also be ndarray of `[batchsize, output data]`. But I got an list of ndarray of `result`.

jason-dai: @yangw1234 please take a look

yangw1234: This piece of code is copied from `zoo.pipeline.api.keras.engine.topology.py`. I think we should return a list if input is list and a ndarray if input is a ndarray.

yangw1234: maybe `np.stack` if input is a ndarray.

PR #1361, opened 2019-05-22
fix -p mac and windows

yangw1234:


yangw1234:
nb passed http://10.239.47.211:18888/view/ZOO-NB/job/ZOO-NB-Deploy-PYTHON/439/console

PR #1391, opened 2019-05-27
Incremental Training for imagenet

qiuxin2012:
With incremental training, we can use less DRAM during the training.
The training workload is like this:
1. Persist origin data on disk
2. Sample a piece of data to memory
3. train these cached samples
4. unpersist these cached samples
5. Goto 2.

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
@@ -36,12 +36,16 @@ case class EveryEpoch() extends ZooTrigger{
       if (state[Int]("epoch") <= lastEpoch) {
         false
       } else {
-        if (zooState.contains("numSlice") && zooState.contains("currentSlice")
-          && zooState[Int]("currentSlice") % zooState[Int]("numSlice") == 0) {
+        if (zooState.contains("numSlice") && zooState.contains("currentSlice")) {
+          if (zooState[Int]("currentSlice") % zooState[Int]("numSlice") == 0) {
+            lastEpoch = state[Int]("epoch")
+            true
+          } else {
+            false
+          }
+        } else {
           lastEpoch = state[Int]("epoch")
           true
-        } else {
-          false
         }
       }
     }
--- a/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
@@ -3,14 +3,22 @@ package com.intel.analytics.zoo.common
 import com.intel.analytics.bigdl.optim.Trigger
 import com.intel.analytics.bigdl.utils.{T, Table}
 
+/**
+ * A trigger specifies a timespot or several timespots during training,
+ * and a corresponding action will be taken when the timespot(s)
+ * is reached.
+ */
 trait ZooTrigger extends Trigger {
   protected var zooState: Table = T()
 
-  def setZooState(zooState: Table): Unit = {
+  /**
+   * We also hold some training metrics to control trigger.
+   * @param zooState zoo state table
+   */
+  private[Zoo] def setZooState(zooState: Table): Unit = {
     this.zooState = zooState
   }
 }
-
 
 /**
  * A trigger that triggers an action when each epoch finishs.
@@ -48,7 +56,7 @@ case class EveryEpoch() extends ZooTrigger{
  */
 case class SeveralIteration(interval: Int) extends ZooTrigger{
   override def apply(state: Table): Boolean = {
-    val curIteration = state[Int]("neval")
+    val curIteration = state[Int]("neval") - 1
     curIteration != 0 && curIteration % interval == 0
   }
 }
--- a/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
@@ -105,7 +105,15 @@ case class MinLoss(min: Float) extends ZooTrigger {
  * @param first first trigger
  * @param others others triggers
  */
-case class And(first : Trigger, others : Trigger*) extends ZooTrigger {
+case class And(first : ZooTrigger, others : ZooTrigger*) extends ZooTrigger {
+  override def setZooState(zooState: Table): Unit = {
+    super.setZooState(zooState)
+    first.setZooState(zooState)
+    others.foreach{zt =>
+      zt.setZooState(zooState)
+    }
+  }
+
   override def apply(state: Table): Boolean = {
     first.apply(state) && others.forall(_.apply(state))
   }
@@ -116,7 +124,15 @@ case class And(first : Trigger, others : Trigger*) extends ZooTrigger {
  * @param first first trigger
  * @param others others triggers
  */
-case class Or(first : Trigger, others : Trigger*) extends ZooTrigger {
+case class Or(first : ZooTrigger, others : ZooTrigger*) extends ZooTrigger {
+  override def setZooState(zooState: Table): Unit = {
+    super.setZooState(zooState)
+    first.setZooState(zooState)
+    others.foreach{zt =>
+      zt.setZooState(zooState)
+    }
+  }
+
   override def apply(state: Table): Boolean = {
     first.apply(state) || others.exists(_.apply(state))
   }
--- a/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
@@ -15,7 +15,7 @@ trait ZooTrigger extends Trigger {
    * We also hold some training metrics to control trigger.
    * @param zooState zoo state table
    */
-  private[Zoo] def setZooState(zooState: Table): Unit = {
+  private[zoo] def setZooState(zooState: Table): Unit = {
     this.zooState = zooState
   }
 }
--- a/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
@@ -1,0 +1,123 @@
+package com.intel.analytics.zoo.common
+
+import com.intel.analytics.bigdl.optim.Trigger
+import com.intel.analytics.bigdl.utils.{T, Table}
+
+trait ZooTrigger extends Trigger {
+  protected var zooState: Table = T()
+
+  def setZooState(zooState: Table): Unit = {
+    this.zooState = zooState
+  }
+}
+
+
+/**
+ * A trigger that triggers an action when each epoch finishs.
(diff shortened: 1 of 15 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/feature/pmem/NativeArray.scala
```diff
@@ -32,6 +32,8 @@ case object PARTITIONED extends DataStrategy
 
 case object REPLICATED extends DataStrategy
 
+case class INCREMENTAL(cachePercentage: Double) extends DataStrategy
```
jason-dai: It makes more sense to make it a new `MemoryType` (e.g., DISK) instead of `DataStrategy`; e.g., we may later support DISK + REPLICATED

zoo/src/main/scala/com/intel/analytics/zoo/examples/inception/ImageNet2012.scala
```diff
@@ -78,27 +77,24 @@ object ImageNet2012 {
...
-          batchSize = batchSize,
-          transformer = (BytesToBGRImg()
+      val featureSet = FeatureSet.rdd(rawData, memoryType = memoryType, dataStrategy)
+      featureSet.transform(BytesToBGRImg()
             -> BGRImgCropper(imageSize, imageSize)
             -> DatasetHFlip(0.5)
-            -> BGRImgNormalizer(0.485, 0.456, 0.406, 0.229, 0.224, 0.225))
-        ))
+            -> BGRImgNormalizer(0.485, 0.456, 0.406, 0.229, 0.224, 0.225)
+            -> BGRImgToSample()
```
jason-dai: Why add toSample and toBatch here?

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -295,6 +296,58 @@ class CachedDistributedFeatureSet[T: ClassTag]
   }
 }
 
+/**
+ * Wrap a RDD as a FeatureSet.
+ * @param buffer
+ */
+// T is the returning value type. like ByteRecord
+class IncrementalFeatureSet[T: ClassTag]
```
jason-dai: DiskFeatureSet?

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -295,6 +296,58 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+    if (train) {
+      if (currentFeatureSet != null) {
+        currentFeatureSet.unpersist()
+      }
+      currentSlice = buffer.sample(false, cachePercentage)
+        .setName(s"${cachePercentage * 100}% of ${origin.name}")
+      currentFeatureSet = DRAMFeatureSet.rdd(currentSlice)
+      currentFeatureSet.cache()
+      currentFeatureSet.data(train)
+    } else {
```
jason-dai: What does it mean? In prediction/evaluation, only data in `currentFeatureSet` are used?

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -295,6 +296,58 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+    .setName("Origin Data Cached on Disk")
+  protected lazy val count: Long = buffer.count()
+
+  protected var currentSlice = buffer.sample(false, cachePercentage)
+  protected var currentFeatureSet: DistributedFeatureSet[T] = FeatureSet.rdd(currentSlice)
+
+  override def data(train: Boolean): RDD[T] = {
+    if (train) {
+      if (currentFeatureSet != null) {
+        currentFeatureSet.unpersist()
```
jason-dai: Why do you init `currentFeatureSet` above and then discard it immediately here? maybe just init `currentFeatureSet` to null here?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -1005,7 +1239,38 @@ private[zoo] class InternalDistriOptimizer[T: ClassTag] (
...
+        state("epoch") = trueEpoch
+        val newEndWhen = Trigger.or(endWhen, Trigger.maxIteration(i * iterations),
+          Trigger.maxEpoch(trueEpoch))
+        this.setEndWhen(newEndWhen)
+        if (checkPointTrigger.isDefined && checkpointDir.isDefined) {
+          // we should setCheckpoint every time before we call optimize(),
+          // as BigDL will overwrite checkpointPath to its subfolder.
+          this.setCheckpoint(checkpointDir.get, checkPointTrigger.get)
+        }
+        this.train()
```
jason-dai: We should set `state("epoch")` to `trueEpoch` immediately after train, otherwise `endWhen(state)` can exit the loop too early?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -955,8 +999,197 @@ private[zoo] class InternalDistriOptimizer[T: ClassTag] (
     _criterion: Criterion[T])
   (implicit ev: TensorNumeric[T]) extends DistriOptimizer[T](_model, _dataset, _criterion)
   with AbstractEstimator[T]{
-  import InternalDistriOptimizer.logger
+  import InternalDistriOptimizer._
   protected var checkpointDir: Option[String] = None
+  protected var cachePercentage: Double = 1.0
```
jason-dai: I don't think you should make `cachePercentage` a variable in `Estimator`; instead, make it part of `FeatureSet`, and `Estimator` can then check its value in train.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -955,8 +999,197 @@ private[zoo] class InternalDistriOptimizer[T: ClassTag] (
...
+            throw t
+          }
+      }
+    }
+
+    InternalOptimizerUtil.getModel(
+      cachedModels, parameters, trainingModel)
+
+    trainingModel
+  }
```
jason-dai: Do we need to call `optimizer.close()` here?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -955,8 +999,197 @@ private[zoo] class InternalDistriOptimizer[T: ClassTag] (
...
+
+  override def close(): Unit = {
```
jason-dai: Shall we call `close` automatically when `train` is done?

zoo/src/main/scala/com/intel/analytics/zoo/examples/inception/ImageNet2012.scala
qiuxin2012: As the origin rdd is persisted on disk, when count the size of the dataset, MTLabeledBGRImgToBatch will through an exception.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
qiuxin2012: optimizer.shutdown integrate to close()

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/estimator/Estimator.scala
```diff
@@ -134,15 +121,22 @@ class Estimator[T: ClassTag] private[zoo](
...
-        case d: DistributedFeatureSet[MiniBatch[T]] =>
-          new InternalDistriOptimizer[T](model, null, criterion)
+    trainSet match {
+      case d: DistributedFeatureSet[MiniBatch[T]] =>
+        if (internalEstimator == null) {
+          internalEstimator = new InternalDistriOptimizer[T](model, null, criterion)
             .setCheckpointDir(modelDir)
             .setOptimMethods(optimMethods)
+        }
+        if (d.originSet().isInstanceOf[DiskFeatureSet[Any]]) {
```
jason-dai: maybe we can define a method `memoryPercentage` in `FeatureSet` which returns 1 by default, and returns `cachePercentage` in DiskFeatureSet, and then you can just call this method inside `internalEstimator.train`

qiuxin2012:
Add ZooTrigger to fix every epoch won't stop at correct iteration.

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -295,6 +304,90 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+          "because numSlice is zero. numSlice should >= 2 " +
+          "in a training FeatureSet.")
+      } else {
+        buffer
+      }
+    } else {
+      if (train) {
+        if (currentFeatureSet != null && trained) {
+          currentFeatureSet.unpersist()
+          newSample()
```
zhichao-li: Wondering does the user need to enlarge the training epoch accordingly as it seems like the epoch is calculated as per num slice?

qiuxin2012: We will reset the epoch number if the epoch is not ended actually.

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -295,6 +304,90 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+        }
+        currentFeatureSet.cache()
+        trained = true
```
cyita: CurrentFeatureSet should be shuffled here.

zoo/src/main/scala/com/intel/analytics/zoo/common/ZooTrigger.scala
```diff
@@ -0,0 +1,147 @@
...
+    this.zooState = zooState
+  }
+}
+
+/**
+ * A trigger that triggers an action when each epoch finishs.
+ * Could be used as trigger in setValidation and setCheckpoint
+ * in Optimizer, and also in TrainSummary.setSummaryTrigger.
+ */
+case class EveryEpoch() extends ZooTrigger{
```
cyita: If memoryType DRAM is used, the validation phase is ignored at the end of each epoch.

qiuxin2012: fixed

zoo/src/main/scala/com/intel/analytics/zoo/examples/inception/ImageNet2012.scala
qiuxin2012: reverted, As I won't count the hold rdd on disk.

qiuxin2012:
jenkins passed

PR #1396, opened 2019-05-27
[backport] Fix tfpark model batch size

yangw1234:


yangw1234:
no critical issue with work around

PR #1400, opened 2019-05-28
Use smaller data for vae

yangw1234:


hkvision:
I will backport all the related PRs into one.

PR #1407, opened 2019-05-29
Support gradient clipping in TFPark

### LINKED ISSUE #1281: TFEstimatorSpec cannot support gradient clipping (and other customized train_ops)
Currently `TFEstimatorSpec` can only support construct from `loss` when training.  This might cause trouble when migrating some cases with **customized training ops**, like **gradient clipping** used in [chines_ocr](https://github.com/YCG09/chinese_ocr/blob/master/ctpn/lib/fast_rcnn/train.py#L106-L111) as follows:
```python
tvars = tf.trainable_variables()
grads, norm = tf.clip_by_global_norm(tf.gradients(total_loss, tvars), 10.0)
train_op = opt.apply_gradients(list(zip(grads, tvars))
sess.run(train_op)
```
Maybe we might support passing `train_op` into `TFEstimatorSpec` just like [tf.estimator.EstimatorSpec](https://www.tensorflow.org/api_docs/python/tf/estimator/EstimatorSpec) ? Or we might at least support **gradient clipping** so that migration of `chinese_ocr` could be correct.

yangw1234:
fixed #1281
In the keras case, if user set `clipNorm` and/or `clipValue` in tf.keras.Optimizers, these values will be set automatically.
In the estimator case, user use
```
estimator.set_constant_gradient_clipping(min, max)
estimator.set_gradient_clipping_by_l2_norm(clip_norm)
```

### DIFF
--- a/pyzoo/zoo/pipeline/api/net/tf_optimizer.py
+++ b/pyzoo/zoo/pipeline/api/net/tf_optimizer.py
@@ -425,6 +425,24 @@ with variable_creator_scope():
     def set_val_summary(self, summary):
         self.optimizer.set_val_summary(summary)
 
+    def set_constant_gradient_clipping(self, min_value, max_value):
+        """
+        Configure constant clipping settings.
+
+
+        :param min_value: the minimum value to clip by
+        :param max_value: the maxmimum value to clip by
+        """
+        self.optimizer.set_gradclip_const(min_value, max_value)
+
+    def set_gradient_clipping_by_l2_norm(self, clip_norm):
+        """
+        Configure L2 norm clipping settings.
+
+
+        :param clip_norm: gradient L2-Norm threshold
+        """
+        self.optimizer.set_gradclip_l2norm(clip_norm)
 
     def optimize(self, end_trigger=None):
         if end_trigger is None:
--- a/pyzoo/zoo/pipeline/api/net/tf_optimizer.py
+++ b/pyzoo/zoo/pipeline/api/net/tf_optimizer.py
@@ -252,7 +252,8 @@ with variable_creator_scope():
 
             if not isinstance(clip_value, tuple):
                 raise ValueError("The clip_value argument should be" +
-                                 " a positive float/int which clips to (-clip_value, clip_value); " +
+                                 " a positive float/int which clips to" +
+                                 " (-clip_value, clip_value); " +
                                  "or a tuple which clips to (min_value, max_value)")
 
         return cls(*(args + [val_split]),
--- a/pyzoo/zoo/pipeline/api/net/tf_optimizer.py
+++ b/pyzoo/zoo/pipeline/api/net/tf_optimizer.py
@@ -310,8 +310,8 @@ with variable_creator_scope():
                    grads, variables, loss.graph, val_outputs, val_labels,
                    bigdl_val_methods, val_spilt,
                    tensors_with_value=tensor_with_value,
-                   clipnorm=clip_norm,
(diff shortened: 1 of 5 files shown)


pyzoo/zoo/pipeline/api/net/tf_optimizer.py
```diff
@@ -200,6 +203,12 @@ def to_floats(vs):
                                               batch_size=batch_size,
                                               optim_method=self.optim_method)
 
+        if self.clip_norm:
+            self.optimizer.set_gradclip_l2norm(self.clip_norm)
+        if self.clip_constant:
+            min_value, max_value = self.clip_constant
```
hkvision: Add some document or assert to tell users that clip_value should be a tuple?

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
```diff
@@ -227,12 +236,15 @@ def _get_arguments_from_loss(loss, optim_method, session, val_outputs, val_label
 
     @classmethod
     def from_loss(cls, loss, optim_method, session=None, val_outputs=None,
-                  val_labels=None, val_method=None, val_split=0.0, **kwargs):
+                  val_labels=None, val_method=None, val_split=0.0,
+                  clip_norm=None, clip_value=None, **kwargs):
```
hkvision: is this needed here? as users can always specify them in `kwargs`

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
yangw1234: sure

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
```diff
@@ -58,7 +58,8 @@ class TFOptimizer:
     def __init__(self, loss, optim_method, sess=None, dataset=None, inputs=None,
                  grads=None, variables=None, graph=None,
                  val_outputs=None, val_labels=None, val_method=None, val_split=0.0,
-                 tensors_with_value=None, session_config=None):
+                 tensors_with_value=None, session_config=None,
+                 clip_norm=None, clip_value=None):
```
jason-dai: Why don't we use `set_constant_gradient_clipping` or `set_gradient_clipping_by_l2_norm` for `TFOptimizer`?

hkvision:
LGTM

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
yangw1234: this is for the keras case, in which the gradient clipping information is hidden in the keras.optimizer

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
yangw1234: when they jump to the source, they will see this option clearly

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
jason-dai: But users of TFOptimizer can also use the gradient clipping support if they want to, yes?

yangw1234: I'll add those methods. Just override the constructor arguments.

yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2329/

PR #1419, opened 2019-06-03
Update TFDataset document

yangw1234:


yangw1234:
jenkins pass , http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2334/
please leave a message if you still have any comments

ImCharlesY:
Hi, I am Charles, the author of [Is there any way to release the memory of rdds that are no longer needed?](https://groups.google.com/forum/#!topic/bigdl-user-group/iSGrruoBoe0). It seems that my replies would be automatically deleted on the user group, thus I failed to reply to the topic. Therefore, please allow me to leave a comment here. I am glad that I can help. I have reviewed the pr and found a mistake as follows.

At line 230, 232 and 238 in the [net.md](https://github.com/intel-analytics/analytics-zoo/pull/1419/files#diff-956f09ecdf10ebe5ef99fe3abe2ffc32R230) file, it seems that the `shape` argument of the second np.ndarray element should be specified as `(20,)` according to the `features` and `labels` arguments.

FYI, also in the [net.md](https://github.com/intel-analytics/analytics-zoo/pull/1419/files#diff-956f09ecdf10ebe5ef99fe3abe2ffc32R238) file, I think that the example given in line 235-239 should be moved to after line 242. Because the `labels` argument should be specified for training and evaluation, it seems there is not a case where the `labels` argument is not specified in training and evaluation. If the element of rdd is given as line 238 showing, then the `fit` and `evaluate` methods of KerasModel would treat the second np.ndarray element as label according to the logic in source code.

`line 238:        [np.zeros(dtype=float, shape=(10,), np.zeros(dtype=float, shape=(20,)))]`

The same applies to the [tf_dataset.py](https://github.com/intel-analytics/analytics-zoo/pull/1419/files#diff-90d6209a3b6f4f6f24e84e7cb92436c4R309) file.

Thanks,
Charles

yangw1234:
@ImCharlesY thanks Charles, I'll update the doc.

PR #1422, opened 2019-06-03
[WIP] automl and automated time sequence predictor

shane-huang:
- API's of feature transformers, model, search engine, pipeline, and time sequence predictor
- an jupyter notebook demonstrating the usage on nytaxi data.


pyzoo/zoo/automl/model/VanillaLSTM.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        Evaluate on x, y
+        :param x: input
+        :param y: target
+        :param metric: a list of metrics in string format
+        :return: a list of metric evaluation results
+        """
+        e = Evaluator()
+        y_pred = self.predict(x)
+        return [e.evaluate(m, y, y_pred) for m in metric]
```
shane-huang: pipeline evaluate needs to be inverse_transformed.

shane-huang:
NY taxi data needs to be handled the same way as anomaly detection example.

shane-huang:
Need to verify if predict datetime is correct. Might change to starting point of future instead of the last timestamp of training - Since most of the time predict seq = 1, it is more natural to use the target datetime.

shane-huang:
notebook will show predicting future seq len > 1.

shane-huang:
multiple metrics with future_seq_len > 1 better return list of tuples.

shane-huang:
there's some problem with this pr and we moved to a new one https://github.com/intel-analytics/analytics-zoo/pull/1443

PR #1448, opened 2019-06-12
update docker to support spark master

glorysdj:
this pr is to add sprk master url support for docker image

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/1448)


docker/zoo/start-notebook.sh
```diff
@@ -30,15 +30,42 @@ echo $SPARK_MAJOR_VERSION
...
+
+# verbose for OpenMP
+if [[ $* == *"verbose"* ]]; then
+    export KMP_SETTINGS=1
+    export KMP_AFFINITY=${KMP_AFFINITY},verbose
+fi
+
 ${SPARK_HOME}/bin/pyspark \
-  --master local[${RUNTIME_EXECUTOR_CORES}] \
+  --master ${RUNTIME_SPARK_MASTER} \
```
jenniew: How can we also specify the executor cores and executor number?

glorysdj: in readme

sudo docker run -itd --net=host \
    -e NotebookPort=12345 \
    -e NotebookToken="1234qwer" \
    -e http_proxy=http://your-proxy-host:your-proxy-port  \
    -e https_proxy=https://your-proxy-host:your-proxy-port  \
    -e RUNTIME_SPARK_MASTER=spark://your-spark-master-host:your-spark-master-port or local[*] \
    -e RUNTIME_DRIVER_CORES=4 \
    -e RUNTIME_DRIVER_MEMORY=20g \
    -e RUNTIME_EXECUTOR_CORES=4 \
    -e RUNTIME_EXECUTOR_MEMORY=20g \
    -e RUNTIME_TOTAL_EXECUTOR_CORES=4 \
    intelanalytics/analytics-zoo:latest

zhichao-li: How about pack the whole `pyspark` parameters into a variable rather than passing them into docker one by one? The idea is it's too verbose to iterate all of the pyspark options in docker and user can pass any parameters supported by pyspark as they wish in this way.
i.e
Before using docker, user would write things like
```
pyspark --master xxx --num-executors 4 --executor-cores 3 ...
```
After
```
docker  -e pyspark_args="--master xxx --num-executors 4 --executor-cores 3"
```

glorysdj: yes, have considered this idea, the cons is we need to check the args, if we need more config,  we will change to pyspark_args way.

jenniew: I think this way may be better as users can pass some customized config for their special case. If it failed, we provide them error message.

glorysdj: ok, i will open another pr to do this.

PR #1451, opened 2019-06-12
support sparse parameter in wideanddeep model

dding3:


zoo/src/main/scala/com/intel/analytics/zoo/common/SparseParameterProcessor.scala
```diff
@@ -0,0 +1,73 @@
...
+                                    parameters: AllReduceParameter[T],
+                                    metrics: Metrics,
+                                    state: Table)(implicit ev2: TensorNumeric[T]) : Unit = {
+    // 1. aggregate sparseG first in each node
+    // 2. aggregate sparseG on driver side
+    globalSparseG = models.mapPartitions(modelIter => {
+      val cached = modelIter.next()
+      val sparseG = cached.localModels.map(
+        // TODO: remove asInstanceOf
+        _.asInstanceOf[sudoLookupTableSparse[U]].getSparseParameters()._2)
```
yangw1234: Can we train a model that only a subsets of the gradients are sparse?

zoo/src/main/scala/com/intel/analytics/zoo/common/SparseParameterProcessor.scala
```diff
@@ -0,0 +1,73 @@
...
+          i += 1
+        }
+        Iterator(res)
+      } else Iterator(sparseG.head)
+    }).reduce(SparseTensorUtils.addSparseTensor[U](_, _))
+
+    //TODO: support cliping sparseG
+
+    // update weight with global gradients in driver
+    if (globalSparseW == null) {
```
yangw1234: I think for a embedding layer, the weight is dense, only the gradient is sparse.

zoo/src/main/scala/com/intel/analytics/zoo/common/SparseTensorUtils.scala
```diff
@@ -0,0 +1,94 @@
...
+      indice ++= t2Indice1.slice(j, t2Indice1.length)
+      indice2 ++= t2Indice2.slice(j, t2Indice1.length)
+      value ++= t2Value.slice(j, t2Indice1.length)
+    }
+    SparseTensor(Array(indice.toArray, indice2.toArray), Storage[T](value.toArray),
+      sparseT._shape, sparseT._shape.length)
+  }
+
+  def addConstant[T: ClassTag](tensor: Tensor[T], constant: T)
+                              (implicit ev: TensorNumeric[T]): SparseTensor[T] = {
```
yangw1234: Shouldn't the zero entries of the `tensor` be also added by the `constant`? This will make it a dense tensor, I think.

zoo/src/main/scala/com/intel/analytics/zoo/common/SparseParameterProcessor.scala
dding3: Yes, put the sparse part in getSparseParameters and dense in getParmeters.

zoo/src/main/scala/com/intel/analytics/zoo/common/SparseParameterProcessor.scala
dding3: Rename `globalSparseW` to `globalW` to make less confusion. Updating `weight` logic should be same here.

zoo/src/main/scala/com/intel/analytics/zoo/common/SparseTensorUtils.scala
dding3: Make sense. Currently this function is a used for test into `sudoSgd`, and used to only update the non zero part. Rename it to `addSparseTensorValueByConstant` to make more sense. :)

yangw1234:
In the embedding case, the gradient of a embedding layer is not totally sparse but a few slice of a dense tensor. Say the batch size is 4 and the embedding size is 256, then the gradient is better represented as 4 dense tensor each of size 256.
Maybe we should take that into consideration.

PR #1460, opened 2019-06-14
We should search and find the necessary jars from python env rather than upload them again

zhichao-li:


zhichao-li:
Build # 2381 - Successful

PR #1468, opened 2019-06-21
Estimator instead  the usage of optimizer in TextClassificationTrainer and NeuralCFSpec

cyita:
Remove optimizer in app and testcase.


zoo/src/test/scala/com/intel/analytics/zoo/models/recommendation/NeuralCFSpec.scala
```diff
@@ -110,16 +112,15 @@ class NeuralCFSpec extends ZooSpecHelper {
...
-      .setOptimMethod(new Adam[Float](learningRate = 1e-2, learningRateDecay = 1e-5))
-      .setEndWhen(Trigger.maxEpoch(100))
-      .optimize()
+    ncf.compile(optimizer = optimMethod,
+      loss = SparseCategoricalCrossEntropy[Float](zeroBasedLabel = false),
+      metrics = List(new Top1Accuracy[Float]()))
+
+    ncf.fit(trainRdds, batchSize = 458, nbEpoch = 100)
+
+    ncf.summary()
```
songhappy: Why you want to change it to use fit and compile? There is one test case to use compile and fit. It is intentionally to keep a couple of testcases to use optimizer cause it is the requirement from the beginning.

zoo/src/test/scala/com/intel/analytics/zoo/models/recommendation/NeuralCFSpec.scala
```diff
@@ -145,16 +146,15 @@ class NeuralCFSpec extends ZooSpecHelper {
...
-      sampleRDD = trainRdds,
-      criterion = ClassNLLCriterion[Float](),
-      batchSize = 458)
+    val optimMethod = new Adam[Float](learningRate = 1e-2, learningRateDecay = 1e-5)
+
+    ncf.compile(optimizer = optimMethod,
+      loss = SparseCategoricalCrossEntropy[Float](zeroBasedLabel = false),
+      metrics = List(new Top1Accuracy[Float]()))
+
+    ncf.fit(trainRdds, batchSize = 458, nbEpoch = 100)
```
songhappy: again, why do you want to change it to use fit and compile?

zoo/src/test/scala/com/intel/analytics/zoo/models/recommendation/NeuralCFSpec.scala
qiuxin2012: So this is a unit test to make sure "compile and fit" API and Optimizer API will return the same result?

cyita: Because I want to replace Optimizer with analytics-zoo API, and "compile and fit" is used in NeuralCFexample. If this UT aims to make sure both usage will work properly, I will use Estimator in these two cases.

jason-dai: Yes, we can just use Estimator in the tests

cyita: Done

zoo/src/test/scala/com/intel/analytics/zoo/models/recommendation/NeuralCFSpec.scala
```diff
@@ -183,14 +187,46 @@ class NeuralCFSpec extends ZooSpecHelper {
...
+
+    val pairPredictionsEst = ncfEst.predictUserItemPair(data)
+    val pairPredictionsEstDF = sqlContext.createDataFrame(pairPredictionsEst).toDF()
+    val outEst = pairPredictionsEstDF.join(datain, Array("userId", "itemId"))
+    val correctCountsEst = outEst.filter(col("prediction") === col("label")).count()
 
+    val accuracyEst = correctCountsEst.toDouble / 458
+
+    // Use compile and fit API
+    val ncf = NeuralCF[Float](100, 100, 5, 5, 5, Array(10, 5), false)
```
qiuxin2012: Use `ncfEst.cloneModule()` method to create this model, so the starting will be the same.

zoo/src/test/scala/com/intel/analytics/zoo/models/recommendation/NeuralCFSpec.scala
```diff
@@ -183,14 +187,46 @@ class NeuralCFSpec extends ZooSpecHelper {
...
     ncf.compile(optimizer = optimMethod,
```
qiuxin2012: Please don't reuse this optimMethod, some buffers are not clear. Please new a new one.

Issue #1471, opened 2019-06-24
jupyter-with-zoo.sh doesn't work


qiuxin2012:
```
xin@xin-dev:~/IntelAnalytics/analytics-zoo6$ bash ${ANALYTICS_ZOO_HOME}/bin/jupyter-with-zoo.sh --master local[*]
Traceback (most recent call last):
  File "/home/xin/.local/lib/python3.6/site-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'allow_remote_access'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/xin/.local/lib/python3.6/site-packages/notebook/notebookapp.py", line 864, in _default_allow_remote
    addr = ipaddress.ip_address(self.ip)
  File "/usr/lib/python3.6/ipaddress.py", line 54, in ip_address
    address)
ValueError: '' does not appear to be an IPv4 or IPv6 address

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/xin/.local/bin/jupyter-notebook", line 11, in <module>
    sys.exit(main())
  File "/home/xin/.local/lib/python3.6/site-packages/jupyter_core/application.py", line 266, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/home/xin/.local/lib/python3.6/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "</home/xin/.local/lib/python3.6/site-packages/decorator.py:decorator-gen-7>", line 2, in initialize
  File "/home/xin/.local/lib/python3.6/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/home/xin/.local/lib/python3.6/site-packages/notebook/notebookapp.py", line 1628, in initialize
    self.init_webapp()
  File "/home/xin/.local/lib/python3.6/site-packages/notebook/notebookapp.py", line 1378, in init_webapp
    self.jinja_environment_options,
  File "/home/xin/.local/lib/python3.6/site-packages/notebook/notebookapp.py", line 159, in __init__
    default_url, settings_overrides, jinja_env_options)
  File "/home/xin/.local/lib/python3.6/site-packages/notebook/notebookapp.py", line 252, in init_settings
    allow_remote_access=jupyter_app.allow_remote_access,
[...]

Issue #1340, opened 2019-05-21
Jupyter notebook script ip=* does not work on some versions of jupyter


yangw1234:
Should we change to ip=0.0.0.0?

PR #1476, opened 2019-06-25
update jupyter-with-zoo.sh 's bounding ip

qiuxin2012:
change from `ip=*` to `ip=0.0.0.0`

fix issue:  #1471  #1340


yangw1234:
LGTM

PR #1478, opened 2019-06-25
Fix tfdataset cannot work with list input

yangw1234:


yangw1234:
http://172.168.2.101:8080/job/ZOO-PR-Validation/2403/
jenkins pass

Issue #1477, opened 2019-06-25
resnet50: torchnet's inference result is not the same with pytorch's result


qiuxin2012:
test code
```
import torchvision
import numpy as np
from zoo.pipeline.api.net.torch_net import TorchNet

from zoo.common.nncontext import init_nncontext
sc = init_nncontext("Object Detection Example")

input = np.random.rand(1, 3, 224, 224)

model = torchvision.models.resnet18().eval()
net = TorchNet.from_pytorch(model, [1, 3, 224, 224])

result = net.predict(input)
print(result.collect())

import torch
with torch.no_grad():
    print(model(torch.Tensor(input)))
```

torchnet's result:
```
[array([ 3.74878109e-01, -1.31922567e+00, -8.85901034e-01, -7.12347269e-01,
        1.06809698e-01,  4.45175201e-01, -6.62672818e-01,  6.14829510e-02,
        2.42397055e-01,  1.10141337e-01, -4.90026206e-01, -1.46618700e+00,
       -1.06230453e-01, -1.46709657e+00,  1.05173290e+00,  5.44399433e-02,
       -5.13154387e-01,  7.99532235e-01, -4.09477279e-02,  1.23269868e+00,
       -1.68987060e+00,  5.49524069e-01, -1.12183675e-01,  1.85302690e-01,
       -2.68771768e-01, -3.74099277e-02, -1.08390200e+00, -7.65769780e-01,
       -6.62536696e-02, -4.39484090e-01, -8.17326486e-01,  1.62058151e+00,
       -3.69018227e-01,  4.17334229e-01,  1.18404031e+00,  1.27722752e+00,
       -1.17876387e+00,  7.46984422e-01,  1.41101200e-02,  3.70763421e-01,
       -5.52040637e-01, -1.05381072e+00,  7.77815938e-01,  9.47480559e-01,
        3.98006558e-01, -5.29138744e-01,  1.87831652e+00,  2.61362553e-01,
       -5.30600786e-01,  7.67063141e-01, -6.36106879e-02, -9.93039846e-01,
       -1.76547747e-02, -4.61912900e-01, -5.79799414e-01, -1.93639994e-01,
        6.19476020e-01,  2.25750938e-01, -2.67607808e-01,  2.85148621e-01,
        4.91327077e-01, -7.21228659e-01, -2.97347218e-01, -1.85416669e-01,
[...]

PR #1481, opened 2019-06-26
[Pytorch] correct offset for inference

hhbyyh:
BigDL Tensor offset starts from 1 by default.

fix issue https://github.com/intel-analytics/analytics-zoo/issues/1477

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
@@ -63,7 +63,7 @@ class TorchNet private(private val modelHolder: TorchModelHolder)
     require(input.isContiguous())
     val data = input.storage().array()
     val size = input.size()
-    val offset = input.storageOffset()
+    val offset = input.storageOffset() - 1
     val result = forward(data, offset, size)
     val resultTensor = Tensor(result.getData, result.getShape)
     output.set(resultTensor)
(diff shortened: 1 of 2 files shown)


qiuxin2012:
@hhbyyh It works. I think you should make the testing code to be a unit test.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -63,7 +63,7 @@ class TorchNet private(private val modelHolder: TorchModelHolder)
     require(input.isContiguous())
     val data = input.storage().array()
     val size = input.size()
-    val offset = input.storageOffset()
+    val offset = input.storageOffset() - 1
```
zhichao-li: It might be ideal if we can have an interface like: ` pytorchModel.forward(zoo_tensor)` then there's zero chance to use the offset incorrectly. Also, do we need to check if the input tensor is continuous or not or it's a trade-off for performance?

1. 

?

zhichao-li: Oops, checking has already been added here.

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2409/ passed.

hhbyyh:
Merging to master. Thanks for the review

PR #1482, opened 2019-06-26
add face generation example using torchnet

qiuxin2012:


jason-dai:
1) change to apps/pytorch
2) add a readme on where this notebook is migrated from

qiuxin2012:
Update the notebook, the main change is run pytorch to generate 64 faces, then load the pytorch model to analytics-zoo and generate 64 faces on spark.

apps/pytorch/README.md
```diff
@@ -0,0 +1,31 @@
...
+pip install analytics-zoo==0.6.0.dev4
+```
+
+## Run Jupyter
+If you want to run spark local, just start jupyter notebook:
+```
+jupyter notebook
+```
+
+If you want to run on a yarn cluster(yarn-client mode), export env `HADOOP_CONF_DIR` and `ZOO_CONDA_NAME` before starting jupyter notebook.
```
jason-dai: Do we need the python dependency on workers? If not, we don't need coda environment.

apps/pytorch/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+export HADOOP_CONF_DIR=[path to your hadoop conf directory]
+export ZOO_CONDA_NAME=[conda name intalled zoo]
```
jason-dai: How does the user set up the above env variable? Please add appropriate guide or external reference.

apps/pytorch/README.md
qiuxin2012: Yes, we use python to create data RDD.

PR #1485, opened 2019-06-27
Compare API difference between Ray and RayOnSpark 

xinzhu-cai:


HzhElena:
Modified rl_pong, parameter_server, multiagent_two_trainers from using ray APIs to RayOnSpark version

jason-dai:
In general this looks good. Several comments:
1) Please add a simple readme on how to run these examples
2) Please update the license header (similar to https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/util/tf_graph_util.py)
3) Do we have any support for testing Ray programs? @zhichao-li @shane-huang

PR #1488, opened 2019-06-28
Using non-freezing graph for TF training

yangw1234:
- [x] Separate TF graph execution logic (including session, BigDL tensor to TF tensor, variable initialization) into another object (GraphRunner)
- [x] Using GraphRunner in TFTrainingHelper
- [x] Change python export_tf to accommodate scala changes.
- [x] Support BatchNorm
- [x] testing on cluster
- [x] Convergence
resnet32 cifar-10 converge on a single node with one model multiple threads mode

image

Maximum accuracy on the test set reached 92.89%. (92.49 reported on the resnet paper)

- [x] performance testing
image
16 node and two models on each node using 22 threads with a batch size of 4096
slightly better than previous implementation


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2610/ jenkins pass

yangw1234:
tested resnet32 on cifar10

yangw1234:
tested on iqiyi model and data

PR #1491, opened 2019-06-28
IBM ICPD support

jenniew:
IBM ICPD tile support.
Add helm charts for deployment and installation documentation


jason-dai:
We can host the document on analytics-zoo.github.io. However, I don't think we should store helm chart related files in the project website; we can create a separate repo on github if needed.

jenniew:
Has moved to https://github.com/intel-analytics/zoo-icpd

PR #1498, opened 2019-07-05
[WIP] Pytorch Training prototype

hhbyyh:
Send a PR for discussion and suggestions.

Right now this use TorchModel for both training and inference. For training, user need to set the loss function as in the python example. 

We may separate the loss/criterion to a new class TorchCriterion for the next step. After that, we may combine torch model and bigdl/AZ criterion.


jason-dai:
Is this rebased with the latest master? Seems it still includes libtorch library.

hhbyyh:
Rebase this to master.

Since the function is still under active development, separating it to zoo-core would bring extra effort in review and additional risk for build break. For now I would suggest to keep the C++ code in the same repo and we may transfer it to zoo-core after function stabilized.

I changed it to download libtorch during build so we don't need to check in the large binary files.

pyzoo/test/zoo/pipeline/api/test_torch_net.py
```diff
@@ -35,6 +36,75 @@ def test_torch_net_predict(self):
...
+        criterion = nn.MSELoss()
+
+        torch_output = model.forward(torch_input)
+        torch_loss = criterion.forward(torch_output, torch_label)
+        torch_loss.backward()
+        torch_grad = model.dense1.weight.grad.flatten().tolist() + model.dense1.bias.grad.flatten().tolist() + \
+                     model.conv1.weight.grad.flatten().tolist() + model.conv1.bias.grad.flatten().tolist() + \
+                     model.dense2.weight.grad.flatten().tolist() + model.dense2.bias.grad.flatten().tolist()
+
+        torch_net = TorchNet.from_pytorch(model, [1, 2], criterion.forward, pred_shape=[1, 1], label_shape=[1, 1])
```
jason-dai: can it be simply `TorchNet.from_pytorch(model, inputShape, criterion, outputShape)`? Users can optionally specify `labelShape` if needed.

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -14,17 +14,32 @@
...
 
 from pyspark import RDD
 from bigdl.nn.layer import Layer
+from bigdl.nn.criterion import Criterion
 from zoo import getOrCreateSparkContext
 from zoo.feature.image import ImageSet
 from bigdl.util.common import callBigDlFunc
 from zoo.pipeline.api.net.tfnet import to_sample_rdd
 
+class TempModule(nn.Module):
```
jason-dai: `FunctionModule` or `FuncModule`?

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -14,17 +14,32 @@
...
+    def __init__(self, lossFunc):
+        super(TempModule, self).__init__()
+        self.func = lossFunc
+
+    def forward(self, x, y):
+        return self.func(x, y)
+
+class TorchIdentityCriterion(Criterion):
```
jason-dai: What's the difference vs. `IdentityCriterion`? @yangw1234

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -33,21 +48,35 @@ class TorchNet(Layer):
...
 
-    def __init__(self, path, bigdl_type="float"):
-        super(TorchNet, self).__init__(None, bigdl_type, path)
+    def __init__(self, path, lossPath=None, bigdl_type="float"):
+        super(TorchNet, self).__init__(None, bigdl_type, path, lossPath)
 
     @staticmethod
-    def from_pytorch(module, input_shape):
+    def from_pytorch(module, input_shape, lossFunc=None, pred_shape=None, label_shape=None,
+                     pred_input=None, label_input=None):
```
jason-dai: Why do we need to support `pred_input` and `label_input`

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -138,21 +179,106 @@ object TorchNet {
...
+        System.out.println("error during loading Torch model")
+        throw io
+      }
     }
+    nativeRef
   }
 
 }
+
+class TorchIdentityCriterion extends AbstractCriterion[Activity, Activity, Float]() {
```
jason-dai: What's difference vs. `IdentityCriterion`?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -17,72 +17,113 @@
...
   @transient
-  private lazy val torchModel = {
+  lazy val nativeRef: Long = {
     println("TorchNet loading in " + this)
-    TorchNet.load(modelHolder.torchBytes)
-  }
+    val ref = if (lossHolder != null)
+      TorchNet.loadPytorchModel(modelHolder.torchBytes, lossHolder.torchBytes)
+    else
+      TorchNet.loadPytorchModel(modelHolder.torchBytes, modelHolder.torchBytes)
```
jason-dai: why do you load pytorch model twice here?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -17,72 +17,113 @@
...
 
-  private def forward(storage: Array[Float], offset: Int, shape: Array[Int]): JTensor = {
-    PytorchModel.forwardNative(this.torchModel, storage, offset, shape).asInstanceOf[JTensor]
+    if (weights == null) {
+      val w = PytorchModel.getWeightNative(ref).clone()
+      weights = Tensor(w, Array(w.length))
+      gradients = Tensor(Array.tabulate(w.length)(i => 0f), Array(w.length))
```
jason-dai: If the model is only used for inference, we do not need to create gradient here

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -17,72 +17,113 @@
...
+    } else{
+      PytorchModel.updateWeightNative(ref, weights.storage().array())
```
jason-dai: DOes it make send to update the pytorch model weight here?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -17,72 +17,113 @@
...
+    val data = gradOutput.storage().array()
+    val size = gradOutput.size()
+    val offset = gradOutput.storageOffset() - 1
+    val result = PytorchModel.backwardNative(nativeRef, data, offset, size)
+    val resultTensor = Tensor(result.getData, result.getShape)
+
+    val g = PytorchModel.getGradientNative(this.nativeRef).clone()
+    System.arraycopy(g, 0, gradients.storage().array(), 0, g.length)
+
+    gradInput.set(resultTensor)
```
jason-dai: `resultTensor` is actually loss, not `gradientInput`?

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+
+JNIEXPORT jlong JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchModel_loadNative
+  (JNIEnv *jenv, jclass jobj, jstring jmodel_path, jstring jloss_path) {
+    const char* p_model_path = jenv->GetStringUTFChars(jmodel_path, NULL);
+    const char* p_loss_path = jenv->GetStringUTFChars(jloss_path, NULL);
+
+    // Deserialize the ScriptModule from a file using torch::jit::load().
+    std::shared_ptr<torch::jit::script::Module> model_ptr = torch::jit::load(p_model_path);
+    std::shared_ptr<torch::jit::script::Module> loss_ptr = torch::jit::load(p_loss_path);
```
jason-dai: we should support if `p_loss_path` is NULL

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -0,0 +1,354 @@
...
+    // Create a vector of inputs.
+    std::vector<torch::jit::IValue> lossInputs;
+
+    at::Tensor y = outputs[nativeRef];
+    lossInputs.push_back(y);
+    lossInputs.push_back(label_tensor);
+
+    std::shared_ptr<torch::jit::script::Module> loss_ptr = lossHandles[nativeRef];
+
+    at::Tensor loss = loss_ptr->forward(lossInputs).toTensor();
```
jason-dai: I think you should still compute loss in the `forward` method of the model or criterion; otherwise `DistriOptimizer` cannot draw the correct training loss chart in, say, TensorBoard.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -0,0 +1,354 @@
...
+    jobject result = jenv -> NewObject(jtensor_class, jtensor_constructor, result_storage, result_shape);
+
+    return result;
+  }
+
+
+JNIEXPORT void JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchModel_releaseNative
+  (JNIEnv * jenv, jclass jobj, jlong nativeRef) {
+    mtx.lock();
+    handles.erase(nativeRef);
```
jason-dai: you also need to release loss and output.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -17,72 +17,113 @@
...
-import scala.collection.mutable
 import scala.reflect.ClassTag
 
 /**
  * [[TorchNet]] wraps a TorchScript model as a single layer.
  */
-class TorchNet private(private val modelHolder: TorchModelHolder)
+class TorchNet private(private val modelHolder: TorchModelHolder,
+                       // TODO: separate it to TorchCriterion?
+                       private val lossHolder: TorchModelHolder)
```
jason-dai: what if `lossHolder` is null, but the model is used in training?

pyzoo/zoo/examples/pytorch/train/SimpleTrainingExample.py
```diff
@@ -0,0 +1,66 @@
...
+    df = get_df(sqlContext)
+
+    torch_model = SimpleTorchModel()
+    becloss = BCELoss()
+
+    model = TorchNet.from_pytorch(module=torch_model,
+                                  input_shape=[1, 2],
+                                  lossFunc=becloss.forward,
+                                  pred_shape=[1, 1], label_shape=[1, 1])
+    classifier = NNEstimator(model, TorchIdentityCriterion(), SeqToTensor([2])) \
```
jason-dai: It does not make sense to ask user to use `TorchIdentityCriterion` explicitly. I think we should have somthing like:

```
model, loss = TorchNet.from_pytorch(module=torch_model, loss=..., ...)
classifier = NNEstimator(model, loss, ...)
```
In this case, we can also use any BigDL/Zoo criterion in the training.

jason-dai:
In addition to the comments inline, `TorchNet` could have either Tensor or Table as input/output; that is, it should extend `AbstractModule[Activity, Activity, Float]`. we can fix this in a follow-up PR.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -0,0 +1,354 @@
...
+            updateWeights(child, xv, index);
+        }
+    }
+
+    return 1;
+}
+
+
+JNIEXPORT jlong JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchModel_loadNative
+  (JNIEnv *jenv, jclass jobj, jstring jmodel_path, jstring jloss_path) {
```
qiuxin2012: ` static native long loadNative(String modelPath, String lossPath);`  two string parameters, but three parameter here. `jobj` is useless?

pyzoo/test/zoo/pipeline/api/test_torch_net.py
hhbyyh: We still need the labelShape for tracing the loss, and I haven't found a way to infer the shape directly from loss.

pyzoo/zoo/pipeline/api/net/torch_net.py
hhbyyh: Sure

pyzoo/zoo/pipeline/api/net/torch_net.py
hhbyyh: IdentityCriterion expects single output `input.toTensor[Float].value()`.

pyzoo/zoo/pipeline/api/net/torch_net.py
hhbyyh: Some torch criterion like `CrossEntropyLoss` expect label of datatype Long. I think we can provide both options (specify shape or sample input)

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: Yes. will check for training.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: removed TorchIdentityCriterion since we have Torch Criterion now.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: temp work around. Corrected.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: changed to gradinput

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Done

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: Use torch criterion now

pyzoo/zoo/examples/pytorch/train/SimpleTrainingExample.py
hhbyyh: Done

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Done

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Done

pyzoo/zoo/pipeline/api/net/torch_criterion.py
```diff
@@ -0,0 +1,66 @@
...
+        :param lossFunc: a function that take two parameters: input and label
+        :param input_shape: list of integers.
+        :param label_shape: list of integers. If not specified, it will be set equal to input_shape
+        :param sample_input: a sample of input.
+        :param sample_label: a sample of label.
+        """
+        temp = tempfile.mkdtemp()
+
+        sample_input = sample_input if sample_input else torch.rand(input_shape)
+        sample_label = sample_label if sample_label else torch.rand(label_shape)
```
jason-dai: what id label_shape is None?

pyzoo/zoo/pipeline/api/net/torch_criterion.py
```diff
@@ -0,0 +1,66 @@
...
+    TorchCriterion wraps a TorchScript model as a single layer, thus the Pytorch model can be used for
+    distributed inference or training.
+    :param path: path to the TorchScript model.
+    """
+
+    def __init__(self, path, bigdl_type="float"):
+        super(TorchCriterion, self).__init__(None, bigdl_type, path)
+
+    @staticmethod
+    def from_pytorch(lossFunc, input_shape, label_shape=None, sample_input=None, sample_label=None):
```
jason-dai: use `loss` instead of `lossFunc`? it's actually a class.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -162,17 +177,19 @@ JNIEXPORT jobject JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchM
     }
     // create a Tensor
     auto torch_input_tensor = torch::from_blob(c_storage + joffset, torch_shape, at::kFloat);
+    torch_input_tensor.set_requires_grad(true);
```
jason-dai: true only in training, not for inference, yes?

pyzoo/zoo/pipeline/api/net/torch_criterion.py
hhbyyh: Thanks. Use input shape implicitly.

pyzoo/zoo/pipeline/api/net/torch_criterion.py
hhbyyh: Sure, update to loss.

this can be a torch loss (e.g. nn.MSELoss()) or a function that take two Tensor parameters: input and label. E.g.
```
def lossFunc(input, target):
        return nn.CrossEntropyLoss().forward(input, target.flatten().long())
```

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Yes. Udpated.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: weights carries the real weight for the model, while the byte array only contains model architecture.

jason-dai: I don't understand the logic here. The weight of PyTorch model should be stored in the native PyTorch model; the `weights` tensor should only contain a copy of the weight on Scala size. Why do we update the native PyTorch model during initialization?

pyzoo/zoo/examples/pytorch/train/resnet_finetune.py
```diff
@@ -0,0 +1,90 @@
...
+from zoo.pipeline.api.net.torch_criterion import TorchCriterion
+from zoo.pipeline.nnframes import *
+from zoo.pipeline.api.keras.metrics import Accuracy
+
+
+# Define model with Pytorch
+class CatDogModel(nn.Module):
+    def __init__(self):
+        super(CatDogModel, self).__init__()
+        self.features = torchvision.models.resnet18(pretrained=True).eval()
```
jason-dai: does it mean that the weights in `self.features` won't be updated during fine tuning?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -54,30 +51,24 @@ class TorchNet private(private val modelHolder: TorchModelHolder,
...
+    val ref = TorchNet.loadPytorchModel(modelHolder.torchBytes)
 
     if (weights == null) {
       val w = PytorchModel.getWeightNative(ref).clone()
       weights = Tensor(w, Array(w.length))
-      gradients = Tensor(Array.tabulate(w.length)(i => 0f), Array(w.length))
+      if(this.isTraining()) {
+        gradients = Tensor(Array.tabulate(w.length)(i => 0f), Array(w.length))
+      } else {
+        gradients = Tensor.ones(1)
```
jason-dai: it can be simply `Tensor()`

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
```diff
@@ -54,30 +51,24 @@ class TorchNet private(private val modelHolder: TorchModelHolder,
...
-      TorchNet.loadPytorchModel(modelHolder.torchBytes, lossHolder.torchBytes)
-    else
-      TorchNet.loadPytorchModel(modelHolder.torchBytes, modelHolder.torchBytes)
+    val ref = TorchNet.loadPytorchModel(modelHolder.torchBytes)
 
     if (weights == null) {
       val w = PytorchModel.getWeightNative(ref).clone()
       weights = Tensor(w, Array(w.length))
-      gradients = Tensor(Array.tabulate(w.length)(i => 0f), Array(w.length))
+      if(this.isTraining()) {
```
jason-dai: It may not always be correct. In particular, the user can set the model to `training` or `evaluate` mode dynamically; so if it is set to `evaluate` first and later to `training`, you will get an empty `gradients` in backward. I think you may need to dynamically set the `gradients` in `accGradParameters`.

In addition, we also need to appropriately set the `training` or `evaluate` mode in the underlying PyTorck model when user calls the corresponding method on `TorchNet`, as some layer (e.g., BatchNorm) may have different behaviors in different modes.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: Right now the code does not save updated weights back to the torch model file (bytes). I just ensure the weights in scala TorchNet are up to date and can be serialized during broadcast.

pyzoo/zoo/examples/pytorch/train/resnet_finetune.py
hhbyyh: Right now I think the weights in resnet are updated also.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchNet.scala
hhbyyh: I will try to change that. 

I'm not quite sure about the batch norm behavior in the tracing model. I listed it as a next step in https://github.com/intel-analytics/analytics-zoo/issues/1528

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -63,33 +66,35 @@ auto getWeights(std::shared_ptr<torch::jit::script::Module> m, std::vector<float
...
-                auto p = static_cast<float*>(x.value().toTensor().grad().storage().data());
+            auto& grad = x.value().toTensor().grad();
+            if (grad.defined()) {
+                size_t x_size = grad.numel();
+                auto p = static_cast<float*>(grad.storage().data());
                 for(size_t i = 0; i < x_size; i++)
                 {
                     xv.push_back(p[i]);
                 }
+                grad = grad.detach();
```
jason-dai: Does it make sense to zero the grads unconditionally here? Maybe we should zero the grads in `TrochNet.zeroGradParameters`

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -231,25 +259,98 @@ JNIEXPORT jobject JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchM
         torch_shape.push_back(*(c_shape + i));
         i++;
     }
-    // create a Tensor
-    auto label_tensor = torch::from_blob(c_storage + joffset, torch_shape, at::kFloat);
+    // create gradOutput Tensor
+    auto gradOutput_tensor = torch::from_blob(c_storage + joffset, torch_shape, at::kFloat);
+
+    at::Tensor y = modelOutputs[nativeRef];
+    y.backward(gradOutput_tensor);
```
jason-dai: do we need to call `y.backward` here? I think calling `loss.backward` (in `lossForwardNative`) is sufficient

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -162,18 +180,26 @@ JNIEXPORT jobject JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchM
     }
     // create a Tensor
     auto torch_input_tensor = torch::from_blob(c_storage + joffset, torch_shape, at::kFloat);
```
jason-dai: better to refactor the code to make converting between torch tensor and JTensor two methods (instead of duplicating the code in every function)

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Loss was calculated based on Loss input only, and is not related to model output any more. Model output was sent to Loss by BigDL. (Tensor -> JTensor -> Tensor)

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: I assumed we can discard gradients after collecting them. I haven't found a condition that this will cause problem. 
zeroGradParameters would iterate through all the layers again in a similar pattern.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Yes, and we need to avoid extra memory during function invoke and avoid unnecessary parameter copy etc.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
jason-dai: I meant `loss.backward` will automatically back propagate through the model (i.e., 'y.backward'), and there is no need to call 'y.backward` explicitly again?

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
jason-dai: So one cannot get parameters twice? Maybe we can use the Scala weight/gradients as a cache for these? We can support it in a follow-up PR.

hhbyyh: We have cache for gradients in scala side. Sure we can add the logic if required.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: loss.backward will end at the loss input, and will not back propagate further to the model now, since the loss input is explicitly feeded from BigDL.

hhbyyh:
validation passed. http://172.168.2.101:8080/view/ZOO-PR-PIPELINE/job/ZOO-PR-Validation/2483/

jason-dai:
LGTM

hhbyyh:
Thanks for the review. Merging to master.

PR #1509, opened 2019-07-11
Support setting core number from python side

yangw1234:
Setting core number in Engine can control the number of model replica in each node.

This is a temporary support and more user friendly api will be added after pr #1488 

Example Usage:
Say you running on a spark cluster with executor core number is 28, and you want to run two tensorflow model on each executor and each of the model using 14 threads.

Then you can you can do the following to achieve this:
1. export OMP_NUM_THREADS=14 before submitting spark job
2. set core number to change the value in engine and make DistriOptimizer only copy 2 model in each executor.

    ```python
    from zoo.common import set_core_number
    set_core_number(core_num)
    ```
3. set the session_config parameter in TFOptimizer to make each model using 14 threads
    ```
    TFOptimizer.from_loss(..., session_config=tf.ConfigProto(inter_op_parallelism_threads=1, intra_op_parallelism_threads=14))
    ```

### DIFF
--- a/pyzoo/test/zoo/common/test_util.py
+++ b/pyzoo/test/zoo/common/test_util.py
@@ -1,0 +1,43 @@
+#
+# Copyright 2018 Analytics Zoo Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pytest
+
+from bigdl.util.common import get_node_and_core_number
+from test.zoo.pipeline.utils.test_utils import ZooTestCase
+from zoo.common import set_core_num
+
+
+class TestUtil(ZooTestCase):
+
+    def test_set_core_num(self):
+        _, core_num = get_node_and_core_number()
+
+        set_core_num(core_num + 1)
+
+        _, new_core_num = get_node_and_core_number()
+
+        if new_core_num == core_num + 1:
+            # set it back
+            set_core_num(core_num)
+        else:
+            assert new_core_num == core_num + 1,\
+                "set_core_num failed, set the core" \
+                " number to be {} but got {}".format(core_num + 1, new_core_num)
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


pyzoo/test/zoo/common/test_util.py
```diff
@@ -0,0 +1,43 @@
...
+
+        set_core_num(core_num + 1)
+
+        _, new_core_num = get_node_and_core_number()
+
+        if new_core_num == core_num + 1:
+            # set it back
+            set_core_num(core_num)
+        else:
+            assert new_core_num == core_num + 1,\
```
hkvision: This assert can be changed to directly raise an Exception?

pyzoo/zoo/common/utils.py
```diff
@@ -38,6 +38,10 @@ def to_list_of_numpy(elements):
     return results
 
 
+def set_core_num(num):
```
hkvision: num -> number to keep align with get node and core number?

pyzoo/test/zoo/common/test_util.py
yangw1234: yeah this if ... else makes no sense if change that

pyzoo/zoo/common/utils.py
yangw1234: sure

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, val_method, output_length, target_length):
 class TFTrainingHelper(Layer):
     def __init__(self, path, configProto):
         if configProto is not None:
-            byte_arr = configProto.SerializeToString()
+            byte_arr = bytearray(configProto.SerializeToString())
```
jenniew: why do this?

jenniew:
The usage should be:
from zoo.common import set_core_number
 set_core_num(core_num)
Not : from bigdl.util.common import get_node_and_core_number
    set_core_num(core_num)

hkvision:
`set_core_num` is the total core number right? What if the user set core num not divisible by executor num? Will it be able to allocate different number of cores for different executors?

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
yangw1234: changes that make it work on python 2

yangw1234:
> The usage should be:
> from zoo.common import set_core_number
> set_core_num(core_num)
> Not : from bigdl.util.common import get_node_and_core_number
> set_core_num(core_num)

@jenniew Yeah, just fixed

yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2437/parameters/

PR #1515, opened 2019-07-15
Update docs relating to TFOptimizer and TFPredictor

Bob-Chou:


hkvision:
What about https://analytics-zoo.github.io/master/?

Bob-Chou:
Modified it :)

PR #1525, opened 2019-07-19
Caffeloader

Litchilitchy:
migrate caffeloader

merge batchnorm and scale


Litchilitchy:
will add unit test soon

qiuxin2012:
Please add a unit test to make sure your model will generate the same output and gradient compared with the model loaded by BigDL.

Litchilitchy:
**this pr is in progress.**

currently, to implement merging all BN and SCALE layers in caffe model, we need to do followings:

1. find consecutive BN and SCALE, replace BN from `affine=false` to `affine=true`, this control that if this BN layer could have weight and bias, then keep parameter of SCALE and drop SCALE layer

2. change nodes information, e.g. `Node.prevs` and `Node.nexts` to corresponding to the updated graph

3. copy parameters of SCALE (in step 1) to BN

**This fix is for support of MKLDNN on caffe models (at least it could support all Resnet models)**

**However, if SCALE is not the only layer after BN, these 2 layers still could not be merged. Besides, single SCALE layer without BN in front of it could still not supported by MKLDNN**

**Any mistakes in my statements or any advice kindly let me know.** @jason-dai @qiuxin2012

Litchilitchy:
fixed

wzhongyuan:
Why not submitting an enhanced PR to BigDL?

zoo/src/main/scala/com/intel/analytics/zoo/models/common/ZooModel.scala
```diff
@@ -28,6 +28,7 @@ import org.apache.spark.rdd.RDD
 
 import scala.reflect.ClassTag
 
+
```
jason-dai: why this change?

PR #1543, opened 2019-07-31
update to bigdl 0.9.0

qiuxin2012:
As bigdl has fixed their bug by https://github.com/intel-analytics/BigDL/pull/2816 in 0.9.0, so we should remove our workaround in this update.


glorysdj:
how about another issue, we have to add analytics-zoo-core dependency explicitly in app pom since last maven pom change for analytics-zoo-core?
Any fix?

Issue #1528, opened 2019-07-22
Pytorch next steps


hhbyyh:
1. Support Table/multiple input and output models.

2. JNI memory optimization. Avoid mem leak and unnecessary copy.

3. Scalability test for training and inference

3. Support export to Pytorch model.

5. Utility to convert data loader to RDD/DataFrame

6. RNN/LSTM training.

hhbyyh:
8. Support transfer learning with Pytorch model

hhbyyh:
9. Investigate the performance impact of short cut the backward process from loss to model.

hhbyyh:
10. Docs
11. Better exception handling and error messages.

PR #1544, opened 2019-08-01
Pytorch training and inference for model with multiple output and multiple input

hhbyyh:


hhbyyh:
The code is ready for review. I'll do more tests with larger model.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -249,99 +279,195 @@ JNIEXPORT void JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchMode
...
+    if (y.isTuple()) {
+        auto yTuple = y.toTuple();
+        assert (input_size == yTuple -> elements().size());
+        for (int i = 0; i < input_size; i++) {
+            auto gradTensor = gradOutput_table -> elements()[i].toTensor();
+            auto outputTensor = yTuple -> elements()[i].toTensor();
+            outputTensor.backward(gradTensor);
+        }
+    } else {
+        y.toTensor().backward(gradOutput_table -> elements()[0].toTensor());
```
jason-dai: looking at the pythorch official tutorials (e.g., https://pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html#train-the-network and https://pytorch.org/tutorials/beginner/pytorch_with_examples.html#pytorch-custom-nn-modules), you just need call `loss.backward`, and there is no need to call 'output.backward`?

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
```diff
@@ -193,50 +190,83 @@ JNIEXPORT jlong JNICALL Java_com_intel_analytics_zoo_pipeline_api_net_PytorchMod
...
 
-    // Create a vector of inputs.
-    std::vector<torch::jit::IValue> inputs;
-    inputs.push_back(torch_input_tensor);
+        input_vector.push_back(input_tensor);
 
+        j_data_vector.push_back(tensor_storage);
+        c_data_vector.push_back(input_c_storage);
+        j_shape_vector.push_back(tensor_shape);
+        c_shape_vector.push_back(input_c_shape);
```
jason-dai: maybe also creating a method for jtensor2Torch?

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -38,16 +38,23 @@ def __init__(self, path, bigdl_type="float"):
         super(TorchNet, self).__init__(None, bigdl_type, path)
 
     @staticmethod
-    def from_pytorch(module, input_shape):
+    def from_pytorch(module, input_shape=None, sample_input=None):
```
jason-dai: we should support the case when `input_shape` is an array of vectors (for table input); not sure if we want `sample_input`, as we can always generate that using `input_shape`

pyzoo/zoo/pipeline/api/net/torch_criterion.py
```diff
@@ -44,9 +44,13 @@ def __init__(self, path, bigdl_type="float"):
         super(TorchCriterion, self).__init__(None, bigdl_type, path)
 
     @staticmethod
-    def from_pytorch(loss, input_shape, label_shape=None, sample_input=None, sample_label=None):
+    def from_pytorch(loss, input_shape=None, label_shape=None,
```
jason-dai: we should support the case when `input_shape` (or `label_shape`) is an array of vectors (for table input/lable); not sure if we want `sample_input` (or `sample_label`), as we can always generate that using `input_shape` (or `label_shape`)

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Right now the data flow in `TorchNet` and `TorchCriterion` are not connected. `TorchNet` is simply an AbstractModule, while `TorchCriterion` is just a AZ Criterion, and AZ handles the data exchange between them. So loss backward will only create gradient for loss input, but not for model output, and thus not for model input.

This gives us the flexibility to combine AZ layers with Torch Layers. Indeed there's a possible optimization step that we can take, i.e. if a torch criterion finds the input is from a torchnet, it can load cached output directly and ignore its input data, and send gradients all the way back to model input, and notify torchnet to ignore the backwards. But I think it's kinda hacky and performance improvement may not be significant.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: Yes. I tried that but have two concerns:
1. I want to avoid return value (a Tensor) copy. Right now the torch tensor is created by :
`auto input_tensor = torch::from_blob(input_c_storage + c_input_offsets[i], input_torch_shape, at::kFloat);` and it's not a pointer. I'm not sure if it's safe to return a reference to stack variable.

2. We need to invoke ReleasePrimitiveArrayCritical after model forward. So we need to keep track or return the handles to the arrays if using a separate function.

Overall, right now I'm having some performance concern when trying to refactor that part. It will be helpful if we can have some Cpp expert to look closely.

pyzoo/zoo/pipeline/api/net/torch_net.py
hhbyyh: Sure, I'll add multiple input shape support.
Right now we are generating random float tensors from the shape, yet some torch criterion, e.g. crossEntrophy requires label type to be int64. 
I think sample_input can help:
1. handle some situations that input shape cannot cover, e.g. data type, or other case that we haven't thought of.
2. user may use that to choose a route for data-based models.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/api/net/native/com_intel_analytics_zoo_pipeline_api_net_PytorchModel.cpp
hhbyyh: I added it in https://github.com/intel-analytics/analytics-zoo/issues/1528

hhbyyh:
Thanks for the review. @jason-dai Those are all very good points.

hhbyyh:
http://172.168.2.101:8080/job/ZOO-PR-Validation-Spark-2.4/864/ passed

hhbyyh:
The dead lock issue is resolved.

pyzoo/zoo/pipeline/api/net/torch_criterion.py
hhbyyh: Please refer to the previous conversation.

jason-dai:
LGTM

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR-PIPELINE/job/ZOO-PR-Validation/2519/ passed. Thanks for the review. Merging to master.

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -38,16 +43,23 @@ def __init__(self, path, bigdl_type="float"):
...
-        Create a TorchNet directly from PyTorch model, e.g. model in torchvision.models
+        Create a TorchNet directly from PyTorch model, e.g. model in torchvision.models.
+        Users need to specify sample_input or input_shape.
         :param module: a PyTorch model
-        :param input_shape: list of integers. E.g. for ResNet, this may be [1, 3, 224, 224]
+        :param input_shape: list of integers, or tuple of list for multiple inputs models. E.g.
+                            for ResNet, this may be [1, 3, 224, 224]
+        :param sample_input. A sample of Torch Tensor or tuple to trace the model.
         """
+        if not input_shape and not sample_input:
```
qiuxin2012: @hhbyyh find a bug here.
If sample_input is provided, will throw an exception like this:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-17-5a5cbb42ceda> in <module>
      1 import torch
      2 #zoo_rcnn = TorchNet.from_pytorch(coco_demo.model, [3, 800, 800], torch.rand(3, 800, 800))
----> 3 zoo_rcnn = TorchNet.from_pytorch(coco_demo.model, sample_input = torch.rand(3, 800, 800))

/opt/anaconda3/envs/py36/lib/python3.6/site-packages/zoo/pipeline/api/net/torch_net.py in from_pytorch(module, input_shape, sample_input)
     53         :param sample_input. A sample of Torch Tensor or tuple to trace the model.
     54         """
---> 55         if not input_shape and not sample_input:
     56             raise Exception("please specify input_shape or sample_input")
     57 

RuntimeError: bool value of Tensor with more than one value is ambiguous
```

Seems torch's tensor has another way to deal with `not`.

```
import torch
t = torch.rand(3, 800, 800)
not t
```
returns
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-19-1c135d71686f> in <module>
----> 1 not t

RuntimeError: bool value of Tensor with more than one value is ambiguous
```

hhbyyh: Got it. Thanks. Will fix it in another PR.

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -56,6 +68,17 @@ def from_pytorch(module, input_shape):
 
         return net
 
+    @staticmethod
+    def get_sample_input(shape, sample):
+        if sample:
```
qiuxin2012: sample problem here. @hhbyyh

PR #1545, opened 2019-08-01
OpenVINO Auto Calibration Tool

qiyuangong:


jason-dai:
- This should be included in pyzoo/zoo/serving/openvino @Litchilitchy 
- We should refactor the functionalities into appropriate pythod methods, and the python script can simply invoke these methods

qiyuangong:
@jason-dai We have removed calibration tool from zoo in 0.8 release. Do we still need this PR?

qiyuangong:
Not necessary any more. PR closed.

Issue #1219, opened 2019-03-15
NNFrame needs to support multiple input columns for training and inference


jenniew:
NNFrame need to support multiple inputs for training and inference.

PR #1553, opened 2019-08-05
support multi input models for nnframes

hhbyyh:
This supports multi-input model for NNFrames API. i.e. user can create

`NNEstimator(zmodel, criterion, [[2, 3],[4]], [1])`, in which the model expects two inputs of size[2, 3] and [4].


jason-dai:
How does the users prepare the input Dataframe? Do they have to combine multiple tensors into one field?

hhbyyh:
User shouldn't be dealing with Tensors during feature engineering steps. Most likely they would use VectorAssembler in Spark ML to concat different fields. I imagine it should work well for cases like Recommendation. 

Btw, this is not in conflict with the multi-features columns solution if we also want to support that. All this PR added is essentially a simple Preprocessing step. And this solution is more intuitive for ML pipeline.

Supporting multiple feature/label columns would change existing NNEstimator by a lot, and I suggest to create a new specific class for that case.

jason-dai:
Please add unit test and document to show how to use `VectorAssembler` to prepare input

hhbyyh:
Sure. Added.

docs/docs/APIGuide/PipelineAPI/nnframes.md
```diff
@@ -50,6 +50,14 @@ More concrete examples are available in package `com.intel.analytics.zoo.example
    type are supported) and convert each feature/label to Tensor according to the specified Tensor
    size.
 
+**2.** `NNEstimator(model, criterion, featureSize: Array[]Array[Int]], labelSize: Array[Int])`
```
jason-dai: eatureSize: Array[[Array[Int]]

hhbyyh: Updated. Thanks.

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR-PIPELINE/job/ZOO-PR-Validation/2549/ passed.

hhbyyh:
cc @jason-dai If there's no further concern, I'll merge this. Thanks

jason-dai:
LGTM

hhbyyh:
http://172.168.2.101:8080/view/ZOO-PR-PIPELINE/job/ZOO-PR-Validation/2552/ passed. Merging to master. Thanks for the review.

sansr:
@jenniew the issue that you reported in #1219 it is solved with this?

jenniew:
@sansr Yes, it solves issue https://github.com/intel-analytics/analytics-zoo/issues/1219

docs/docs/APIGuide/PipelineAPI/nnframes.md
```diff
@@ -109,6 +117,57 @@ estimator = NNEstimator(model, criterion, SeqToTensor([2]), ArrayToTensor([2]))\
...
+flatten = ZLayer.Flatten()(user_embedding)
+dense1 = ZLayer.Dense(2)(x2)
+gru = ZLayer.LSTM(4, input_shape=(2, 2))(x3)
+
+merged = ZLayer.merge([flatten, dense1, gru], mode="concat")
+zy = ZLayer.Dense(2)(merged)
+
+zmodel = ZModel([x1, x2, x3], zy)
+criterion = ClassNLLCriterion()
+classifier = NNClassifier(zmodel, criterion, [[1], [2], [2, 2]]) \
```
zhichao-li: Maybe this can be a nice syntax sugar as `VectorAssembler` is just an implementing detail, and somehow can be transparent. i.e we can transform the feature cols to `VectorAssembler` internally once user invoke:
```
estimator.setFeatureCols(["user", "age", "income", "history"], shape=xxx)
```

hhbyyh: Thanks @zhichao-li. I like the idea, and IMO we need to ensure the number of features columns matches the number of shapes, e.g. a one-to-one mapping between feature column and shape, this way we don't even need a `VectorAssembler` internally.

This PR supports the multi-input models for simple array data. To support multiple feature columns, ideally we need to support multiple preprocessing steps. E.g. for siamese or even triplet image models, we need to support different preprocessing step for different feature columns and model inputs.

PR #1557, opened 2019-08-08
Support multiple optimMethods in TFTrainingHelper

hkvision:
- Train a TF unfreeze graph using multiple optim methods based on scope
- Bug fix for Python Estimator.
- Unit tests.


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper.scala
```diff
@@ -57,6 +57,27 @@ private[zoo] class TFTrainingHelper(tfnet: TFNet,
...
+    val map = collection.mutable.Map[String, Tensor[Float]]()
+    var i = 0
+    while (i < variables.length) {
+      map(variables(i)) = weights(i)
+      i +=1
+    }
+    map
+  }
+
+  override def apply(name: String): Option[AbstractModule[Activity, Activity, Float]] = {
```
yangw1234: Should we make sure that each `TFSubGraph`'s weights' underlying storage is a continuous subset of the `TFTrainingHelper`'s weights' storage.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper.scala
```diff
@@ -57,6 +57,27 @@ private[zoo] class TFTrainingHelper(tfnet: TFNet,
...
+    val targetVariables = if (name == getName()) variables else variables.filter(_.contains(name))
```
yangw1234: Should we right `val targetVariables = variables.filter(_.startsWith(name))`.

Say we have variables namely "a/b/c", "c/a/b" and "a". The sub-graph "a" should contains "a" and "a/b/c".

hkvision: Sure.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper.scala
hkvision: https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala#L504
It is checked in Optimizer? But anyway, it will be safer to check it here.

hkvision: @yangw1234 Seems that the most secure way to do the check is via `checkSubModules` in Optimizer, if I write by myself, does the following covers all cases?
```
val targetStorage = targetWeights.flatMap(x => x.storage().array().slice(x.storageOffset(), x.storageOffset() + x.nElement()))
val totalStorage = Module.flatten(parameters()._1).storage().array()
require(totalStorage.containsSlice(targetStorage))
```
Take a look, thanks.

hkvision:
Can move the ut to Python if necessary. Python API doesn't change.

hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/2530/

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper.scala
```diff
@@ -144,6 +162,21 @@ private[zoo] class TFTrainingHelper(tfnet: TFNet,
...
+  override def updateOutput(input: Activity): Activity = {
+    input
+  }
+
+  override def updateGradInput(input: Activity, gradOutput: Activity): Activity = {
+    gradInput
+  }
+
+  override def parameters(): (Array[Tensor[Float]], Array[Tensor[Float]]) = {
+    (weights, weights.map(_ => Tensor[Float]()))
```
qiuxin2012: grad's length should be the same with weights

hkvision:
Need to sort variable names otherwise weights under the same scope are not contiguous.

hkvision:
Sorted variables by name @yangw1234

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper.scala
```diff
@@ -58,6 +56,27 @@ private[zoo] class TFTrainingHelper(tfnet: TFNet,
     setWeights(ws)
   }
 
+  private val weightsMap = {
```
yangw1234: You should change TFTrainingHelper2 which is the not-freeze implementation

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
```diff
@@ -562,6 +563,7 @@ def from_keras(cls, keras_model, dataset, optim_method=None, val_spilt=0.0, **kw
         inputs = model_inputs + model_targets
 
         variables = keras_model._collected_trainable_weights
+        variables.sort(key=lambda variable: variable.name)
```
yangw1234: do this in the constructor? If we add other factory methods to create TFOptimizer, we might forget to sort the variables.

zoo/src/test/scala/com/intel/analytics/zoo/pipeline/api/net/TFNetSpec.scala
```diff
@@ -160,4 +164,36 @@ class TFNetSpec extends FlatSpec with Matchers with BeforeAndAfter {
...
+      i += 1
+    }
+
+    val rdd = sc.parallelize(data)
+    val dataSet = DataSet.rdd(rdd) -> SampleToMiniBatch[Float](128)
+    val optimizer = new DistriOptimizer[Float](
+      layer,
+      dataSet.asInstanceOf[DistributedDataSet[MiniBatch[Float]]],
+      new IdentityCriterion())
+      .setOptimMethods(
```
yangw1234: I think we should enable user to set it TFOptimizer, KerasModel and TFEstimator.

hkvision: Supported in TFOptimizer and TFEstimator. Pending for KerasModel.

pyzoo/zoo/pipeline/api/net/tf_optimizer.py
hkvision: Added remarks. There is little possibility that we have other factory methods.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper.scala
hkvision: Changed.

pyzoo/test/zoo/tfpark/test_tfpark_tfoptimizer.py
```diff
@@ -66,15 +65,22 @@ def test_tf_optimizer_metrics(self):
...
                                               val_outputs=[output],
                                               val_labels=[label_tensor],
                                               val_method=Accuracy(), metrics={"loss": loss})
+            initial_weights = optimizer.tf_model.training_helper_layer.get_weights()
             optimizer.optimize(end_trigger=MaxEpoch(1))
+            updated_weights = optimizer.tf_model.training_helper_layer.get_weights()
+            for i in [0, 1]:  # weights and bias combined with "dense/" should be updated
+                assert not np.allclose(initial_weights[i], updated_weights[i])
+            for i in [2, 3]:  # weights and bias combined with "dense_1" should be unchanged
+                assert np.allclose(initial_weights[i], updated_weights[i])
```
hkvision: Added verification of weights @yangw1234

hkvision:
Please take a look if there are further comments. @yangw1234

hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2853/

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper2.scala
```diff
@@ -304,3 +327,17 @@ case class TrainMeta2(inputs: Array[String],
...
+  override def updateOutput(input: Activity): Activity = {
+    input
+  }
+
+  override def updateGradInput(input: Activity, gradOutput: Activity): Activity = {
+    gradInput
+  }
+
+  override def parameters(): (Array[Tensor[Float]], Array[Tensor[Float]]) = {
+    (weights, weights.map(_ => Tensor[Float]()))
```
yangw1234: Nitpick: One might wonder why the gradients of `TFSubGraph` does not have to be the same instances of the original `TFTrainingHelper`'s. The truth is they are never used in DistriOptimizer. And the DistriOptimizer only use the weights' size information to slice on the original weights and gradients. It is better leave some comments about that. Also, this way, if the implementation of DistriOptimizer changed, may lead to incorrect result.

yangw1234:
LGTM

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TFTrainingHelper2.scala
hkvision: Added detailed comments. If the implementation of DistriOptimizer is changed, then I suppose many many code segments would be affected :)

hkvision:
Thanks for your review @yangw1234

hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/2855/

PR #1559, opened 2019-08-09
Session recommender document

songhappy:


lordk911:
The website ( www.officedepot.com ) which could offer sample data for Session recommender can't be access from some country, so could you offer the data schema or some litte sample data in the document.

PR #1567, opened 2019-08-16
Resnet50InferenceModel example

Le-Zheng:
add Resnet50InferenceModel.scala, ImageClassificationStreaming.scala
add resources file


docker/jenkins-ci/jenkins-slave/Dockerfile
```diff
@@ -4,23 +4,26 @@ MAINTAINER The Analytics-Zoo Authors https://github.com/intel-analytics/analytic
```
glorysdj: we don't need thius

PR #1574, opened 2019-08-19
Option to add optim_method when creating TFOptimizer from_keras

hkvision:
Gives more flexibility if the mapping from keras optimizer to bigdl optimizer is not totally complete.

For example, learning rate scheduler can't be handled by `TFOptimizer.to_bigdl_optim_method(optim_method)`, and in this case, alternatively, users can directly pass in a BigDL optim_method as a workaround.


hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2556/

PR #1575, opened 2019-08-19
Add Polynomial Decay based on epochs

hkvision:
Exactly the same logic as the scheduler used in Glodwind case.


hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2557/
Thanks for your review @qiuxin2012

PR #1580, opened 2019-08-21
Simple TFNet example 

hhbyyh:
The simple example for training a TF model (without loss) with TFNet and NNFrames.


hhbyyh:
cc @yangw1234 Can you please help take a look?

yangw1234:
LGTM

PR #1600, opened 2019-09-05
Save Keras-like model to pure keras or tensorflow protobuf.

qiuxin2012:
Save keras-like model to pure keras or tensorflow protobuf.
Python, analytics-zoo(include bigdl and pyspark), tensorflow should be installed.

PR #1606, opened 2019-09-09
Save KerasModel to pure keras or tf protobuf

qiuxin2012:
Add support to Keras Model.
Related pr: #1600


qiuxin2012:
jenkins passed

PR #1608, opened 2019-09-10
Fix confusing scaler variable name

mokahaiku:
The notebook uses StandardScaler and stores it under variable name min_max_scaler, which is confusing.


jason-dai:
Thanks for the fix

PR #1610, opened 2019-09-11
Fix imageset read relative path

yangw1234:
Fix imageset read relative path


yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2612/

PR #1619, opened 2019-09-16
Fix incorrect code comment

wendyshusband:
There is an obvious error in the comments.
The variate meanB is mean of channel B rather than  mean of channel R


hkvision:
Thanks for correcting the typo!

PR #1655, opened 2019-10-09
Enhance TextSet Python API

hkvision:
- Support memory type
- Support constructing by indices directly
requested by @jenniew


pyzoo/zoo/feature/text/text_feature.py
```diff
@@ -109,3 +113,6 @@ def keys(self):
         :return: List of String
         """
         return callBigDlFunc(self.bigdl_type, "textFeatureGetKeys", self.value)
+
+    def __reduce__(self):
+        return TextFeature, (self.value, self.bigdl_type)
```
hkvision: Currently Python TextFeature is not serializable, i.e. we cannot construct RDD of TextFeature, any idea? @jenniew

PR #1672, opened 2019-10-11
[Automl] Change default look back value and evaluation output format while future_seq_len=1

yushan111:
main changes:
1. add result analysis in ray 0.7.4 (searcher.run() returns ExperimentAnalysis instead of trials)
2. change default look back length from 1 to 2. Add input check and some corresponding UTs
3. while future_seq_len=1, output one value instead of an array

### DIFF
--- a/pyzoo/test/zoo/automl/feature/test_time_sequence_feature.py
+++ b/pyzoo/test/zoo/automl/feature/test_time_sequence_feature.py
@@ -154,15 +154,19 @@ class TestTimeSequenceFeature(ZooTestCase):
         feat = TimeSequenceFeatureTransformer(future_seq_len=1, dt_col="datetime",
                                               target_col="values", drop_missing=True)
         with pytest.raises(ValueError, match=r".*past sequence length.*"):
-            feat.fit_transform(train_df[:10], **config)
+            feat.fit_transform(train_df[:20], **config)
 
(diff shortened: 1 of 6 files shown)


shane-huang:
Does it break if we use ray 0.7.2?  ray 0.7.4 has some issues I think.

jason-dai:
I think we should use Ray 0.6.6 (https://analytics-zoo.github.io/master/#ProgrammingGuide/rayonspark/) according to @zhichao-li ?

zhichao-li:
Previously, we test against 0.6.6 and it works smoothly, but the latest version would have some known issue. Take a look at these issues: https://github.com/ray-project/ray/issues/5223 and #5454

richardliaw:
@yushan111 @shane-huang @jason-dai @zhichao-li 

I'm from the ray team and maintain tune - have you all joined the ray slack? let's setup a channel to chat!

yushan111:
> main changes:
> 
> 1. add result analysis in ray 0.7.4 (searcher.run() returns ExperimentAnalysis instead of trials)
> 2. change default look back length from 1 to 2. Add input check and some corresponding UTs
> 3. while future_seq_len=1, output one value instead of an array

Code related to ray 0.7.4 is removed. 
I think it works well with ray version<= 0.7.0 .

PR #1677, opened 2019-10-11
Support TorchNet long input for Embedding

hkvision:
TODO: multiple inputs with different dtypes.
TODO: update code doc

### DIFF
--- a/pyzoo/zoo/pipeline/api/net/torch_net.py
+++ b/pyzoo/zoo/pipeline/api/net/torch_net.py
@@ -43,7 +43,7 @@ class TorchNet(Layer):
         super(TorchNet, self).__init__(None, bigdl_type, path)
 
     @staticmethod
-    def from_pytorch(module, input, check_trace=True):
+    def from_pytorch(module, input, dtype=torch.float, check_trace=True):
         """
         Create a TorchNet directly from PyTorch model, e.g. model in torchvision.models.
         Users need to provide an example input or the input tensor shape.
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, path, bigdl_type="float"):
         super(TorchNet, self).__init__(None, bigdl_type, path)
 
     @staticmethod
-    def from_pytorch(module, input, check_trace=True):
+    def from_pytorch(module, input, dtype=torch.float, check_trace=True):
```
hhbyyh: Input parameter can accept a sample Tensor. In this case, maybe we can try to send a long tensor. Does that work for Embedding?

hkvision: Yeah we can send a long tensor. But as discussed with @jason-dai it will be more user-friendly if for users to pass a shape only?

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -131,3 +144,10 @@ def predict(self, x, batch_per_thread=1, distributed=True):
                 return [Layer.convert_output(result) for result in results]
             else:
                 raise TypeError("Unsupported prediction data type: %s" % type(x))
+
+
+def generate_torch_input(size, dtype):
+    if dtype == torch.long:
+        return torch.LongTensor(size).random_(0, 1)
```
qiuxin2012: All zero...
See https://pytorch.org/docs/stable/tensors.html#torch.Tensor.random_

pyzoo/zoo/pipeline/api/net/torch_net.py
```diff
@@ -63,11 +63,24 @@ def from_pytorch(module, input, check_trace=True):
         if input is None:
             raise Exception("please provide an example input or input Tensor size")
 
-        sample = TorchNet.get_sample_input(input)
+        sample = TorchNet.get_sample_input(input, dtype)
         temp = tempfile.mkdtemp()
 
+        from torch import nn
+
+        class OuterModule(nn.Module):
```
qiuxin2012: I think this should be named LongWrapperModule

hkvision: This can be potentially used to cast to other types not only long. In forward, we can add more cases in the future if we encounter data types that can't be directly handled.

pyzoo/zoo/pipeline/api/net/torch_net.py
hkvision: I intentionally do this because we don't know what's the embedding size. This will be used for forward for any model with embedding as input. Thus the safest way to generate a random input is all zeros so that for all embedding no error will be raised even if the embedding dimension is only 1.

PR #1713, opened 2019-10-18
Reduce vino logs

qiyuangong:
* Replace OpenVINO related decompression output with one line log info.
* Remove unnecessary `ls`.


qiyuangong:
Jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/2684/

PR #1720, opened 2019-10-21
Add init_spark_on_k8s

hkvision:
TODO:
- Add and update document.
- More code refactor.

### DIFF
--- a/pyzoo/zoo/ray/util/raycontext.py
+++ b/pyzoo/zoo/ray/util/raycontext.py
@@ -45,7 +45,7 @@ class JVMGuard():
                           pids)
         except Exception as err:
             print(traceback.format_exc())
-            print("Cannot sucessfully register pid into JVMGuard")
+            print("Cannot successfully register pid into JVMGuard")
             for pid in pids:
                 os.kill(pid, signal.SIGKILL)
             raise err
@@ -85,14 +85,14 @@ class RayServiceFuncGenerator(object):
             print("MKL cores is {}".format(cores))
             modified_env.update(self._get_MKL_config(cores))
         if self.verbose:
-            print("Executing with these environment setting:")
+            print("Executing with these environment settings:")
             for pair in modified_env.items():
                 print(pair)
             print("The $PATH is: {}".format(modified_env["PATH"]))
         return modified_env
 
     def __init__(self, python_loc, redis_port, ray_node_cpu_cores, mkl_cores,
-                 password, object_store_memory, waitting_time_sec=6, verbose=False, env=None,
+                 password, object_store_memory, waiting_time_sec=6, verbose=False, env=None,
                  extra_params=None):
         """object_store_memory: integer in bytes"""
         self.env = env
@@ -103,7 +103,7 @@ class RayServiceFuncGenerator(object):
         self.mkl_cores = mkl_cores
         self.ray_exec = self._get_ray_exec()
         self.object_store_memory = object_store_memory
-        self.waiting_time_sec = waitting_time_sec
+        self.waiting_time_sec = waiting_time_sec
         self.extra_params = extra_params
         self.verbose = verbose
         self.labels = """--resources='{"trainer": %s, "ps": %s }' """ % (1, 1)
@@ -241,10 +241,10 @@ class RayContext(object):
             ray_node_cpu_cores=self.ray_node_cpu_cores,
             mkl_cores=self._get_mkl_cores(),
             password=password,
-            object_store_memory=self._enrich_object_sotre_memory(sc, object_store_memory),
+            object_store_memory=self._enrich_object_store_memory(sc, object_store_memory),
             verbose=verbose,
             env=env,
-            waitting_time_sec=waiting_time_sec,
+            waiting_time_sec=waiting_time_sec,
             extra_params=extra_params)
         self._gather_cluster_ips()
         from bigdl.util.common import init_executor_gateway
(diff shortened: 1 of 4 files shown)


pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -69,8 +69,8 @@ def _prepare_env(self, cores=None):
         if self.env:
             modified_env.update(self.env)
         cwd = os.getcwd()
-        modified_env["PATH"] = "{}/{}:{}".format(cwd, "/".join(self.python_loc.split("/")[:-1]),
```
hkvision: Actually is this necessary? At least for local mode, cwd can be any location...

pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -163,7 +164,7 @@ def _start_ray_node(self, command, tag, wait_before=5, wait_after=5):
 
     def _get_ray_exec(self):
         python_bin_dir = "/".join(self.python_loc.split("/")[:-1])
-        return "{}/python {}/ray".format(python_bin_dir, python_bin_dir)
```
hkvision: Is this necessary? Can we directly start with `ray`? Seems in the docker image python and ray may not under the same location...

pyzoo/zoo/util/spark.py
```diff
@@ -216,3 +215,64 @@ def _submit_opt():
...
+
+        submit_args, conf = _submit_opt()
+        print(submit_args)
+        print(conf)
+
+        if not spark_conf:
+            spark_conf = {}
+
+        # if "spark.executor.extraClassPath" in spark_conf:
+        #     spark_conf["spark.executor.extraClassPath"] = "{}:{}".format(
```
hkvision: If driver and executor python envs are different, given `PYSPARK_PYTHON`, it may not be successful to find the jar under pip install. Thus adding driver jars via `--jars`.

pyzoo/zoo/util/spark.py
```diff
@@ -216,3 +215,64 @@ def _submit_opt():
...
+            return command
+
+        def _submit_opt():
+            conf = {
+                "spark.driver.memory": driver_memory,
+                "spark.driver.cores": driver_cores,
+                "spark.scheduler.minRegisteredResourcesRatio": "1.0",
+                "spark.executor.instances": num_executors,
+                "spark.kubernetes.container.image": container_image,
+                "spark.cores.max": num_executors * executor_cores}
```
hkvision: Seems we must provide this in conf, otherwise will raise an error... wired...

pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -162,8 +164,11 @@ def _start_ray_node(self, command, tag, wait_before=5, wait_after=5):
         return process_info
 
     def _get_ray_exec(self):
-        python_bin_dir = "/".join(self.python_loc.split("/")[:-1])
-        return "{}/python {}/ray".format(python_bin_dir, python_bin_dir)
+        if self.python_loc:
+            python_bin_dir = "/".join(self.python_loc.split("/")[:-1])
+            return "{}/python {}/ray".format(python_bin_dir, python_bin_dir)
```
hkvision: This should be specific for yarn python env from conda pack. For local or k8s, can directly start with `ray`.

hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2707/

pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -68,9 +69,10 @@ def _prepare_env(self, cores=None):
         modified_env = os.environ.copy()
         if self.env:
             modified_env.update(self.env)
-        cwd = os.getcwd()
-        modified_env["PATH"] = "{}/{}:{}".format(cwd, "/".join(self.python_loc.split("/")[:-1]),
-                                                 os.environ["PATH"])
+        if self.python_loc:
```
zhichao-li: why adding the checking here? same behavior for the yarn environment?

pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -155,15 +157,18 @@ def _start_ray_node(self, command, tag, wait_before=5, wait_after=5):
...
         time.sleep(wait_after)
         return process_info
 
     def _get_ray_exec(self):
-        python_bin_dir = "/".join(self.python_loc.split("/")[:-1])
-        return "{}/python {}/ray".format(python_bin_dir, python_bin_dir)
+        if self.python_loc:
+            python_bin_dir = "/".join(self.python_loc.split("/")[:-1])
+            return "{}/python {}/ray".format(python_bin_dir, python_bin_dir)
+        else:
```
zhichao-li: have you done the regression test for the yarn environment after this change?

hkvision: Have ran a simple test on yarn and there's no problem. Actually this change doesn't affect yarn as self.python_loc is always not None.

pyzoo/zoo/ray/util/raycontext.py
hkvision: For k8s or local, there is no need to add to PATH? I suppose this is done for conda-pack env only.

pyzoo/zoo/ray/util/raycontext.py
zhichao-li: ok, maybe we should either throw exception for the 'None' branch or remove the checking entirely.

zhichao-li:
The changes looks good as long as the yarn regression test is covered. :)

pyzoo/zoo/ray/util/raycontext.py
hkvision: But for k8s or local, I set it to None, since we can directly use `ray` to start, not `{}/python {}/ray`?

hkvision:
cc @jason-dai @glorysdj

PR #1730, opened 2019-10-30
[Bug Fix] SavedModel loaded TFNet return zero grad

yangw1234:
1. Fix variable not initialized after clone bug
2. return zero grad instead of null


yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2704/

PR #1735, opened 2019-11-04
Minor fix for detecting sc is local

hkvision:
Reported by tencent when testing.
"local" in self.sc.master can be standalone `spark://localhost:7777`

Even though we are not supporting for standalone, better fix it in case.


hkvision:
http://10.239.47.211:18888/job/ZOO-PR-Validation/2710/ connection error for multiple times. Jenkins should pass.

PR #1740, opened 2019-11-05
Start script support adding option which is a string with spaces

yangw1234:
Support adding option which is a string with spaces, e.g. `--option "a b c"`.

Iterating $* will result
```
--option
a
b
c
```

Iterating $@ will result
```
--option
a
b
c
```
Iterating "$*" will result
```
--option a b c
```
Iterating ”$@" will result
```
--option
a b c
```


yangw1234:
tested it manually

PR #1750, opened 2019-11-10
[Bug Fix] Fix estimator NPE problem

yangw1234:
It wrong to use `Some.apply` to construct a `Option` if the value might be a `null`.

`val a = Some[SomeType](null)` will return a Some with value null. `a.get` will not throw an exception, but `a.map(_.someAction)` will throw a NPE.

The correct way is to use `Option.appy`


yangw1234:
@cyita

yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2732/

PR #1753, opened 2019-11-12
remove bigdl engine init when load model

glorysdj:
this pr is to remove unnecessary code for bigdl engine init before model loading


glorysdj:
http://10.239.47.211:18888/view/ZOO-PR/job/ZOO-PR-Validation/2745/

PR #1757, opened 2019-11-13
update bigdl version to 0.10.0

glorysdj:
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/1757)


hkvision:
Also need to update python dependency and related doc.

PR #1763, opened 2019-11-14
first version of code refactor for feature transformer

yushan111:
_transformer.py_ defines the basic class of all transformers
_pre_roll_transformer.py_,  _roll_transformer.py_, _post_roll_transformer.py_ are based on `Transformers` and define the basic classes of the three kind of transformers
_time_sequence_feature_transformer.py_ is the top file and defines the feature transformer. The old feature transformer file is _time_sequence.py_ in the feature repo


yushan111:
Note that the changes in time_sequence_feature_transformer.py is auto-folded in since there are too many diffs. You may need to load diff manually.

PR #1770, opened 2019-11-15
[OpenVINO] Model broadcasting support

qiyuangong:


qiyuangong:
@glorysdj @yangw1234

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/OpenVINOModel.scala
```diff
@@ -66,14 +116,99 @@ class OpenVINOModel(var executableNetworkReference: Long = -1,
...
+}
+
+object OpenVINOModel {
+
+  private val modelBytesRegistry = new RegistryMap[(Array[Byte], Array[Byte])]()
+
+  val logger = LoggerFactory.getLogger(getClass)
+
+  @transient
+  private lazy val inDriver = NetUtils.isDriver
```
glorysdj: why need inDriver here? maybe we should keep the OpenVINOModel in pojo style

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/OpenVINOModel.scala
```diff
@@ -16,16 +16,66 @@
...
+
+  @transient
+  private lazy val supportive: OpenVinoInferenceSupportive = {
+    OpenVINOModel.logger.info("Prepare OpenVinoInferenceSupportive")
+    OpenVinoInferenceSupportive.forceLoad()
+    new OpenVinoInferenceSupportive()
+  }
+
+  @transient
+  private lazy val executableNetworkReference: Long = {
```
glorysdj: this makes that the first prediction will contain time of loading model

glorysdj:
My biggest concern is that model will hold 2 copies of weights, one is in model, the other is in bytes holder, since the weights is really big in DL, may be not proper in standalone/web serving scenarios. Can we add a flag to support both, one is creating a model without bytes holder, one is creating a model with bytes holder?

qiyuangong:
> My biggest concern is that model will hold 2 copies of weights, one is in model, the other is in bytes holder, since the weights is really big in DL, may be not proper in standalone/web serving scenarios. Can we add a flag to support both, one is creating a model without bytes holder, one is creating a model with bytes holder?

@yangw1234

yangw1234:
@qiyuangong 
`
def copy(num: Int): Array(this)
`
I think the copy implementation may not be correct.
First it ignores the num and just return a one element array.
Second, why simply return `this` instead of creating a new model? Let openvino to serve multiple requests?

If this is done intentionally, I think 2 copies of weights might be ok since there is only one instance of the model.  

If the correct way is to create multiple instances of model in JVM, maybe we can share the model bytes holder.

@glorysdj 's solution also sounds good to me too.

PR #1776, opened 2019-11-19
inference-tensorboard separated

Litchilitchy:
separated tensorboard pr to here


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceSummary.scala
```diff
@@ -0,0 +1,48 @@
...
+
+package com.intel.analytics.zoo.pipeline.inference
+
+import com.intel.analytics.zoo.tensorboard.Summary
+import com.intel.analytics.zoo.tensorboard.FileWriter
+import com.intel.analytics.zoo.tensorboard.FileReader
+
+
+
+class InferenceSummary(
```
qiuxin2012: scala doc and indent

PR #1783, opened 2019-11-21
Upgrade pip to bigdl 0.10 and add ray to extras require

hkvision:
Tested manually for
`pip install analytics-zoo[ray]` which will be more convenient.


hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2762/

PR #1784, opened 2019-11-21
add MTNet and UTs

yushan111:
add a new MTNet smoke recipe in time_sequence_predictor (for temporary). Since `past_seq_len` has different restrictions on different models. In MTNet, `past_seq_len` depends on other fields (`n` and `T`) therefore cannot be randomly generated. In other models, `past_seq_len` does not have any dependencies. Plus, the `look_back` option in `RandomRecipe`, `GridRandomRecipe`, etc need to be specially designed for MTNet.


yushan111:
This PR is based on [PR of change the way of model selection](https://github.com/intel-analytics/analytics-zoo/pull/1781)

PR #1793, opened 2019-11-25
add callZooFunc and change all callBigDlFunc to callZooFunc

### LINKED ISSUE #1425: Can callBigDlFunc in Python api be optimized?
In `callBigDlFunc` we have 

            try:
                api = getattr(jinvoker, name)
                result = callJavaFunc(api, *args)
            except Exception as e:
                error = e
                if "does not exist" not in str(e):
                    raise e
            else:
                return result

However before raise error, first we try `callJavaFunc`, which has expensive data conversion `_py2java`, specifically,

    def callJavaFunc(func, *args):
        """ Call Java Function """
        gateway = _get_gateway()
        args = [_py2java(gateway, a) for a in args] // this is expensive
        result = func(*args)
        return _java2py(gateway, result)

 (`_get_gateway()` would get a list of api, and `callBigDlFunc` tries them one by one), the specific code is

    JavaCreator.set_creator_class([])
    JavaCreator.add_creator_class("com.intel.analytics.zoo.pipeline.nnframes.python.PythonNNFrames")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.feature.python.PythonImageFeature")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.pipeline.api.keras.python.PythonAutoGrad")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.models.python.PythonZooModel")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.pipeline.api.keras.python.PythonZooKeras2")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.feature.python.PythonTextFeature")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.feature.python.PythonFeatureSet")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.pipeline.api.net.python.PythonZooNet")
    JavaCreator.add_creator_class("com.intel.analytics.zoo.pipeline.inference.PythonInferenceModel")
    for clz in creator_classes:
        JavaCreator.add_creator_class(clz)

 If the error in `callBigDlFunc` is raised, the data conversion is in vain. Could we avoid the unnecessary time cost? @zhichao-li @jason-dai 

Could kindly help to check @qiyuangong 

PS: the above 2 pieces of code is BigDl code

qiuxin2012:
When finding the right java method, callBigDlFunc will convert the python arguments to java a lot of times in [this failover](https://github.com/intel-analytics/BigDL/blob/a11e02e09279401b23ed77a42765b1e38282d24c/pyspark/bigdl/util/common.py#L585). This will lead to a very bad performance when arguments is very large.
To work around this issue, add a new callZooFunc and move the arguments converting out of the for loop.  
Change all callBigDlFunc in zoo to callZooFunc.
fixed #1425

### DIFF
--- a/pyzoo/zoo/common/utils.py
+++ b/pyzoo/zoo/common/utils.py
@@ -18,7 +18,6 @@ import numpy as np
 
 
 def to_list_of_numpy(elements):
-
     if isinstance(elements, np.ndarray):
         return [elements]
     elif np.isscalar(elements):
--- a/pyzoo/zoo/common/utils.py
+++ b/pyzoo/zoo/common/utils.py
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from bigdl.util.common import Sample as BSample, JTensor as BJTensor, JavaCreator, _get_gateway, _java2py, _py2java
+from bigdl.util.common import Sample as BSample, JTensor as BJTensor,\
+    JavaCreator, _get_gateway, _java2py, _py2java
 import numpy as np
 
 
--- a/pyzoo/zoo/common/utils.py
+++ b/pyzoo/zoo/common/utils.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from bigdl.util.common import Sample as BSample, JTensor as BJTensor, callBigDlFunc, JavaCreator, _get_gateway, _java2py, _py2java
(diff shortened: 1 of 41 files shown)


jason-dai:
Why change estimator's callBigDlFunc only?

qiuxin2012:
jenkins passed

Litchilitchy:
^ ^

PR #1801, opened 2019-11-28
reuse Jtensor's float array

qiuxin2012:
reuse jtensor's float array.
To float tensor, `map` will clone a new float array, we can reuse the origin float array rather than clone a new one.


yangw1234:
LGTM

qiuxin2012:
jenkins passed

PR #1802, opened 2019-11-29
Rename and refactor ray related

hkvision:


hkvision:
http://10.239.47.211:18888/job/ZOO-PR-Validation/2821/
Two jobs fail due to git error.

PR #1810, opened 2019-12-04
use rng instead of Random in itemid's random and fix the distribution.

qiuxin2012:
use rng instead of Random in itemid's random.
Current random `val iid = Math.max(ran.nextInt(itemCount), 1)` is not a uniform distribution, one has `2/itemCount`'s possibility and, itemCount's possibility is 0%.


qiuxin2012:
jenkins passed

Issue #1764, opened 2019-11-15
[automl] add random walk model


shane-huang:
Random walk model (or naive model) are widely used for non-stationary data, and works remarkably well for many economic and financial time series. 

The forecasts from a random walk model are equal to the last observation, as future movements are unpredictable, and are equally likely to be up or down. 

Formula as :
`
y (t+h) = y (t)
`

shane-huang:
references : 

- https://otexts.com/fpp2/stationarity.html
- https://people.duke.edu/~rnau/Notes_on_the_random_walk_model--Robert_Nau.pdf

endymecy:
https://github.com/intel-analytics/analytics-zoo/pull/1835

PR #1835, opened 2019-12-11
add random walk model

endymecy:


endymecy:
this commit fixes #1764

PR #1837, opened 2019-12-11
Remove default mkl settings when start ray

### LINKED ISSUE #1831: Why mkl_cores for ray local is always set to 1?
https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/ray/util/raycontext.py#L307

This affects the performance running ray locally.

cc @jason-dai

hkvision:
Default mkl-related environment variables would hinder performance for different frameworks.
Removed all of them.
If users want to set some/all of these environments, they could choose to pass in the env argument.

Fix #1831

### DIFF
--- a/pyzoo/zoo/ray/util/raycontext.py
+++ b/pyzoo/zoo/ray/util/raycontext.py
@@ -55,35 +55,23 @@ class RayServiceFuncGenerator(object):
     """
     This should be a pickable class.
     """
-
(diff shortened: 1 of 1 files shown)


hkvision:
http://10.239.47.211:18888/job/ZOO-PR-Validation/2878/ Two jobs failed due to unrelated error.

PR #1838, opened 2019-12-11
Add MXNet support for RayOnSpark

hkvision:
- [x] MXNet Trainer and Runner
- [x] LeNet example
- [x] Resnet18_v1 on cifar10 example
- [x] Accuracy comparison with MXNet
- [x] Performance comparison with MXNet

### DIFF
--- a/pyzoo/zoo/examples/ray/mxnet/__init__.py
+++ b/pyzoo/zoo/examples/ray/mxnet/__init__.py
@@ -1,0 +1,15 @@
+#
+# Copyright 2018 Analytics Zoo Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


hkvision:
Use `mxnet-mkl-1.6.0b20191025`, set `OMP_NUM_THREADS=44` and `KMP_AFFINITY=granularity=fine,compact,1,0`, running 2 workers and 2 servers on 4 Almaren nodes got throughput ~162 imgs/s per worker for resnet18_v1 on cifar10, which is the same as using MXNet only.
@jason-dai

hkvision:
@jason-dai updated for both symbolic and imperative API. Take a look?

hkvision:
This is relatively outdated. I would create a new PR.

PR #1839, opened 2019-12-12
Refine python autotests

hkvision:


hkvision:
python example tests passed http://10.239.47.211:18888/job/ZOO-PR-Python-ExampleTests/60/

PR #1840, opened 2019-12-12
jep dataset

qiuxin2012:
Provide a way to reuse tensorflow's dataset.

like this:
```python
def train_fn():
    import tensorflow_datasets as tfds
    t = tfds.load(data_dir="/tmp/tfds/mnist", name="mnist", split=tfds.Split.TRAIN, as_supervised=True)
    return t.map(_normalize_train_img).shuffle(1024) \
        .batch(512).prefetch(tf.data.experimental.AUTOTUNE)


def test_fn():
    import tensorflow_datasets as tfds
    t = tfds.load(data_dir="/tmp/tfds/mnist", name="mnist", split=tfds.Split.TEST, as_supervised=True)
    return t.map(_normalize_test_img) \
        .batch(128).prefetch(tf.data.experimental.AUTOTUNE)

...

train_featureSet = FeatureSet.tf_dataset(train_fn)
test_featureSet = FeatureSet.tf_dataset(test_fn)

 ...

estimator.train_minibatch(train_featureSet,
                          zooCriterion, 
                          end_trigger=MaxEpoch(10), 
                          checkpoint_trigger=EveryEpoch(),
                          validation_set=test_featureSet, 
                          validation_method=[Accuracy()])
```

### DIFF
--- a/pyzoo/zoo/util/spark.py
+++ b/pyzoo/zoo/util/spark.py
@@ -207,6 +207,8 @@ class SparkRunner():
             else:
                 spark_conf["spark.executor.extraClassPath"] = zoo_bigdl_path_on_executor
 
+            spark_conf["spark.executorEnv.PYTHONHOME"] = self.PYTHON_ENV
+
             for item in spark_conf.items():
                 conf[str(item[0])] = str(item[1])
             sc = self._create_sc(submit_args, conf)
--- a/zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
@@ -16,19 +16,27 @@
 
 package com.intel.analytics.zoo.feature
 
+import java.nio.file.Paths
+import java.util
 import java.util.concurrent.atomic.AtomicInteger
 
 import com.intel.analytics.bigdl.DataSet
-import com.intel.analytics.bigdl.dataset.{AbstractDataSet, DistributedDataSet, Transformer}
+import com.intel.analytics.bigdl.dataset.{AbstractDataSet, DistributedDataSet, MiniBatch, Transformer}
+import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.utils.RandomGenerator
+import com.intel.analytics.zoo.core.TFNetNative
 import com.intel.analytics.zoo.feature.common.{ArrayLike, ArrayLikeWrapper}
 import com.intel.analytics.zoo.feature.pmem._
 import com.intel.analytics.zoo.pipeline.api.keras.layers.utils.EngineRef
+import com.intel.analytics.zoo.pipeline.api.net.{PytorchModel, PytorchModelWrapper}
+import org.apache.spark.{SparkContext, TaskContext}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.storage.StorageLevel
 import org.slf4j.{Logger, LoggerFactory}
+import jep._
 
 import scala.reflect.ClassTag
+import scala.collection.JavaConverters._
 
 /**
  * A set of data which is used in the model optimization process. The FeatureSet can be access in
@@ -186,7 +194,6 @@ trait DistributedFeatureSet[T] extends AbstractFeatureSet[T, RDD[T]] {
 /**
  * Wrap a featureSet to DistributedDataSet.
  * @param featureSet
- * @param ev$1
  * @tparam T
  */
 private[zoo] class DistributedDataSetWrapper[T: ClassTag](featureSet: DistributedFeatureSet[T])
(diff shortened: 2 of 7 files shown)


pyzoo/zoo/util/spark.py
```diff
@@ -207,6 +220,11 @@ def _submit_opt():
             else:
                 spark_conf["spark.executor.extraClassPath"] = zoo_bigdl_path_on_executor
 
+            ld_path = self._ld_path_for_executor()
+            spark_conf["spark.executorEnv.LD_PRELOAD"] = ld_path[0]
+            spark_conf["spark.executorEnv.LD_LIBRARY_PATH"] = ":".join(ld_path[1])
```
yangw1234: Override LD_LIBRARY_PATH might mess up executor evironment. It is safer to use spark.executor.extraLibraryPath.

qiuxin2012: removed, seems we only need `spark.executorEnv.PYTHONHOME`

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -321,6 +325,143 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+      |by = bytes(b % 256 for b in pyjarray)
+      |${loaderName} = pickle.loads(by)
+      |""".stripMargin
+    interpRdd.mapPartitions{iter =>
+      val interp = iter.next()
+      interp.exec(imports)
+      interp.set("pyjarray", bcDataSet.value)
+      interp.exec(load)
+      interp.exec("by = bytes(b % 256 for b in pyjarray)")
+      interp.exec("loader = pickle.loads(by)")
```
jason-dai: why execute load twice?

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -321,6 +325,143 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+      Iterator.single(interp)
+    }.count()
+
+  }
+
+  private var sharedInterpRDD: RDD[SharedInterpreter] = null
```
jason-dai: `jepRDD`

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -321,6 +325,143 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+            .cache()
+          originRdd.count()
+          originRdd.count()
+          // load pytorch library before jep, or libCaffe2.so will conflict.
+          originRdd.mapPartitions{
+            _ => PytorchModelWrapper.load()
+            Iterator.single(1)
+          }.count()
+          sharedInterpRDD = originRdd.mapPartitions { iter =>
+            val interp = getOrCreateInterpreter()
```
jason-dai: do we need to have a shared Interpreter?

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -321,6 +325,143 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+  }
+
+  override def data(train: Boolean): RDD[T] = {
+    val loaderName = this.loaderName
+    val iterName = s"${loaderName}_iter"
+    sharedInterp.mapPartitions{ dataIter =>
+      val interp = dataIter.next()
+      val len = interp.getValue(s"len($loaderName)").asInstanceOf[Long]
+      interp.exec(s"${iterName} = enumerate($loaderName)")
+      new Iterator[T] {
```
jason-dai: need to use `train` to customize its behavior

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -321,6 +325,143 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+        var i = 0
+        val nextCode =
+          s"""
+            |batch_idx, (data, target) = next($iterName)
```
jason-dai: what's the expected structure of the result from `next`?

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
```diff
@@ -321,6 +325,143 @@ class CachedDistributedFeatureSet[T: ClassTag]
...
+    }
+
+  }
+
+  override def shuffle(): Unit = {
+
+  }
+
+  override def size(): Long = {
+    data(false).count()
```
jason-dai: this may not work for TF DataSet @yangw1234

jason-dai:
we should refactor this PR so that the interface can support both PyTorch DataLoader and TensorFlow DataSet

jason-dai:
We should refactor this PR so that the interface can support both PyTorch DataLoader and TensorFlow DataSet @qiuxin2012 @yangw1234

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
qiuxin2012: We can create only one Shared Interpreter in a JVM. Or jep will throw a execption to notice use to close the former one.

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
qiuxin2012: Index + data + target, data and target are tensor or array of tensor.
We expected this kind of structure as MiniBatch and Optimizer only have this two fields.

yangw1234:
Could we define the interface like this?

FeatureSet.python accept a python function. The function **accepts a partition id and return a iterator of tuple of two**. The first element in the tuple is a list of numpy.ndarray representing the features and the second element is also a list of numpy.ndarray representing the targets.

I think this might be more flexible as user can define any data generation function.

jason-dai:
> Could we define the interface like this?
> 
> FeatureSet.python accept a python function. The function **accepts a partition id and return a iterator of tuple of two**. The first element in the tuple is a list of numpy.ndarray representing the features and the second element is also a list of numpy.ndarray representing the targets.
> 
> I think this might be more flexible as user can define any data generation function.

I think we can focus on supporting TF DataSet and PyTorch DataLoader at this moment; we can add support for generic python methods later.

zoo/src/main/scala/com/intel/analytics/zoo/feature/FeatureSet.scala
yangw1234: I think we can use similar policy used by keras. [tensorflow.keras.engine.training_utils.infer_steps_for_dataset](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/training_utils.py#L1762)
1. User can optionally pass the dataset size through our api
2. If user pass in the dataset size, we will use that size.
    If user does not pass in the dataset size and the size can be inferred from the static information of the dataset, then we will use the inferred size.
    Otherwise, throw an exception

pyzoo/zoo/examples/tensorflow/tfpark/tf_optimizer/train_lenet.py
```diff
@@ -16,43 +16,70 @@
...
+        executor_cores=num_cores_per_executor,
+        executor_memory="2g",
+        driver_memory="10g",
+        driver_cores=1,
+        spark_conf={"spark.rpc.message.maxSize": "1024",
+                    "spark.task.maxFailures":  "1",
+                    "spark.driver.extraJavaOptions": "-Dbigdl.failure.retryTimes=1"})
+
+
+    dataset = TFDataset.from_dataset(trainfunc,
```
qiuxin2012: cloud pickler can't pickle tf.data.dataset too. @jason-dai

qiuxin2012:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/2976/

PR #1841, opened 2019-12-12
Final PR of Cluster Serving

Litchilitchy:
Docs to be refine later


zoo/src/main/scala/com/intel/analytics/zoo/tensorboard/FileWriter.scala
```diff
@@ -39,7 +39,9 @@ private[zoo] class FileWriter(val logDirectory : String, flushMillis: Int = 1000
 
   private val eventWriter = new EventWriter(logDirectory, flushMillis, fs)
 
-  val threadPool = new ThreadPool(1)
+
```
Litchilitchy: @yangw1234 I changed this code back here, if use `ThreadPool(1)` it would get stuck. Not sure about the cause, so just used urgent fix.

yangw1234: ok

pyzoo/zoo/serving/quick_start.py
```diff
@@ -14,14 +14,14 @@
 # limitations under the License.
 #
 
-from zoo.serving.client.helpers import RedisQueue
+from zoo.serving.client.helpers import Input, Output
```
jason-dai: don't use `zoo.serving.client.helpers`; just `zoo.serving.client`

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -19,10 +19,19 @@
...
+    """
+    base level of API control
+    select data pipeline here, Redis/Kafka/...
+    interface preserved for API class
+    """
     def __init__(self, file_path=None):
+        dt = datetime.datetime.now()
+        self.log_file = open("./cluster-serving.log", "w")
         if file_path:
             with open(file_path) as f:
```
jason-dai: you should read `cluster-serving/config.yaml` if `file_path=None`

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
...
-        self.db.xadd("image_stream", d)
 
-    # def get_results(self, key):
-    #     return self.db.hgetall(key)
-    def get_results(self):
+        inf = self.db.info()
+
+        try:
+            if inf['used_memory'] >= inf['maxmemory'] * 0.6:
+                raise redis.exceptions.ConnectionError
```
jason-dai: you should shown meaningful error messages to the users here

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
...
 
         img_encoded = self.base64_encode_image(data)
+        uri = uri + str(datetime.datetime.now())
         d = {"uri": uri, "image": img_encoded}
-        self.db.xadd("image_stream", d)
 
-    # def get_results(self, key):
-    #     return self.db.hgetall(key)
-    def get_results(self):
+        inf = self.db.info()
```
jason-dai: do we need to do this for every enqueue? what's the impact on performance and redis load?

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
...
+                raise redis.exceptions.ConnectionError
+            self.db.xadd("image_stream", d)
+            print("Write to Redis successful")
+        except redis.exceptions.ConnectionError:
+            print("Redis queue is full, please wait for inference "
+                  "or delete the unprocessed records.")
+            time.sleep(1)
+
+        except redis.exceptions.ResponseError as e:
+            print(e, "Redis queue is full, please dequeue or delete.")
```
jason-dai: why different error messages in the two cases?

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -39,6 +48,11 @@ def __init__(self, file_path=None):
         for i in range(len(self.data_shape)):
             self.data_shape[i] = int(self.data_shape[i])
 
+
+class Input(API):
```
jason-dai: `InputQueue`

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
...
+            print(e, "Redis queue is full, please dequeue or delete.")
+            time.sleep(1)
+
+    @staticmethod
+    def base64_encode_image(img):
+        # base64 encode the input NumPy array
+        return base64.b64encode(img).decode("utf-8")
+
+
+class Output(API):
```
jason-dai: `OutputQueue`

pyzoo/zoo/serving/client/helpers.py
Litchilitchy: in most cases, the error is first case, while the stream size is beyond our limit 60% of total, but there may be second case that the result are too much and the total memory is exhausted thus the stream size is not beyond 60% and just results take too much space, in this case we suggest user to dequeue

pyzoo/zoo/serving/client/helpers.py
Litchilitchy: not tested yet, but impact on performance is not obvious. Pushing is still very fast, single thread throughput over 100 at least.

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -37,6 +38,32 @@ object ClusterServing {
   Logger.getLogger("com.intel.analytics.zoo").setLevel(Level.INFO)
 
 
+  case class Rec(uri: String, value: String)
```
jason-dai: `Record`? `Request`?

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
         data = cv2.imencode(".jpg", img)[1]
 
         img_encoded = self.base64_encode_image(data)
+        uri = uri + str(datetime.datetime.now())
```
jason-dai: why append the timestamp?

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
...
+        super().__init__(file_path)
+
+    def dequeue(self):
         res_list = self.db.keys('result:*')
         decoded = {}
         for res in res_list:
             res_dict = self.db.hgetall(res.decode('utf-8'))
-            res_id = res_dict[b'_1'].decode('utf-8')
-            res_value = res_dict[b'_2'].decode('utf-8')
+            res_id = res.decode('utf-8').split(":")[1]
```
jason-dai: why split?

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
...
         return decoded
 
-    @staticmethod
-    def base64_encode_image(img):
-        # base64 encode the input NumPy array
-        return base64.b64encode(img).decode("utf-8")
+    def query(self, uri):
+        res_dict = self.db.hgetall("result:"+uri)
+        if not res_dict:
+            raise EOFError("Record does not exist.")
```
jason-dai: I don't think it's error; just return null

pyzoo/zoo/serving/client/helpers.py
Litchilitchy: just for test, will delete later. The same uri will overwrite the previous one in output table, thus in test if there is no time stamp the result table can never grow very large.

pyzoo/zoo/serving/client/helpers.py
Litchilitchy: result is regulated by redis schema `result:uri`, and we should also add a limit that uri could not contain any `:`

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -85,14 +106,27 @@ object ClusterServing {
     val redisDB = new Jedis(helper.redisHost, helper.redisPort.toInt)
 
     val query = images.writeStream.foreachBatch{ (batchDF: DataFrame, batchId: Long) =>
+
+      if (helper.updateConfig()) {
```
jason-dai: put this in another method

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -164,16 +198,40 @@ object ClusterServing {
 
 
         val resDf = spark.createDataFrame(resultPartitions)
-        resDf.write
-          .format("org.apache.spark.sql.redis")
-          .option("table", "result")
-          .mode(SaveMode.Append).save()
 
+        var errFlag: Boolean = true
```
jason-dai: put this in another method

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -164,16 +198,40 @@ object ClusterServing {
...
+        while (errFlag) {
+          try {
+            resDf.write
+              .format("org.apache.spark.sql.redis")
+              .option("table", "result")
+              .option("key.column", "uri")
+              .mode(SaveMode.Append).save()
+            println("result saved at " + microBatchSize.toString)
```
jason-dai: don't use println

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -164,16 +198,40 @@ object ClusterServing {
...
+            case e: redis.clients.jedis.exceptions.JedisDataException =>
+              errFlag = true
+              println("not enough space in redis to write !" + e.toString)
+              Thread.sleep(1000)
+            case e: Exception => {
+              errFlag = true
+              println("--!! never seen" + e.toString)
+              Thread.sleep(1000)
+            }
+
```
jason-dai: ideally we should also check for stopping the service here

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -164,16 +198,40 @@ object ClusterServing {
...
+              redisDB.xlen("image_stream") - previousLen, true)
+            println("Xtrim completed")
+            errFlag = false
+          }
+
+          catch {
+            case e: redis.clients.jedis.exceptions.JedisDataException =>
+              errFlag = true
+              println("not enough space in redis to write !" + e.toString)
+              Thread.sleep(1000)
```
jason-dai: Is there a way to exist from this error loop?

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -165,14 +175,40 @@ class ClusterServingHelper {
 //    engineType = getYaml(paramsConfig, "engine_type", "mklblas")
     topN = getYaml(paramsConfig, "top_n", "5").toInt
 
+    val logConfig = configList.get("log").asInstanceOf[HM]
+    logErrorFlag = if (getYaml(logConfig, "error", "y") ==
```
jason-dai: `logErrorFlag` should always be true - it makes no sense if we don't report error

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -165,14 +175,40 @@ class ClusterServingHelper {
...
         .toLowerCase() == "mklblas") {
         blasFlag = true
       }
+      else blasFlag = false
     }
+    else blasFlag = false
 
   }
+  def updateConfig(): Boolean = {
+    val lastModTime = Files.getLastModifiedTime(Paths.get(configPath)).toString
```
jason-dai: don't overload config file for checking restarting; instead, using some other flags (other files or even some flags stored in redis)

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -85,14 +106,27 @@ object ClusterServing {
     val redisDB = new Jedis(helper.redisHost, helper.redisPort.toInt)
 
     val query = images.writeStream.foreachBatch{ (batchDF: DataFrame, batchId: Long) =>
+
+      if (helper.updateConfig()) {
+        loadSerialParams(helper)
+        if (bcModel != null) bcModel.destroy()
+        if (model != null) model.doRelease()
```
jason-dai: we still need to
1) release the broadcasted model in each executor
2) release cached buffer (bytearray, graph, etc.) in driver and executors

jason-dai:
1) do we have (automated) tests for all the error handling?
2) how to we manage the errors reported by spark (on both driver and executors)?
3) we should provide the command to start, stop and restart the service

docker/cluster-serving/config.yaml
```diff
@@ -6,33 +6,30 @@ model:
 data:
   # default, localhost:6379
   src: localhost:6379
-
   # default, 3, 224, 224
   shape: 3,224,224
+params:
+  # default, 4
```
jason-dai: all default values should be the same as the values provided in the config file

docker/cluster-serving/config.yaml
```diff
@@ -6,33 +6,30 @@ model:
 data:
   # default, localhost:6379
   src: localhost:6379
```
jason-dai: `input` instead of `src`, so as to be consistent with client API

docker/cluster-serving/config.yaml
```diff
@@ -6,33 +6,30 @@ model:
...
-
   # default, 3, 224, 224
   shape: 3,224,224
+params:
+  # default, 4
+  batch_size: 64
+  # default, mklblas
+  engine_type: mkldnn
```
jason-dai: why mkldnn?

docker/cluster-serving/config.yaml
```diff
@@ -6,33 +6,30 @@ model:
 data:
   # default, localhost:6379
   src: localhost:6379
-
   # default, 3, 224, 224
   shape: 3,224,224
```
jason-dai: image_shape

docker/cluster-serving/config.yaml
```diff
@@ -6,33 +6,30 @@ model:
...
+params:
+  # default, 4
+  batch_size: 64
+  # default, mklblas
+  engine_type: mkldnn
+  # default, 1
+  top_n: 5
+log:
```
jason-dai: why do we need to expose this to users? just always output error and Tensorboard messages.

pyzoo/zoo/serving/client/helpers.py
```diff
@@ -52,22 +66,48 @@ def enqueue_image(self, uri, img):
         data = cv2.imencode(".jpg", img)[1]
```
jason-dai: how do we make sure `img` is valid when used in this method call?

docs/docs/ClusterServingGuide/APIGuide.md
```diff
@@ -0,0 +1,94 @@
...
+output_api = Output()
+d = output_api.query("my-image") 
+json.loads(d)
+```
+
+#### dequeue
+[view source]()
+
+```
+get_result()
```
jason-dai: should be `dequeue` instead, yes?

docs/docs/ClusterServingGuide/VisualizationAndLog.md
```diff
@@ -0,0 +1,35 @@
+# Visualization and Log
+This page shows how to see log and visualization in Analytics Zoo Cluster Serving.
+## Visualization
+We integrate Tensorboard into Cluster Serving, to enable this feature, use following config in [Configuration](). By default, this feature is disabled. By disabling this feature, you could have a slight gain of serving performance since there is some cost to stat the information.
+```
```
jason-dai: I don't think we need a config for this; we should always enable it

docs/docs/ClusterServingGuide/VisualizationAndLog.md
```diff
@@ -0,0 +1,35 @@
...
+log:
+  summary: y
+```
+Tensorboard service is started with Cluster Serving, once your serving is run, you can go to `localhost:6006` to see visualization of your serving.
+
+Analytics Zoo Cluster Serving provides 3 attributes in Tensorboard so far, `Micro Batch Throughput`, `Partition Number`, `Total Records Number`.
+
+* `Micro Batch Throughput`: The overall throughput, including preprocessing and postprocessing of your serving, the line should be relatively stable after first few records. If this number has a drop and remains lower than previous, you might have lost the connection of some nodes in your cluster.
```
jason-dai: change to `Serving Throughput`; it should be records per second

docs/docs/ClusterServingGuide/VisualizationAndLog.md
```diff
@@ -0,0 +1,35 @@
...
+
+* `Partition Number`: The partition number of your serving, this number should be stable all the time, and note that if you have N nodes in your cluster, you should have this partition number at least N.
```
jason-dai: why do we need this?

docs/docs/ClusterServingGuide/VisualizationAndLog.md
```diff
@@ -0,0 +1,35 @@
...
+
+* `Total Records Number`: The total number of records that serving gets so far.
+
+**Note**: If you run serving on local mode, you could get another attribute `Throughput`, this is the throughput of prediction only, regardless of preprocessing and post processing. If you run serving on cluster mode, you could only see this attribute on remote nodes.
```
jason-dai: why do we need this if we cannot get it in distributed mode?

docs/docs/ClusterServingGuide/VisualizationAndLog.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+## Log
+We use log to save serving information and error, to enable this feature, use following config in [Configuration](). By default, this feature is enabled.
+```
+log:
+  error: y
```
jason-dai: we should always enable error messages, no?

docs/docs/ClusterServingGuide/Install.md
```diff
@@ -0,0 +1,14 @@
+# Install and Run
+Currently Analytics Zoo Cluster Serving supports install and run with docker.
+## Install and Run with Docker
+To install with docker, please see following steps.
+
+1. Install [Docker]()
+
+2. Get Analytics Zoo Cluster Serving docker image.
+
+3. Clone Analytics Zoo repository or download the zip and unzip it.
```
jason-dai: why do we need this?

docs/docs/ClusterServingGuide/Install.md
```diff
@@ -0,0 +1,14 @@
...
+
+4. Go to `analytics-zoo/docker/cluster-serving`. Set configuration in `config.yaml`, for details of configuration, see [Configuration Guide]().
+
+5. run `docker run --name cluster-serving --net=host -v $(pwd)/model:/opt/work/model -v $(pwd)/config.yaml:/opt/work/config.yaml analytics-zoo/cluster-serving:0.7.0-spark_2.4.0`
```
jason-dai: this is too complex; if needed, provide a script to make it easier for the user

docs/docs/ClusterServingGuide/Install.md
```diff
@@ -0,0 +1,14 @@
+# Install and Run
+Currently Analytics Zoo Cluster Serving supports install and run with docker.
```
jason-dai: why do we assume docker only?

pyzoo/zoo/serving/client/helpers.py
Litchilitchy: if `img` is loaded by `opencv-python`, it must be valid. If some invalid file is loaded by this API, `opencv-python` would return None. If user uses `opencv-python`, then we just check if it is None, otherwise there is no guarantee, e.g. push a arbitrary ndarray. If we want to make all guarantee, we must set this param to image path, or we do more shape and data range checks.

docker/cluster-serving/config.yaml
Litchilitchy: ok

docker/cluster-serving/config.yaml
Litchilitchy: `config.yaml` file usually pushed with my test, will deleted this option for user. No dnn in this release

docs/docs/ClusterServingGuide/APIGuide.md
Litchilitchy: yes, will refine docs later based on recent modifications.

docs/docs/ClusterServingGuide/VisualizationAndLog.md
Litchilitchy: ok

docs/docs/ClusterServingGuide/VisualizationAndLog.md
Litchilitchy: have not yet figure out a method to stat every second, the trigger is counted by micro-batch.

docs/docs/ClusterServingGuide/VisualizationAndLog.md
Litchilitchy: will delete

docs/docs/ClusterServingGuide/VisualizationAndLog.md
Litchilitchy: This is pure predict time. However the user may not understand, can also be deleted.

docs/docs/ClusterServingGuide/VisualizationAndLog.md
Litchilitchy: will always enable and remove this option in config.

docs/docs/ClusterServingGuide/Install.md
Litchilitchy: also support other ways to run, will re-write this part.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -198,7 +200,17 @@ class ClusterServingHelper {
     }
     else blasFlag = false
 
+    new File("started").createNewFile()
+
   }
+  def checkStop(): Boolean = {
+    if (!Files.exists(Paths.get("started"))) {
```
jason-dai: better to check "stopped" file?

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -184,7 +185,8 @@ class ClusterServingHelper {
     if (logErrorFlag) {
       logFile = {
         val logF = new File("./cluster_serving.log")
-        logF.createNewFile()
+        if (Files.exists(Paths.get("./cluster_serving.log")))
```
jason-dai: do we want to overwrite existing log files? maybe create a new one?

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
Litchilitchy: From our current analysis, better to check "started", if check "stopped", when user runs `stop.sh` it will generate a "stopped", if it is not running, it will also generate and next time user starts the serving the serving will stop. This is because user can run multiple times scripts. But if we put this file generation in code, user have to submit multiple spark jobs to generate this error, which is anyway we want to prevent(run multiple serving)

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
Litchilitchy: we can 1) always append to a single log file 2) create a new one with file name identified by time stamp 3) overwrite. I think maybe 1) is better

docker/cluster-serving/docker-build.sh
```diff
@@ -1,6 +1,5 @@
 docker build \
     --build-arg http_proxy=http://10.239.4.101:913/ \
```
glorysdj: change to proxy-host:proxy-port

docker/cluster-serving/zoo-cluster-serving.sh
```diff
@@ -2,14 +2,19 @@
 
 # --------------config
 
-/opt/work/redis-5.0.5/src/redis-server --port $REDIS_PORT > /opt/work/redis.log &
-echo "redis server started, please check log in /opt/work/redis.log" &
-sleep 1
+
+/opt/work/redis-5.0.5/src/redis-server > /opt/work/redis.log &
+echo "redis server started, please check log in /opt/work/redis.log" && sleep 1
```
glorysdj: put XXX.logs into logs folder
we may need to create a analytics-zoo-cluster-serving folder in /opt/work
and put related files into folders like bin, logs, lib, etc.

docker/cluster-serving/zoo-cluster-serving.sh
```diff
@@ -2,14 +2,19 @@
...
 
 # sleep for 1 sec to ensure server is ready and client could connect
 /opt/work/redis-5.0.5/src/redis-cli config set stop-writes-on-bgsave-error no
 /opt/work/redis-5.0.5/src/redis-cli config set save ""
```
glorysdj: bind 0.0.0.0 is requred

Litchilitchy: I think we can put `start,stop` in a `scripts` folder and build in docker image, and use another start, stop refer to `start,stop` in container in our `docker `folder`

jason-dai:
Let's focus on fixing the implementation in this PR, and leave document fix in another PR.

docs/docs/ClusterServingGuide/VisualizationAndLog.md
jason-dai: I meant that the value of `Serving Throughput` should be records per second

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
jason-dai: I would prefer 2); the user can always look for the latest log

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
jason-dai: change it to `running` then

pyzoo/zoo/serving/client.py
```diff
@@ -29,17 +28,21 @@ class API:
...
-                config['data']['port'] = host_port[1]
-        else:
-            config = {'data': {'host': "localhost", 'port': "6379", 'shape': "3,224,224"}}
+        import os
+        a = os.listdir("../../..")
+
+        try:
+            file_path = "../../../docker/cluster-serving/config.yaml"
+        except:
+            raise EOFError("config file does not exist!")
```
jason-dai: show the expected file path in the error message

pyzoo/zoo/serving/client.py
```diff
@@ -49,24 +52,44 @@ def __init__(self, file_path=None):
...
+class InputQueue(API):
+    def __init__(self):
+        super().__init__()
+        self.c, self.h, self.w = None, None, None
+        self.data_shape_check()
+
+    def data_shape_check(self):
+        for num in self.data_shape:
+            if num <= 0:
+                raise Exception("Your image shape config is invalid, please check.")
```
jason-dai: show the image shape value in the error message

pyzoo/zoo/serving/client.py
```diff
@@ -49,24 +52,44 @@ def __init__(self, file_path=None):
...
+            if 0 < num < 5:
```
jason-dai: shall we assume that shape is always (C, H, W)?

pyzoo/zoo/serving/client.py
```diff
@@ -49,24 +52,44 @@ def __init__(self, file_path=None):
...
         :return:
         """
-        if img.shape[0] != self.data_shape[1] or img.shape[1] != self.data_shape[2]:
-
-            raise AssertionError("Your image shape " + str(img.shape) + " does not match "
-                                 "that in your config " + str(self.data_shape) + ", please check")
+        if isinstance(img, str):
+            img = cv2.imread(str)
+            if not img:
+                print("You have pushed an image with invalid path, skipped.")
```
jason-dai: show the file path value in the error message

pyzoo/zoo/serving/client.py
Litchilitchy: this code will infer shape as CHW or HWC, but H is always before W, C would be given value only in range (0,5)

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -246,20 +247,27 @@ object ClusterServing {
...
         val microBatchThroughPut = (microBatchSize / microBatchLatency).toFloat
+
+        totalCnt += microBatchSize.toInt
+
         if (model.inferenceSummary != null) {
-          model.inferenceSummary.addScalar(
-            "Micro Batch Throughput", microBatchThroughPut, batchId)
+          (timeStamp until timeStamp + microBatchLatency.toInt).foreach( time => {
+            model.inferenceSummary.addScalar(
+              "Micro Batch Throughput", microBatchThroughPut, time)
```
jason-dai: change to "Serving Throughput"

Litchilitchy:
@jason-dai @glorysdj @qiyuangong 

All codes seems ready.

Except that summary on executor need too much handling and, in fact, this is a dispensable feature, so I deleted this feature.

pyzoo/zoo/serving/client.py
```diff
@@ -0,0 +1,141 @@
...
+        data = cv2.imencode(".jpg", img)[1]
+
+        img_encoded = self.base64_encode_image(data)
+
+        d = {"uri": uri, "image": img_encoded}
+
+        inf = self.db.info()
+
+        try:
+            if inf['used_memory'] >= inf['maxmemory'] * 0.6:
```
glorysdj: do we need a Convention object or a config file to hold parameters like 0.6, etc.

Litchilitchy: maybe no? users do not know these concepts and they do not know the size of stream and result, e.g. image stream per image is much larger than a result string.

glorysdj:
if we want to support running with or without docker, do we need to build a tar and let user to download and unzip it as a cluster serving distribution to start service quickly.

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -166,19 +185,15 @@ object ClusterServing {
...
-//                  .Base64.getDecoder.decode(pathBytesBatch(i).getAs[String]("image")))
-//                val path = pathBytesBatch(i).getAs[String]("uri")
-
-
               })
             })
           })
           pathBytesChunk.mapPartitions(pathBytes => {
             val localModel = bcModel.value
+            val t = Tensor[Float](batchSize, C, W, H)
```
jason-dai: what if `batchSize` is different from `thisBatchSize`?

Litchilitchy: will still predict with `batchSize`, but some records are empty (those out of `thisBatchSize`), and will not be returned in Results. So the results will be the same but this batch inference time will still not be reduced compared with full batch.

However, we still need to use `batchSize` to generate Tensor, on one hand it can be reused, on the other hand even if we new a Tensor every time, it will raise shape error in `forward` method. It seems the first `forward` has set some shape attributes and it assumes that all inputs will be the same in one program.

jason-dai: The records are not empty; could be random. I think we can resize the Tensor accordingly

Litchilitchy: So resize the Tensor after predicting? will try

In this way each partition will do resize once, which is the last batch being the only batch satisfying (`thisBatchSize != batchSize`)

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -39,29 +39,36 @@ object ClusterServing {
...
+//    topN = helper.topN
+//    coreNum = helper.coreNum
+//    nodeNum = helper.nodeNum
+//    modelType = helper.modelType
+//    blasFlag = helper.blasFlag
+//
+//    C = helper.dataShape(0)
+//    W = helper.dataShape(1)
+//    H = helper.dataShape(2)
+//  }
```
jason-dai: remove the code instead of commenting it out

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -123,29 +142,39 @@ object ClusterServing {
 //        if (helper.logSummaryFlag) model.setInferenceSummary(
 //          InferenceSummary(".", helper.dateTime + "-ClusterServing"))
 //      }
```
jason-dai: remove the code instead of commenting it out

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -201,8 +154,9 @@ class ClusterServingHelper {
 
     logFile = {
       val logF = new File("./cluster_serving.log")
-      if (Files.exists(Paths.get("./cluster_serving.log")))
+      if (Files.exists(Paths.get("./cluster_serving.log"))) {
```
jason-dai: are we appending or overwriting here?

pyzoo/zoo/serving/client.py
jason-dai: But why don't we ask the user to specify the image_shape as (C, H, W)?

pyzoo/zoo/serving/client.py
jason-dai: Maybe we can make some of the values (e.g., 0.6 here, 3000 in wait, 1s in listener, etc) configurable; we don't need to expose them to users, but that we may change them in deployment if needed.

pyzoo/zoo/serving/client.py
Litchilitchy: Sure we can, so throw Error if not CHW?

PR #1851, opened 2019-12-17
Remove repartition in wide and deep notebook

hkvision:
This is unnecessary and may cause performance overhead.
repartition(4) the assumption of 4 cores locally is also not reasonable.


jason-dai:
LGTM

PR #1853, opened 2019-12-18
update example notebook

yushan111:
1. update example notebook
    

- remove 

```python
    import seaborn
``` 
since it triggers the kernel died while running the code below in a built docker image
```python
    import TimeSequencePredictor as tsp 
```

- change metrics name

- use `RandomRecipe` since the default `SmokeRecipe` doesn't fit very well.


glorysdj:
LGTM

PR #1856, opened 2019-12-19
Cluster Serving bugs fix, and scripts

Litchilitchy:


Litchilitchy:
PR summary:

1) add CHW or HWC flag, to avoid Tensorflow transpose of 4D Tensor, increasing the benchmark.

2) New approach of Redis control, by using percentage to trim instead of trimming processed stream, test passed and works well so far.

3) adding scripts.

This PR is directly re-opened from the previous merged one, thus commits are in mess... To check the newest updates, just check last 3 commits.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -81,13 +81,19 @@ class ClusterServingHelper {
   var logErrorFlag: Boolean = true
   var logSummaryFlag: Boolean = false
 
+  /**
+   * model relative
```
jason-dai: model related

PR #1862, opened 2019-12-22
Fix ONNX unit tests

yangw1234:
1. Map Cast node to Identity, as all output in Analytics-Zoo is float, cast is already done implicitly 
2. Add print graph code if loading fails


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2896/ onnx test case fixed, there still a flaky test related to torchnet

PR #1863, opened 2019-12-22
Fix torchnet predict resnet test connection error

yangw1234:


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2899/
torchnet case fixed
still have one flaky example test

PR #1870, opened 2019-12-25
Cluster serving doc

Litchilitchy:

PR #1871, opened 2019-12-25
Serving scripts

Litchilitchy:


docker/cluster-serving/config.yaml
```diff
@@ -0,0 +1,28 @@
+## Analytics-zoo Cluster Serving
+
+model:
+  # model path must be set
+  path: /opt/work/model/
+data:
+  # default, localhost:6379
+  src:
```
jason-dai: how to configure it when client is running on a remote node?

docker/cluster-serving/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+  # default, 3, 224, 224
+  image_shape:
+params:
+  # default, 4
+  batch_size:
+  # default, 1
+  top_n:
+spark:
+  # default, local[*]
+  master:
```
jason-dai: combine this (cluster) and model into a new category? Only these two are mandatory.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -58,7 +59,11 @@ class ClusterServingHelper {
   type HM = LinkedHashMap[String, String]
 
 //  val configPath = "zoo/src/main/scala/com/intel/analytics/zoo/serving/config.yaml"
-  val configPath = "config.yaml"
+  val configPath = if (env.contains("ANALYTICS_ZOO_HOME")) {
+    env("ANALYTICS_ZOO_HOME") + "/scripts/cluster-serving/config.yaml"
+  } else {
+    "config.yaml"
+  }
```
jason-dai: I don't think pip users will know ANALYTICS_ZOO_HOME? @hkvision

Litchilitchy: For pip user they can not go to our serving directory. By default we ask user to run python script in serving directory so that we can use relative path to find the config file. Actually we just need to find our config file here. If using `pip install` can we find this?

hkvision: Yeah we don't need env variable. I think we can simulate the way we find the JAR, i.e. put the config in a specific path and find it automatically.

Litchilitchy: python can find the path now, when pip install it is in `usr/lib/pythonX/dist-packages/..` and we can use `__file__` to get the relative path no matter where the user run the zoo python lib, but don't know how to find it in scala..(code here)

Litchilitchy: @jason-dai I think we can add a param of config path in scala. By the way we have abandoned all params and move them in config, so maybe I should add this param back. And in this way we can pass this param to scala when we use python to start serving

Litchilitchy: I tried to fix it here https://github.com/intel-analytics/analytics-zoo/pull/1870#issuecomment-569564612. Code finish, testing now.

jason-dai:
Several comments:
1) Why there are both `cluster-serving-start` and `start-cluster-serving.sh` (and for stop and restart as well)?
2) How (and when) to shut down Redis? What if Redis is already running when start cluster serving? (and similarly for TensorBoard)
3) for Redis related configations (such as REDIS_MAXMEM), we may need to put it in yaml too.

Litchilitchy:
> 
> 
> Several comments:
> 
>     1. Why there are both `cluster-serving-start` and `start-cluster-serving.sh` (and for stop and restart as well)?
> 
>     2. How (and when) to shut down Redis? What if Redis is already running when start cluster serving? (and similarly for TensorBoard)
> 
>     3. for Redis related configations (such as REDIS_MAXMEM), we may need to put it in yaml too.

1. should be both cluster-serving-start, will change name

2. If docker I think we can shut down with docker, else user should shut down by themselves 

3. ok

jason-dai:
> > Several comments:
> > ```
> > 1. Why there are both `cluster-serving-start` and `start-cluster-serving.sh` (and for stop and restart as well)?
> > 
> > 2. How (and when) to shut down Redis? What if Redis is already running when start cluster serving? (and similarly for TensorBoard)
> > 
> > 3. for Redis related configations (such as REDIS_MAXMEM), we may need to put it in yaml too.
> > ```
> 
> 1. should be both cluster-serving-start, will change name

But why do we need both? Can we keep just one?

> 2. If docker I think we can shut down with docker, else user should shut down by themselves

What will happen if Redis is already running when start cluster serving?

> 3. ok

Litchilitchy:
> 
> 
> > > Several comments:
> > > ```
> > > 1. Why there are both `cluster-serving-start` and `start-cluster-serving.sh` (and for stop and restart as well)?
> > > 
> > > 2. How (and when) to shut down Redis? What if Redis is already running when start cluster serving? (and similarly for TensorBoard)
> > > 
> > > 3. for Redis related configations (such as REDIS_MAXMEM), we may need to put it in yaml too.
> > > ```
> > 
> > 
> > 
> > 1. should be both cluster-serving-start, will change name
> 
> But why do we need both? Can we keep just one?
>  they are the same, the 3 in `docker` dir are for docker build, (maybe docker build can not go to upper directories, so I just copy them here, for users they can always see only 3, not 6)   

> > 1. If docker I think we can shut down with docker, else user should shut down by themselves
> 
> What will happen if Redis is already running when start cluster serving?
>  Nothing happens with redis, redis will not start again, just submit spark job. 

> > 1. ok

jason-dai:
Please summarize the semantics/behaviors of these different commands, such as:
1) cluster-serving-start
    - Start Redis if not already running
    - Start Tensorboard?
    - Start cluster serving 
2) cluster-serving-stop
    - ...
3) cluster-serving-restart
    - ...

Litchilitchy:
> 
> 
> Please summarize the semantics/behaviors of these different commands, such as:
> 
>     1. cluster-serving-start
>        
>        * Start Redis if not already running
>        * Start Tensorboard?
>        * Start cluster serving
> 
>     2. cluster-serving-stop
>        
>        * ...
> 
>     3. cluster-serving-restart
>        
>        * ...

My idea based on latest dicussion:
1. cluster-serving-start
start serving
2. cluster-serving-stop
stop serving
3. cluster-serving-restart
stop serving, then start serving

(all of them do not access redis and tensorboard)

I think if users choose to not use docker, to deploy redis by themselves, maybe we better not give control to their redis and tensorboard. So I think we can put start redis and tensorboard in our docker run, and after that, user can go into docker container to start serving.

If not using docker, just get all dependencies ready and `cluster-serving-start`.

Or if accessing redis or tensorboard is better in `cluster-serving-start`, I would add them. By the way, if users deploy redis by themselves, we do not know where redis locates... How to start..?

Litchilitchy:
@jason-dai based on current behavior, running docker does not start serving. I think it is better to combine start, stop and launching serving in doc. Updated in [doc](https://github.com/Litchilitchy/analytics-zoo/blob/cluster-serving-doc/docs/docs/ClusterServingGuide/ProgrammingGuide.md)

pyzoo/zoo/serving/server.py
```diff
@@ -0,0 +1,110 @@
...
+        except Exception:
+            print("WARNING: An initialized config file already exists.")
+        """
+        code below this line are all deprecated
+        """
+
+        self.serving_start_path = os.path.join(__file__ + "/../../"
+                                               'bin/cluster-serving/start-cluster-serving.sh')
+        self.serving_stop_path = os.path.join(__file__ + "/../../"
+                                              'bin/cluster-serving/start-cluster-serving.sh')
```
jason-dai: should be `stop-cluster-serving.sh`?

pyzoo/zoo/serving/server.py
```diff
@@ -0,0 +1,110 @@
...
+        self.serving_restart_path = os.path.join(__file__ + "/../../"
+                                                 'bin/cluster-serving/start-cluster-serving.sh')
```
jason-dai: should be `restart-cluster-serving.sh`?

jason-dai:
> > Please summarize the semantics/behaviors of these different commands, such as:
> > ```
> > 1. cluster-serving-start
> >    
> >    * Start Redis if not already running
> >    * Start Tensorboard?
> >    * Start cluster serving
> > 
> > 2. cluster-serving-stop
> >    
> >    * ...
> > 
> > 3. cluster-serving-restart
> >    
> >    * ...
> > ```
> 
> My idea based on latest dicussion:
> 
> 1. cluster-serving-start
>    start serving
> 2. cluster-serving-stop
>    stop serving
> 3. cluster-serving-restart
>    stop serving, then start serving
> 
> (all of them do not access redis and tensorboard)
> 
> I think if users choose to not use docker, to deploy redis by themselves, maybe we better not give control to their redis and tensorboard. So I think we can put start redis and tensorboard in our docker run, and after that, user can go into docker container to start serving.
> 
> If not using docker, just get all dependencies ready and `cluster-serving-start`.
> 
> Or if accessing redis or tensorboard is better in `cluster-serving-start`, I would add them. By the way, if users deploy redis by themselves, we do not know where redis locates... How to start..?

I think it makes sense to require that both redis and tensorboard are in execution path? We can then start/stop them as needed.

pyzoo/zoo/serving/server.py
Litchilitchy: We are using `/bin/bash` scripts now so this is no longer used, will delete

pyzoo/zoo/serving/server.py
Litchilitchy: same as above, these code will be removed

Litchilitchy:
> 
> 
> > > Please summarize the semantics/behaviors of these different commands, such as:
> > > ```
> > > 1. cluster-serving-start
> > >    
> > >    * Start Redis if not already running
> > >    * Start Tensorboard?
> > >    * Start cluster serving
> > > 
> > > 2. cluster-serving-stop
> > >    
> > >    * ...
> > > 
> > > 3. cluster-serving-restart
> > >    
> > >    * ...
> > > ```
> > 
> > 
> > My idea based on latest dicussion:
> > 
> > 1. cluster-serving-start
> >    start serving
> > 2. cluster-serving-stop
> >    stop serving
> > 3. cluster-serving-restart
> >    stop serving, then start serving
> > 
> > (all of them do not access redis and tensorboard)
> > I think if users choose to not use docker, to deploy redis by themselves, maybe we better not give control to their redis and tensorboard. So I think we can put start redis and tensorboard in our docker run, and after that, user can go into docker container to start serving.
> > If not using docker, just get all dependencies ready and `cluster-serving-start`.
> > Or if accessing redis or tensorboard is better in `cluster-serving-start`, I would add them. By the way, if users deploy redis by themselves, we do not know where redis locates... How to start..?
> 
> I think it makes sense to require that both redis and tensorboard are in execution path? We can then start/stop them as needed.

Then if user decides to install them by themselves they have to provide redis home path as well as tensorboard

jason-dai:
> > > > Please summarize the semantics/behaviors of these different commands, such as:
> > > > ```
> > > > 1. cluster-serving-start
> > > >    
> > > >    * Start Redis if not already running
> > > >    * Start Tensorboard?
> > > >    * Start cluster serving
> > > > 
> > > > 2. cluster-serving-stop
> > > >    
> > > >    * ...
> > > > 
> > > > 3. cluster-serving-restart
> > > >    
> > > >    * ...
> > > > ```
> > > 
> > > 
> > > My idea based on latest dicussion:
> > > 
> > > 1. cluster-serving-start
> > >    start serving
> > > 2. cluster-serving-stop
> > >    stop serving
> > > 3. cluster-serving-restart
> > >    stop serving, then start serving
> > > 
> > > (all of them do not access redis and tensorboard)
> > > I think if users choose to not use docker, to deploy redis by themselves, maybe we better not give control to their redis and tensorboard. So I think we can put start redis and tensorboard in our docker run, and after that, user can go into docker container to start serving.
> > > If not using docker, just get all dependencies ready and `cluster-serving-start`.
> > > Or if accessing redis or tensorboard is better in `cluster-serving-start`, I would add them. By the way, if users deploy redis by themselves, we do not know where redis locates... How to start..?
> > 
> > 
> > I think it makes sense to require that both redis and tensorboard are in execution path? We can then start/stop them as needed.
> 
> Then if user decides to install them by themselves they have to provide redis home path as well as tensorboard

For Redis, just ask them to add its path to env PATH; for TensorBoard, ask them to use pip install.

Litchilitchy:
To summarize:

start: start redis and tensorboard if they are not running, and start serving

stop: stop serving

restart: stop and start

Docker: when docker run, start redis and tensorboard? Or just don't do anything.

jason-dai:
> To summarize:
> 
> start: start redis and tensorboard if they are not running, and start serving
> 
> stop: stop serving
> 
> restart: stop and start
> 
> Docker: when docker run, start redis and tensorboard? Or just don't do anything.

For Docker, we don't need to do anything special.
For stop, maybe we can add a `stop all` option to stop everything.

Litchilitchy:
@jason-dai 

updated in this [section](https://github.com/Litchilitchy/analytics-zoo/blob/cluster-serving-doc/docs/docs/ClusterServingGuide/ProgrammingGuide.md#3-launching-service)

jason-dai:
> > > > Several comments:
> > > > ```
> > > > 1. Why there are both `cluster-serving-start` and `start-cluster-serving.sh` (and for stop and restart as well)?
> > > > 
> > > > 2. How (and when) to shut down Redis? What if Redis is already running when start cluster serving? (and similarly for TensorBoard)
> > > > 
> > > > 3. for Redis related configations (such as REDIS_MAXMEM), we may need to put it in yaml too.
> > > > ```
> > > 
> > > 
> > > 
> > > 1. should be both cluster-serving-start, will change name
> > 
> > 
> > But why do we need both? Can we keep just one?
> > they are the same, the 3 in `docker` dir are for docker build, (maybe docker build can not go to upper directories, so I just copy them here, for users they can always see only 3, not 6)

I don't think we need start-/stop-/restart-/zoo-cluster-serving.sh files for docker at all; the user should always log into docker container and then config/stat cluster serving in the same way as if they have manually installed Cluster Serving.

Litchilitchy:
> 
> 
> > > > > Several comments:
> > > > > ```
> > > > > 1. Why there are both `cluster-serving-start` and `start-cluster-serving.sh` (and for stop and restart as well)?
> > > > > 
> > > > > 2. How (and when) to shut down Redis? What if Redis is already running when start cluster serving? (and similarly for TensorBoard)
> > > > > 
> > > > > 3. for Redis related configations (such as REDIS_MAXMEM), we may need to put it in yaml too.
> > > > > ```
> > > > 
> > > > 
> > > > 
> > > > 1. should be both cluster-serving-start, will change name
> > > 
> > > 
> > > But why do we need both? Can we keep just one?
> > > they are the same, the 3 in `docker` dir are for docker build, (maybe docker build can not go to upper directories, so I just copy them here, for users they can always see only 3, not 6)
> 
> I don't think we need start-/stop-/restart-/zoo-cluster-serving.sh files for docker at all; the user should always log into docker container and then config/stat cluster serving in the same way as if they have manually installed Cluster Serving.

Can do this, if using pip install, we actually put these scripts files into /usr/bin, and we do not need them when building docker anymore because I put them in pip package.

Will use pip install analytics-zoo in docker and remove these scripts.

PR #1884, opened 2020-01-06
Move Docker section and add Cluster Serving entry

Wesley-Du:
1. Move Docker section from Overview to a standalone folder/page;
2. Add cluster serving programming & API guides entries to document website


jason-dai:
LGTM

PR #1891, opened 2020-01-09
[OpenVINO] Avoid keeping model bytes when running without Spark

qiyuangong:


qiyuangong:
Jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/2930/

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/OpenVINOModel.scala
```diff
@@ -49,24 +49,17 @@ class OpenVINOModel(var modelHolder: OpenVINOModelHolder,
     OpenVINOModel.logger.info("Lazy loading OpenVINO model")
     var nativeRef = -1L
     try {
```
yangw1234: This try block may not be executed when deserialized on executor. 
Maybe
```scala
@transient
private lazy val nativeRef = {

 # loading code
 ...
}
```

PR #1911, opened 2020-01-14
improve index.md formatting

Wesley-Du:


jason-dai:
Why `logo.jpg` is still changed?

PR #1922, opened 2020-01-14
Add apps Auto Test

respecteverything:


apps/run-ray-app-test-pip.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+clear_up () {
```
hkvision: Is it necessary to put ray in a separate file? I think we can just put it with other pip apps?

apps/run-ray-app-test-pip.sh
```diff
@@ -0,0 +1,25 @@
...
+exit_status=$?
+if [ $exit_status -ne 0 ];
+then
+    clear_up
+    echo "ray paramater-server failed"
+    exit $exit_status
+fi
+now=$(date "+%s")
+time1=$((now-start))
+echo "#1 ray paramater-server time used:$time1 seconds"
```
hkvision: new line

apps/run-ray-app-test-pip.sh
respecteverything: Since we still test python 2.7 at Jenkins and Ray cannot run on python 2, I separate it from others. But if we don't need to test py2, I think we can put it together.

hkvision: I will remove 2.7 test.

apps/run-app-tests.sh
```diff
@@ -355,6 +355,52 @@ echo "#8 using_variational_autoencoder_to_generate_faces time used:$time8 second
...
+now=$(date "+%s")
+time14=$((now-start))
+echo "#14 anomaly-detection-hd time used:$time14 seconds"
+
+echo "#15 start app test for pytorch face-generation"
+#timer
+start=$(date "+%s")
+${ANALYTICS_ZOO_HOME}/apps/ipynb2py.sh ${ANALYTICS_ZOO_HOME}/apps/pytorch/face_generation
+sed '/get_ipython()/d' ${ANALYTICS_ZOO_HOME}/apps/pytorch/face_generatio.py
+${SPARK_HOME}/bin/spark-submit \
```
hkvision: use script to launch instead

apps/run-app-tests.sh
```diff
@@ -355,6 +355,54 @@ echo "#8 using_variational_autoencoder_to_generate_faces time used:$time8 second
...
 if [ $RUN_PART3 = 1 ]; then
+echo "#14 start app test for anomaly-detection-hd"
+#timer
+start=$(date "+%s")
+chmod +x $ANALYTICS_ZOO_HOME/bin/data/HiCS/get_HiCS.sh
+$ANALYTICS_ZOO_HOME/bin/data/HiCS/get_HiCS.sh
+${ANALYTICS_ZOO_HOME}/apps/ipynb2py.sh ${ANALYTICS_ZOO_HOME}/apps/anomaly-detection-hd/autoencoder-zoo
+sed -i '/get_ipython()/d' ${ANALYTICS_ZOO_HOME}/apps/anomaly-detection-hd/autoencoder-zoo.py
+sed -i '/plt./d' ${ANALYTICS_ZOO_HOME}/apps/anomaly-detection-hd/autoencoder-zoo.py
+${ANALYTICS_ZOO_HOME}/bin/spark-shell-with-zoo.sh \
```
hkvision: shouldn't it be spark-submit-python-with-zoo.sh?

apps/run-app-tests.sh
```diff
@@ -355,6 +355,54 @@ echo "#8 using_variational_autoencoder_to_generate_faces time used:$time8 second
...
+        --driver-cores 2  \
+        --driver-memory 12g  \
+        --total-executor-cores 2  \
+        --executor-cores 2  \
+        --executor-memory 12g \
+        --py-files ${ANALYTICS_ZOO_PYZIP},${ANALYTICS_ZOO_HOME}/apps/anomaly-detection-hd/autoencoder-zoo.py  \
+        --properties-file ${ANALYTICS_ZOO_CONF} \
+        --jars ${ANALYTICS_ZOO_JAR} \
+        --conf spark.driver.extraClassPath=${ANALYTICS_ZOO_JAR} \
+        --conf spark.executor.extraClassPath=${ANALYTICS_ZOO_JAR} \
```
hkvision: These are not needed if use script.

PR #1925, opened 2020-01-14
Remove python2.7 and spark1.6 in doc

hkvision:
Also some spark 1.6 stuff that I notice.


hkvision:
@glorysdj What about docker default image python version here: https://github.com/intel-analytics/analytics-zoo/blob/master/docs/docs/DockerUserGuide/index.md#analytics-zoo-in-docker?

PR #1929, opened 2020-01-15
[Issue #1896] KerasModel support saving and loading from HDFS

### LINKED ISSUE #1896: tensorflow and tf.keras model (KerasModel) couldn't save to hdfs path when run the job in Yarn-Cluster mode
When use tf.keras APIs and run jobs in Yarn Cluster mode , the TF or Keras model couldn't save artifacts to hdfs path , like this line of code , only local path is acceptable 
model.save_weights("/path")

yangw1234:
If the model/weight path starts with "hdfs", KerasModel will load/save from HDFS.
Fix #1896

### DIFF
--- a/pyzoo/zoo/common/utils.py
+++ b/pyzoo/zoo/common/utils.py
@@ -17,6 +17,11 @@ from bigdl.util.common import Sample as BSample, JTensor as BJTensor,\
     JavaCreator, _get_gateway, _java2py, _py2java
 import numpy as np
 import os
+import tempfile
+import uuid
+import shutil
+
+from urllib.parse import urlparse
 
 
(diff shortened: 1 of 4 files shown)


pyzoo/zoo/common/utils.py
```diff
@@ -47,6 +52,51 @@ def to_list_of_numpy(elements):
     return results
 
 
+def save_file(save_func, path):
```
jason-dai: I think we should just implement the method in scala and then call it from python? there is no point to manually parsing the file address in python.

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2954/ jenkins pass, manually test it for hdfs. S3 support will be addressed on follow up pr.

PR #1932, opened 2020-01-16
Fix ner test for tf 1.15.0

yangw1234:


yangw1234:
python test pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2944/

PR #1969, opened 2020-02-04
Support TensorFlow dataset

yangw1234:
Support tf.data.Dataset in TFDataset.

Usage:
```python
dataset = tfds.load(name="mnist", split=split)
dataset = dataset.map(lambda data: (((tf.to_float(data['image']) - mnist.TRAIN_MEAN)/mnist.TRAIN_STD), data['label']))
tf_dataset = TFDataset.from_tf_data_dataset(train_dataset,
                                         batch_size=BATCH_SIZE)
```

To do:

- [x]  Implement sharding 
- [x]  Verify works if data is stored on hdfs
- [x] Unit Tests

Currently Limitations:
1. The dataset should not be batched
2. User is responsible for putting the data file in every executor or on HDFS


pyzoo/zoo/tfpark/tf_dataset.py
```diff
@@ -627,6 +639,116 @@ def get_num_partitions(self):
...
+        raise NotImplementedError()
+
+    @staticmethod
+    def _assert_not_batched(dataset):
+        from tensorflow.python.data.ops import dataset_ops
+        if isinstance(dataset, dataset_ops.DatasetV1Adapter):
+            TFDataDataset._assert_not_batched(dataset._dataset)
+        elif isinstance(dataset, dataset_ops.BatchDataset):
+            raise ValueError("Dataset should not be batched,"
+                             "please use a dataset with the batch operation")
```
jason-dai: without

pyzoo/zoo/tfpark/tf_dataset.py
```diff
@@ -627,6 +639,116 @@ def get_num_partitions(self):
...
+                 sequential_order=False, shuffle=True):
+        from tensorflow.python.keras.engine import training_utils
+
+        if shuffle:
+            training_utils.verify_dataset_shuffled(tf_data_dataset)
+
+        # we assume that the dataset user passed is not batched,
+        # because we currently did not find a way to get out a
+        # batch of strings in java api
+        TFDataDataset._assert_not_batched(tf_data_dataset)
```
jason-dai: should check if this dataset can be supported the first thing in this method (batching, generator, tf_function, etc.)

pyzoo/zoo/tfpark/tf_dataset.py
```diff
@@ -627,6 +639,116 @@ def get_num_partitions(self):
...
+        if validation_dataset is not None:
+            TFDataDataset._assert_not_batched(validation_dataset)
+
+        node_num, core_num = get_node_and_core_number()
+        self.per_node_batch_size = batch_size // node_num
+
+        tf_data_dataset = tf_data_dataset.batch(self.per_node_batch_size)
```
jason-dai: how about batch_per_thread?

pyzoo/zoo/tfpark/tf_dataset.py
```diff
@@ -627,6 +639,116 @@ def get_num_partitions(self):
...
+        if validation_dataset is not None:
+            validation_dataset = validation_dataset.batch(self.per_node_batch_size)
+
+        assert batch_size % (node_num * core_num) == 0, "batch_size must be a multiple of total number of cores"
```
jason-dai: print the value of node_num and core_num here

pyzoo/zoo/tfpark/tf_dataset.py
```diff
@@ -627,6 +639,116 @@ def get_num_partitions(self):
...
+
+        self.validation_dataset = validation_dataset
+        self.validation_iterator = None
+        self.validation_next_ops = None
+        self.data_count = data_count
+        self.validation_data_count = validation_data_count
+        if validation_dataset is not None:
+            self.validation_iterator = self.validation_dataset.make_initializable_iterator()
+            self.validation_next_ops = nest.flatten(self.validation_iterator.get_next())
+            assert validation_data_count is not None
```
jason-dai: error message

pyzoo/zoo/tfpark/tf_dataset.py
```diff
@@ -627,6 +639,116 @@ def get_num_partitions(self):
...
+
+    def get_prediction_data(self):
+        raise NotImplementedError()
+
+    def get_evaluation_data(self):
+        raise NotImplementedError()
+
+    def get_training_data(self):
+        import tensorflow as tf
+        serialized_graph = bytearray(tf.get_default_graph().as_graph_def().SerializeToString())
```
jason-dai: how do you guarantee tf.get_default_graph() returns the dataset graph?

zoo/src/main/scala/com/intel/analytics/zoo/tfpark/TFDataFeatureSet.scala
```diff
@@ -0,0 +1,248 @@
...
+    val initOp = this.initIteratorOp
+    val names = this.outputNames.toVector
+    val types = this.outputTypes.toVector
+    val shardIdx = this.shardIndex
+
+    graphRunnerRDD.mapPartitionsWithIndex { case (idx, dataIter) =>
+      val graphRunner = dataIter.next()
+      def intiIterator(): Unit =  {
+        graphRunner.runTargets(Vector(initOp),
+          inputs = Vector(Tensor.scalar[Float](idx.toFloat)),
```
jason-dai: why float?

yangw1234: Currently, GraphRunner assumes all numeric values are provided in float form and GraphRunner will transform them into the types required by the TF graph.

pyzoo/zoo/tfpark/tf_dataset.py
yangw1234: Do you mean we should use batch_pre_thread here?
There is only one tf_data_dataset on each partition or node, so I think we should use the batch_per_node instead of batch_per_thread here.

jason-dai: `batch_per_thread` is a parameter in the constructor, yes? I think only one of `batch_size` and `batch_per_thread` can be specified by the user, yes?

yangw1234: Yeah, you are right. Changed that.
Now, `batch_size` must be set for training and `batch_per_thread` must be set for inference.
For training, we create `node_num` partitions and use `batch_size / node_num` for `dataset.batch`; for inference, we create `node_num * core_num` partitions and use `batch_per_thread` for `dataset.batch`

yangw1234:
hi @jason-dai , do you think it is ok to not support `dataset.repeat` and remove the `data_count` argument in constructor?

I think most of the case user does not know the dataset size and they might be more willing to remove the `dataset.repeat` line in their code than to figure out the dataset size.

If user cannot change that line, they can disable the checking and call `dataset.take(dataset_size)` to do this manually.

If this causes user too much trouble, we can always easily add the support and argument back, but if we add it now, it is hard to remove it in the future.

yangw1234:
BigDL DistriOptimizer will always call `val numSamples = dataset.data(train = false).map(_.size()).reduce(_ + _)` to calculate dataset size. This might be a problem for very large datasets. @jason-dai

yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2979/

PR #1971, opened 2020-02-06
Support pre-made estimators

yangw1234:
1. This PR address the following issues:

    a. enable user to define `model_fn` the way the do in tf.estimator.Estimator, except using ZooOptimizer. E.g.
    ```python
    import tensorflow as tf
    from zoo.tfpark import TFEstimator, ZooOptimizer
    def model_fn(features, labels, mode):
            loss, logits = create_model(features, labels)
            train_op = ZooOptimizer(tf.train.AdamOptimizer()).minimize(loss)
            return tf.estimator.EstimatorSpec(mode, train_op=train_op, predictions=logits, loss=loss)
        else:
            return tf.estimator.EstimatorSpec(mode, predictions=logits)

    estimator = TFEstimator.from_model_fn(model_fn, model_dir="/tmp/estimator")
    ```
    b. enable user to use TensorFlow's pre-made Estimators. E.g.
    ```python
    import tensorflow as tf
    linear = tf.estimator.LinearClassifier(feature_columns=feature_columns,
                                           optimizer=ZooOptimizer(tf.train.FtrlOptimizer(0.2)))
    estimator = TFEstimator(linear)
    ```
     c. Multiple optimizers and gradient clipping is naturally support using this implementation. LARS might also be supported.Other GAN models support might be benefit from this.

2. Implementation

     a. Create a ZooOptimizer that wraps a tf.train.Optimizer. When the `compute_gradient` method is called on ZooOptimzier, ZooOptimizer will add an additional `Identity` op with a special name after each gradient tensor, so that we can find those gradient tensor later.

     b. When trained on scala side, what we **previous** do is to:

         (1) assign newly updated weight to TF graph,
               by calling `session.run(assign, feed={var:var_value})`
         (2) get all the gradients,
               by calling `session.run(gradients, feed={input:input_value})`
         (3) use bigdl to sync and update weights
      
      what we do **now** is to:

         (1) use the newly averaged gradients (saved in bigdl weights) to update
              the variables in TF graph, by calling `session.run(train_op,
              feed={gradient: avg_grad_value})`
         (2) get all the gradients in this iteration,
               by calling session.run(gradients, feed={input:input_value})
         (3) use bigdl to sync gradients. But instead of using a bigdl OptimMethod
              to update weights, we use a `FakeOptimMethod` that copy all the averaged
              gradient value to weights, so that we can use the averaged gradient to
              update the variables in TF graph.


yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/2971/

yangw1234:
Further GAN features and some other examples depends on this. I'll merge it first to unblock the development.

PR #1977, opened 2020-02-10
tensorflow and pytorch parallel control

Litchilitchy:
This is a PR created for jenkins test, if passed, will test the performance afterwards.

parallel control of tensorflow and pytorch.

Debugged with @cyita , The predict of 1 model, using `OMP_NUM_THREADS` 8, 16, 32 has nearly linear performance gain, latency is about 1s, 0.55s, 0.30s respectively. But when thread num is 64, the performance is much worse than 32.

Thus temporarily, we set the `par num` to 32, in this mod, the only param we need to control is `par num`, which is the `OMP_NUM_THREADS` **we want to use in maximum** in inference, and any cores over 32 will be assigned with another model, etc.  

e.g. If user say they have 8 cores, serving will use 8 cores, 1 model, if user say they have 96 cores, serving will use 96 cores, with `OMP_NUM_THREADS=par num(32 now)`, 3 models.


jason-dai:
You should separate the object detection support into another PR

PR #1978, opened 2020-02-10
Fix non-trainable param count in KerasNet model.summary

### LINKED ISSUE #1975: model.summary doesn't correctly show nontrainable parameters
If we freeze some layers in a model, the freeze layers are still show as trainable parameters in model.summary.

hkvision:
#1975


dding3:
Can we have a ut which covered KerasParameter with trainable=false

hkvision:
Yeah, there would be some problems with Parameter. But since it is of low priority, I will merge this first  so that freezing embedding would work.
Any further comments? @dding3

hkvision:
jenkins passed: http://10.239.47.211:18888/job/ZOO-PR-Validation/2984/

dding3:
> Yeah, there would be some problems with Parameter. But since it is of low priority, I will merge this first so that freezing embedding would work.
> Any further comments? @dding3

Ok.

hkvision:
I merge this first. Leave the issue open for future fix if we have time.

PR #1981, opened 2020-02-13
Cluster Serving new tf, pytorch, openvino

Litchilitchy:


zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -195,48 +197,93 @@ object ClusterServing {
               })
             })
           })
+
+          // per batch size should be equal to OMP_NUM_THEADS to get max performance
+
+          val perBatchSize = helper.batchSize
+          val upperParNum = (coreNum - 1) / perBatchSize + 1
```
jason-dai: what if perBatchSize is large than coreNum?

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -195,48 +197,93 @@ object ClusterServing {
...
-              }
-
-              (0 until thisBatchSize).toParArray.map(i => {
-                val value = PostProcessing.getInfofromTensor(topN,
-                  result.select(1, i + 1).squeeze())
-                Record(pathByteBatch(i)._1, value)
+            pathBytes.grouped(coreNum).flatMap(batch => {
+              (0 until upperParNum).toParArray.flatMap(cpIndex => {
+                val localModel = bcModel.value
+                val threadTensor = if (chwFlag) {
```
jason-dai: I don't think you should create the `threadTensor` repeatedly here; instead, you may create a `threadTensor[upperParNum]` array as the first thing inside `mapPartitions`

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
Litchilitchy: this design is still in progress. The threads will conflict with each other and the performance gain than old one is limited. (We hope could be about 8x but this is only 2x) I think we need to consider thread binding or strong resource isolation. @qiyuangong @cyita

Litchilitchy:
openvino does not work in this PR

jason-dai:
Separate asnc part into a new PR.

Litchilitchy:
> Separate asnc part into a new PR.

async PR has been separated

this PR is the working PR of our testing now because we have to use both new features. Will merge async PR afterwards and double check this PR after that.

Litchilitchy:
@jason-dai This PR could be suspended because we are using new container solution?

PR #1982, opened 2020-02-13
Cluster Serving object detection support 

Litchilitchy:


zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
```diff
@@ -18,9 +18,32 @@ package com.intel.analytics.zoo.serving.utils
...
+  /**
+   * Image classification post processing
+   * get result from 1-D Tensor of softmax
+   * and create a json string to store the result
+   * @param topN
+   * @param result
+   * @return
+   */
+  def getInfofromTensor(topN: Int, result: Tensor[Float], task: String): String = {
+    if (task == "classification") {
```
jason-dai: Why do we need different task config? We can simply unify the output format for classification and detection.

Litchilitchy: If then we will infer the task type? The object detection has 2d output and classification has 1d

jason-dai: Let's say the output result is a Tensor `T` with `N` rows, each row can be a 1-d or n-d Tensor; you just need to return the `N` rows in JSON format,  each row can simple be a list (with all elements appended senquentially) in JSON

Litchilitchy: Noted.

Litchilitchy: @jason-dai One more question, the classification is 1D and if we give a param `topN` we will do sort and return highest value of them, the object detection is 2D and the prob is given by the 2nd element in each of 1D Tensor, so seems we can not use a regular scheme to return the result if the `topN` is given?

Litchilitchy:
Will create a new PR for new schema

PR #1984, opened 2020-02-14
Cluster Serving s3,hdfs support

Litchilitchy:


zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -354,6 +354,18 @@ class ClusterServingHelper {
    * @param location
    */
   def parseModelType(location: String): Unit = {
+    /**
+     * Download file to local if the scheme is remote
+     * Currently support hdfs, s3
+     */
+    val scheme = location.split(":").head
+    val remoteSchemeSet: Set[String] = Set("hdfs", "s3", "s3n", "s3a")
+    val localModelPath = if (remoteSchemeSet.contains(scheme)) {
```
jason-dai: I think you should check if the fie path is local or not, similar to https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/common/utils.py#L87

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -354,6 +354,18 @@ class ClusterServingHelper {
...
+      FileUtils.copyToLocal(location + "/*", "model/")
+      "model/"
```
jason-dai: shall we use temp instead of the current folder?

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
Litchilitchy: This ask local file to have to scheme `file://` other wise it does not consider it as local? We can do this if needed, shall we?

jason-dai: We can also support other local file path patterns. I think the question is whether it is easier to check for local file pattern, or remote file pattern.

Litchilitchy: I think it is easy to do. The only difference is currently I am checking remote path first using `hdfs, s3...` and this code use `file://` to check local path first. I can not figure out there is basic difference here...

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
Litchilitchy: Noted.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
jason-dai: I think it might be easier to detect local file pattern (as there might be potentially many different remote file systems)? On the other hand, we need to support other local file patterns (such as `/folder` or `./folder`) @yangw1234

yangw1234: I think we have already supported these patterns because I will see a file as local if 1. it has no scheme header or 2. it starts with `file://`. `/folder` and './folder' falls into the first category.

Litchilitchy: What is the condition that it has no scheme from `hdfs, s3, s3a, s3n` and also not local?

jason-dai: The point is that you should check for local file patterns, which is somewhat limited (namely `file://folder`, `/folder`, `./folder`); everything else can be considered remote and can be read using `hadoop.fs`

Litchilitchy: Fixed in #2035

PR #1985, opened 2020-02-14
Fix ray get in related unit tests

hkvision:
The previous code is not running in parallelism, as ray.get is a blocking operation.
Update in this PR in order not to mislead others.


hkvision:
jenkins passed http://10.239.47.211:18888/job/ZOO-PR-Validation/2986/ (doc job failure is irrelevant).

PR #1986, opened 2020-02-14
Improve GANEstimator and example

yangw1234:
1. automatically load existing checkpoint in model_dir
2. change to running gradient update in tf graph
3. change example to conditional GAN, that is output digit by label

image_20100


yangw1234:
jenkins currently down, test it manually, merge it

PR #1988, opened 2020-02-17
Fix TFDataDataset evaluate

yangw1234:


yangw1234:
test it manually, jenkins currently down for because of network issue

PR #1993, opened 2020-02-18
Refactor TFDataset

yangw1234:


yangw1234:
Remove MapDataset
1. not friendly for user to do map on batched tensors
2. does not work on Keras Model
3. we can use tf.data.Dataset for the use case that needs this

yangw1234:
ut passed http://172.168.2.101:8080/job/ZOO-PR-Validation-Spark-2.4/1375/

yangw1234:
example tests pass http://172.168.2.101:8080/job/ZOO-PR-Python-ExampleTests/206/

PR #1999, opened 2020-02-23
Fix tensorflow not init bug in cluster

yangw1234:


yangw1234:
ut passed and manually test examples

PR #2002, opened 2020-02-25
refine Inference Model API & add TFNet API

Le-Zheng:


zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -35,101 +35,123 @@ public AbstractInferenceModel(boolean autoScalingEnabled, int concurrentNum) {
...
     doLoadTF(savedModelDir, inputShape, ifReverseInputChannels, meanValues, scale, input);
   }
 
+  @Deprecated
   public void loadTF(byte[] savedModelBytes, int[] inputShape, boolean ifReverseInputChannels, float[] meanValues, float scale, String input) {
     doLoadTF(savedModelBytes, inputShape, ifReverseInputChannels, meanValues, scale, input);
   }
 
+  @Deprecated
   public void loadTFAsCalibratedOpenVINO(String modelPath, String modelType, String checkpointPath, int[] inputShape, boolean ifReverseInputChannels, float[] meanValues, float scale,
```
qiyuangong: This method should be removed.

Le-Zheng:
Brief description:
1. Method name of "doLoad, doLoadCaffe, doLoadTF, doLoadPyTorch, doLoadOpenVINO" were changed to "loadBigDL, loadCaffe, loadTFFrozen, loadTFSaved, loadPyTorch, loadOpenVINO" in InferenceModel.scala, and make it matched in PythonInferenceModel.scala, AbstractInferenceModel.java
2. Deprecated methods about loading TF as OpenVINO in InferenceModel.scala, PythonInferenceModel.scala and AbstractInferenceModel.java
3.  Add method for loadingTF frozen model with inputs and outputs. For some downloaded frozen model like [Mobilenet_v1](https://github.com/tensorflow/models/blob/master/research/slim/nets/mobilenet_v1.md#pre-trained-models), users can load inputs and ouputs easily, not need to generate a ".json" file putting that mannually.

jason-dai:
1) What methods are user visible? For user visible methods, we need to keep and deprecate them, instead of removing or changing them
2) Instead of `loadTFFrozen` or `loadTFSaved`, maybe better to have `loadTF(String type, ...)` instead?
3) Need to change Python APIs accordingly, which can be in a follow-up PR

Le-Zheng:
Yes, methods about loading TF as OpenVINO are kept and not changed. They are deprecated with `@deprecated("this method will be deprecated", "0.8.0")`.
Ok, the related python API and docs will be updated in another following PR. 
There is a conflict. When using `loadTF` for both frozen and saved model,  new added **frozen model** method`loadTF(modelPath: String, inputs: Array[String], outputs: Array[String])`  will have the same type parameters with existing **saved model** method.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -77,7 +77,7 @@ class InferenceModel(private var autoScalingEnabled: Boolean = true,
    * @param modelPath  the file path of the model
    * @param weightPath the file path of the weights
    */
```
glorysdj: need to add @param for blas，and please check other methods if any @param is missing

glorysdj:
AbstractInferenceModel should have same apis as InferenceModel, so if IM has a method loadBigDL(modelpat, weightpath, blas), AbstractIM should also have this to load blas or non-blas models, maybe keep doLoadXX methods in IM to avoid name conflicts.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -35,101 +35,123 @@ public AbstractInferenceModel(boolean autoScalingEnabled, int concurrentNum) {
...
 
-  public void loadCaffe(String modelPath, String weightPath) {
-    doLoadCaffe(modelPath, weightPath, true);
+  public void loadCaffe(String modelPath, String weightPath) { loadCaffe(modelPath, weightPath, true);
   }
 
-  public void loadTF(String modelPath) {
-    doLoadTF(modelPath);
+  public void loadTFFrozen(String modelPath) {
+    loadTFFrozen(modelPath);
```
glorysdj: this is recursive call, should call doLoadXXX

glorysdj: please also check the uts and make sure all of them are passed

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -463,6 +550,32 @@ class InferenceModel(private var autoScalingEnabled: Boolean = true,
     offerModelQueue()
   }
 
+  private def doLoadTensorflowModel(modelPath: String,
```
qiyuangong: Tensorflow and TF. I think can unify them. :)

jason-dai:
1) For user visible APIs (that is, existing Java APIs in AbstractInferenceModel and Scala APIs in InferenceModel), we should keep and deprecate them as much as we can, instead of removing or changing them. For instance, don't simply rename 'load' to `loadBigDL`; otherwise, you will break existing user code.

2) We may need to remove methods that load TF to OpenVINO once model optimizer is removed @qiyuangong 

3) We may need to change method `loadTF` to `loadTensorFlow` to avoid conflicts; and I think we should specify the type `FrozenModel` and `SavedModel` as a parameter of the method.

qiyuangong:
[Remove Optimization in Zoo-core PR](https://github.com/intel-analytics/analytics-zoo-core/pull/79) is ready. But, merging this Zoo-core PR will make these `Deprecated methods` unavailable. 

I think we can focus on this PR first. :)

jason-dai:
> [Remove Optimization in Zoo-core PR](https://github.com/intel-analytics/analytics-zoo-core/pull/79) is ready. But, merging this Zoo-core PR will make these `Deprecated methods` unavailable.
> 
> I think we can focus on this PR first. :)

We can simply remove the methods that depends on model optimizer in this PR

qiyuangong:
> > [Remove Optimization in Zoo-core PR](https://github.com/intel-analytics/analytics-zoo-core/pull/79) is ready. But, merging this Zoo-core PR will make these `Deprecated methods` unavailable.
> > I think we can focus on this PR first. :)
> 
> We can simply remove the methods that depends on model optimizer in this PR

OK. That's simpler. :)

I will sync with @Le-Zheng. We  also need to modify some examples with such methods.

zoo/src/main/java/com/intel/analytics/zoo/pipeline/inference/AbstractInferenceModel.java
```diff
@@ -51,28 +51,28 @@ public void loadCaffe(String modelPath, String weightPath) {
...
 
-  public void loadTF(String modelPath, int intraOpParallelismThreads, int interOpParallelismThreads, boolean usePerSessionThreads) {
-    doLoadTF(modelPath, intraOpParallelismThreads, interOpParallelismThreads, usePerSessionThreads);
+  public void loadTensorflow(String modelPath, int intraOpParallelismThreads, int interOpParallelismThreads, boolean usePerSessionThreads) {
+    doLoadTensorflow(modelPath, intraOpParallelismThreads, interOpParallelismThreads, usePerSessionThreads);
   }
 
-  public void loadTF(String modelPath, String[] inputs, String[] outputs) {
-    doLoadTF(modelPath, inputs, outputs);
+  public void loadTensorflow(String modelPath, String modelType, String[] inputs, String[] outputs) {
```
qiyuangong: `modelType`  is designed for OpenVINO load TensorFlow.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
jason-dai: I think it's for using MKLBLAS or MKLDNN

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/InferenceModel.scala
```diff
@@ -453,16 +477,111 @@ class InferenceModel(private var autoScalingEnabled: Boolean = true,
   }
 
   private def doLoadTensorflowModel(modelPath: String,
+                                    modelType: String,
+                                    intraOpParallelismThreads: Int,
+                                    interOpParallelismThreads: Int,
+                                    usePerSessionThreads: Boolean): Unit = {
+    modelType match {
+      case null | "" =>
```
jason-dai: why do we want to support null or "" here?

Le-Zheng: `doLoadTensorflow(modelPath，modelType，intraOpParallelismThreads，interOpParallelismThreads，usePerSessionThreads）` is only available for `frozenModel`. @yangw1234 
I think null or " " could be an option for default `frozenModel` .

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/inference/PythonInferenceModel.scala
```diff
@@ -138,4 +144,4 @@ class PythonInferenceModel[T: ClassTag](implicit ev: TensorNumeric[T]) extends P
     val outputActivity = model.doPredict(inputActivity)
     activityToList(outputActivity)
   }
-}
+}
```
qiyuangong: Add an empty line at the end of each file. :)

Le-Zheng:
`doLoadTensorflow` functions keep consistent.
PR validation http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3043/

PR #2003, opened 2020-02-25
change TFPark to from_train_op

yangw1234:
Hi @jason-dai, I still think from_train_op is the more intuitive API for user and it supports multiple optimizers. Maybe we can make `loss` optional if user does not have the access to `loss`.

### DIFF
--- a/docs/docs/ProgrammingGuide/TFPark/tensorflow.md
+++ b/docs/docs/ProgrammingGuide/TFPark/tensorflow.md
@@ -206,26 +206,10 @@ with slim.arg_scope(lenet.lenet_arg_scope()):
      logits, end_points = lenet.lenet(images, num_classes=10, is_training=True)
 
 loss = tf.reduce_mean(tf.losses.sparse_softmax_cross_entropy(logits=logits, labels=squeezed_labels))
-```
 
-You can also construct your model using Keras provided by Tensorflow.
+optimizer = ZooOptimizer(tf.train.AdamOptimizer(1e-3))
 
-```python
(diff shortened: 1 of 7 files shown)


yangw1234:
http://172.168.2.101:8080/job/ZOO-PR-Validation-Spark-2.4/1383/ ut passed

jason-dai:
Two questions:
1) Can we support any `train_op` (assuming the user has used `ZooOptimizer`)?
2) The API still require `loss`; then why is API is more convenient than asking the user to pass `loss` and `optimizer` to TFOptimizer?

docs/docs/ProgrammingGuide/TFPark/tensorflow.md
```diff
@@ -206,26 +205,10 @@ with slim.arg_scope(lenet.lenet_arg_scope()):
...
-from tensorflow.keras.models import Model
-from tensorflow.keras.layers import *
-
-data = Input(shape=[28, 28, 1])
 
-x = Flatten()(data)
-x = Dense(64, activation='relu')(x)
-x = Dense(64, activation='relu')(x)
-predictions = Dense(10, activation='softmax')(x)
+optimizer = ZooOptimizer(tf.train.AdamOptimizer(1e-3))
```
jason-dai: Maybe we can use something like `tfpark.Optim` or `tfpark.OptimMethod` instead of `ZooOptimizer`?

yangw1234:
Hi @jason-dai I'll rethink about this. Close it for now.

PR #2004, opened 2020-02-26
fix init_spark_on_yarn 

jenniew:
Add zoo classpath and bigdl classpath to driver classpath to make it consistent with executor classpath. Avoid potential incompatible class issue between driver and executor.
Manually tested.

### DIFF
--- a/pyzoo/zoo/util/spark.py
+++ b/pyzoo/zoo/util/spark.py
@@ -108,17 +108,17 @@ class SparkRunner():
                             "Please set it manually by python_location")
         return out.strip()
 
-    def _get_bigdl_jar_name_on_driver(self):
+    def _get_bigdl_classpath_jar_name_on_driver(self):
         from bigdl.util.engine import get_bigdl_classpath
         bigdl_classpath = get_bigdl_classpath()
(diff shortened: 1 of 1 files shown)


hkvision:
In what cases will this be an error?
Coz in all our submitting scripts, we never specify `--driver-class-path`, shall we change them all?

jenniew:
When analytics zoo jar and spark jars have same class but different version, there will be incompatible class issue. 
In bin/analytics-zoo-bash.sh, we already do that: 
--conf spark.driver.extraClassPath=${ANALYTICS_ZOO_JAR} \
--conf spark.executor.extraClassPath=${ANALYTICS_ZOO_JAR}

hkvision:
Okay.
Then what about following here: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/util/spark.py#L204 to modify, use spark.driver.extraClassPath same as the script?

jenniew:
Changing the spark conf with this way: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/util/spark.py#L204 is not working in client mode. We need to set driver classpath in submit option.

PR #2008, opened 2020-02-28
update python inference model api and docs

Le-Zheng:


Le-Zheng:
Rename method of loading Analytics Zoo/BigDL model `load` to `loadBigDL`, `doLoadBigDL` and `load_bigdl` for java, scala and python.

PR #2027, opened 2020-03-05
update related docs for inference

Le-Zheng:


Le-Zheng:
Shall we remove part of **2. Load with OpenVINO backend** in this PR?

qiyuangong:
> Shall we remove part of **2. Load with OpenVINO backend** in this PR?

I think we can add a link [Converting a TensorFlow Model into OpenVINO model](https://docs.openvinotoolkit.org/latest/_docs_MO_DG_prepare_model_convert_model_Convert_Model_From_TensorFlow.html) and some comments.

PR #2030, opened 2020-03-06
Cluster Serving output Tensor to ndarray string

Litchilitchy:
This PR outputs the ndarray type of Tensor, e.g. [[1,2],[3,4]]

For object detection, we can use this type of output. But..

We are currently use top-N 1D Tensor filter by default, if using this one by default, we have to modify our quick start example, and add that filter in config file.

Thus I only updated the API but have not given entrance to this method. Maybe when we have more types of output and we can review how to set the entrance here.

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
@@ -62,6 +62,8 @@ object ClusterServing {
     val W = helper.dataShape(1)
     val H = helper.dataShape(2)
 
+    val filter = helper.filter
+
     /**
      * chwFlag is to set image input of CHW or HWC
      * if true, the format is CHW
@@ -170,7 +172,7 @@ object ClusterServing {
                 val localPartitionModel = bcModel.value
                 val result = localPartitionModel.doPredict(tensors.addSingletonDimension()).toTensor
 
-                val value = PostProcessing.getInfofromTensor(topN, result.squeeze())
+                val value = PostProcessing(result)
 
                 Record(path, value)
               })
@@ -227,8 +229,7 @@ object ClusterServing {
               }
 
               (0 until thisBatchSize).toParArray.map(i => {
-                val value = PostProcessing.getInfofromTensor(topN,
-                  result.select(1, i + 1).squeeze())
+                val value = PostProcessing(result.select(1, i + 1))
                 Record(pathByteBatch(i)._1, value)
               })
 
--- a/zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
@@ -16,11 +16,49 @@
 
 package com.intel.analytics.zoo.serving.utils
 
-import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.{Storage, Tensor}
 
-object PostProcessing {
+class PostProcessing(t: Tensor[Float]) {
+  def tensorToNdArrayString(): String = {
+    val sizeArray = t.size()
+    var strideArray = Array[Int]()
+    val totalSize = {
+      var res: Int = 1
+      (0 until sizeArray.length).foreach(i => res *= sizeArray(i))
+      res
+    }
+
+
+    (0 until sizeArray.length).foreach(i => {
+      var res: Int = 1
+      (0 to i).foreach(j => {
+        res *= sizeArray(j)
+      })
+      strideArray = strideArray :+ res
+    })
+    val flatTensor = t.resize(totalSize).toArray()
+    var str: String = ""
+    (0 until flatTensor.length).foreach(i => {
+      (0 until sizeArray.length).foreach(j => {
+        if (i % strideArray(j) == 0) {
+          str += "["
+        }
+      })
+      str += flatTensor(i).toString
+      (0 until sizeArray.length).foreach(j => {
+        if ((i + 1) % strideArray(j) == 0) {
+          str += "]"
+        }
+      })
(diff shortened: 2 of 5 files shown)


jason-dai:
Can this format directly loaded by ndarray? If so, maybe we can just use this format for all the cases. If there is a TopN filer, we can simply add a dimension (i.e., the first dimension is the class id).

Litchilitchy:
> Can this format directly loaded by ndarray? If so, maybe we can just use this format for all the cases. If there is a TopN filer, we can simply add a dimension (i.e., the first dimension is the class id).

Checked, we can use 1 line code to read this `output`. User could
```
import numpy as np
import json
arr = np.array(json.loads(output))
```

And we will use this as universal output, and change Top-N to a 2*n ndarray

Litchilitchy:
Fixed. Can support any type output of models now. @jason-dai @cyita

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -170,7 +172,7 @@ object ClusterServing {
                 val localPartitionModel = bcModel.value
                 val result = localPartitionModel.doPredict(tensors.addSingletonDimension()).toTensor
 
-                val value = PostProcessing.getInfofromTensor(topN, result.squeeze())
+                val value = PostProcessing(result)
```
jason-dai: This will break the existing example and make it slower; I don't think you should just merge the PR.

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -227,8 +229,7 @@ object ClusterServing {
               }
 
               (0 until thisBatchSize).toParArray.map(i => {
-                val value = PostProcessing.getInfofromTensor(topN,
-                  result.select(1, i + 1).squeeze())
+                val value = PostProcessing(result.select(1, i + 1))
```
jason-dai: This will break the existing example and make it slower; I don't think you should merge the PR.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
```diff
@@ -39,5 +77,25 @@ object PostProcessing {
...
+      val tmpTensor = Tensor(Storage(Array(list(i)._1, list(i)._2)))
+      res.select(1, i + 1).copy(tmpTensor)
+    })
+    res
+  }
+}
+object PostProcessing {
+  def apply(t: Tensor[Float], filter: String = null, args: Array[Int] = null): String = {
+    val cls = new PostProcessing(t)
+    val filteredTensor = if (filter == "topN") {
```
jason-dai: Maybe do a pattern match.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
```diff
@@ -39,5 +77,25 @@ object PostProcessing {
     value += "\"}"
     value
   }
-
+  def topN(topN: Int): Tensor[Float] = {
+    val list = TensorUtils.getTopN(topN, t)
+    val res = Tensor[Float](list.size, 2)
+    (0 until list.size).foreach(i => {
+      val tmpTensor = Tensor(Storage(Array(list(i)._1, list(i)._2)))
+      res.select(1, i + 1).copy(tmpTensor)
```
jason-dai: copy is too expensive

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
```diff
@@ -16,11 +16,49 @@
...
+    (0 until sizeArray.length).foreach(i => {
+      var res: Int = 1
+      (0 to i).foreach(j => {
+        res *= sizeArray(j)
+      })
+      strideArray = strideArray :+ res
+    })
+    val flatTensor = t.resize(totalSize).toArray()
+    var str: String = ""
+    (0 until flatTensor.length).foreach(i => {
```
jason-dai: Does this work for tensors with 3 or more dimensions? I think you will need to print recursively

Litchilitchy: it works for any dimension, every time meet any stride will add a brace

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
Litchilitchy: will use setvalue in next PR fix

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
Litchilitchy: The operation are almost the same and TopN is currently the default filter, instead of None. The output are almost the same and the only change will be few line modification to "output -> readable format"

jason-dai: `filter` is default to null in `PostProcessing`, no?

```
object PostProcessing {
  def apply(t: Tensor[Float], filter: String = null, args: Array[Int] = null): String = {
...
```

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
jason-dai: I don't think setvalue works in this case; I think you should directly print from a list of tensors.

Litchilitchy: fixed in #2039

PR #2035, opened 2020-03-09
Cluster Serving s3 check change checking remote to checking local

Litchilitchy:


jason-dai:
Do we have any test for this feature?

Litchilitchy:
> Do we have any test for this feature?

Adding unit test for all the features.

This feature will use hdfs config in hadoop dir, will ask @glorysdj for environment later

PR #2039, opened 2020-03-10
Cluster Serving output change copy to setvalue

Litchilitchy:


zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/PostProcessing.scala
```diff
@@ -81,8 +81,8 @@ class PostProcessing(t: Tensor[Float]) {
     val list = TensorUtils.getTopN(topN, t)
     val res = Tensor[Float](list.size, 2)
     (0 until list.size).foreach(i => {
-      val tmpTensor = Tensor(Storage(Array(list(i)._1, list(i)._2)))
-      res.select(1, i + 1).copy(tmpTensor)
+      res.setValue(i + 1, 1, list(i)._1)
+      res.setValue(i + 1, 2, list(i)._2)
```
jason-dai: 1) So it is assumed that TopN only works for 2d Tensor?
2) Why don't you just do sth like (you also need to deal with "," though)
```scala
var str = "["
for (t <- list) {
   str += toNdArrayString(t)
}
str += "]"
```

Litchilitchy: this also works, loop and recursion could have the same results. This PR is closed, and everything related are moved to #2042

Litchilitchy: @jason-dai There are 2 parts, 1 is filter, specific filter support only specific shape of tensors (this could be extended in the future, like topN params in which dimension to filter). 2 is print, every shape of result  could be printed.

PR #2043, opened 2020-03-10
Cluster Serving Raw streaming supported with multiple mods

Litchilitchy:
I think latency issue is solved.

Modifications mainly including:

1. use streaming instead of structured streaming, but some support are not stable, thus.

2. use accumulator instead of count RDD

3. streaming query in main workload is changed to streaming context

4. to wrap the blas and non-blas, class is added to extend serializable

5. streaming redis reading has some field schema change compared to structured streaming.

6. accumulator is added in every atom op, not iterator, because iterator can be used only once by default, do not want to use it on accumulator.

7. also include some refactoring


Litchilitchy:
A `SSServing.scala` is added to this PR for backup, if everything works well, it could be removed in future

zoo/src/main/scala/com/intel/analytics/zoo/serving/InferenceStrategy.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.zoo.pipeline.inference.InferenceModel
+import org.apache.spark.rdd.RDD
+import com.intel.analytics.zoo.serving.ClusterServing.{Params, Record}
+import org.apache.spark.broadcast.Broadcast
+import org.apache.spark.util.LongAccumulator
+
+
+class InferenceStrategy(params: Params,
```
Litchilitchy: If I do not extend serializable here, error will occur, saying task not serializable...Any one have ideas? @yangw1234 @qiuxin2012 I suppose I only call this method on driver not executors...

Litchilitchy: change it to object it works...

Litchilitchy:
@jason-dai @glorysdj @cyita 

Local test finished, will test stability soon. Code finished if no new bug discovered.

Spark redis jar is not needed anymore

PR #2060, opened 2020-03-11
Add cluster serving singlenode env setup script

cyita:


cyita:
Usage:
sudo bash cluster-serving-singlenode-env.sh --worker_num=3 --cpuset=0-35 --container-prefix test_

If worker_num is unset, a suitable worker_num will be calculated (16~31 cores/worker). 
If worker_num == 0, master and worker containers won't be started. And local mode is suggested to run cluster-serving.

If CPUSET is unset, all cores on the machine will be used.

scripts/cluster-serving/cluster-serving-singlenode-env.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
+#set -x
+#set -e
+ 
+SPARK_MASTER_PORT=7077
+SPARK_MASTER_WEBUI_PORT=8080
+REDIS_PORT=6379
```
jason-dai: How do you make sure `REDIS_PORT` is the same as the one set in config.yaml?

scripts/cluster-serving/cluster-serving-singlenode-env.sh
```diff
@@ -0,0 +1,153 @@
...
+	if [ $CORE_START -eq $TOTAL_CORE_NUM ]; then
+		CORE_START=$((TOTAL_CORE_NUM - 1))
+	fi	
+fi
+TOTAL_CORE_NUM=$((CORE_END - CORE_START + 1))
+
+if [ -z $WORKER_NUM ]; then
+	echo "Calculating worker num..."
+	# 16 ~ 31 core/worker
+	WORKER_NUM=$((TOTAL_CORE_NUM / 16))
```
jason-dai: What if TOTAL_CORE_NUM is 44 or 56? A lot of cores can be wasted.

cyita: If TOTAL_CORE_NUM=44
WORKER_NUM=2
WORKER_CORE_NUM=44/2=22
so each worker will be assigned 22 cores.

scripts/cluster-serving/cluster-serving-singlenode-env.sh
```diff
@@ -0,0 +1,153 @@
...
+then
+    echo "Port $REDIS_PORT is in use. Please check redis port when running cluster-serving"
+#    exit 1
+fi
+
+host_ip=`ifconfig | grep -Eo 'inet (addr:)?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*' | grep -v '127.0.0.1'|grep -v '172.17.0.1'`
+
+if [ $WORKER_NUM -ne 0 ]; then
+  echo "Start a master node..."
+  docker run -itd --name "$PREFIX"cluster_master --net=host -e mode=master --add-host `hostname`:$host_ip intelanalytics/docker-spark:v1
```
jason-dai: Is this standard Spark docker image?

scripts/cluster-serving/cluster-serving-singlenode-env.sh
```diff
@@ -0,0 +1,153 @@
...
+      TEMP_CORE_END=$((CORE_START + WORKER_CORE_NUM - 1))
+    fi
+    echo "Start worker $i, cpu-set: $CORE_START - $TEMP_CORE_END"
+    docker run -itd --name "$PREFIX"cluster_worker$i --net=host -e SPARK_MASTER=spark://`hostname`:$SPARK_MASTER_PORT -e mode=worker --add-host `hostname`:$host_ip --cpuset-cpus $CORE_START-$TEMP_CORE_END intelanalytics/docker-spark:v1
+    CORE_START=$((TEMP_CORE_END + 1))
+  done
+fi
+
+echo "Start cluster serving..."
+docker run -itd --name "$PREFIX"cluster-serving --net=host -e OMP_NUM_THREADS=$WORKER_CORE_NUM -e KMP_AFFINITY=granularity=fine,compact,1,0 -e KMP_BLOCKTIME=20 intelanalytics/zoo-cluster-serving:0.7.0
```
jason-dai: 1) Shall we run cluster serving inside master? 
2) Can we automatically start cluster serving?
3) Do we need set env variables like OMP_NUM_THREADS for cluster serving?

scripts/cluster-serving/cluster-serving-singlenode-env.sh
```diff
@@ -0,0 +1,153 @@
...
+
+if [ $WORKER_NUM -ne 0 ]; then
+  echo "Master: spark://`hostname`:$SPARK_MASTER_PORT"
+  echo "Data Src: $host_ip:Redis Port(Default 6379)"
```
jason-dai: It seems that the configure redis address is not used in cluster-serving-start.sh @Litchilitchy?

scripts/cluster-serving/cluster-serving-singlenode-env.sh
cyita: Redis will be started in the cluster-serving container when running cluster-serving-start, and port information will be printed.

If the default REDIS_PORT is in use, a warning message will print.
Users can use any redis_port in config.yaml, as long as there is a Redis server.

scripts/cluster-serving/cluster-serving-singlenode-env.sh
cyita: 1. In this case, if the cluster-serving encounters some problems, the master node will down.
2. May need some changes on cluster serving Dockerfile @Litchilitchy 
3. The default value is all, and I will sync with @Litchilitchy

scripts/cluster-serving/cluster-serving-singlenode-env.sh
cyita: No, I will change it to bde2020/spark-master(worker):2.4.3-hadoop2.7

scripts/cluster-serving/cluster-serving-singlenode-env.sh
jason-dai: Then why do we need to check REDIS_PORT here?

scripts/cluster-serving/cluster-serving-singlenode-env.sh
jason-dai: No need to change to bde2020; on the other hand, can we build cluster-serving docker based on docker-spark?

jason-dai:
> Usage:
> sudo bash cluster-serving-singlenode-env.sh --worker_num=3 --cpuset=0-35 --container-prefix test_
> 
> If worker_num is unset, a suitable worker_num will be calculated (16~31 cores/worker).
> If worker_num == 0, master and worker containers won't be started. And local mode is suggested to run cluster-serving.
> 
> If CPUSET is unset, all cores on the machine will be used.

1) We should focus on the default behavior (i.e., no parameters are provided by the user)
2) Ideally we should provide the script for user to directly run cluster serving on single node in _performance mode_
3) Please also update the document (programming guide for cluster serving)

scripts/cluster-serving/cluster-serving-singlenode-env.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
+#set -x
+#set -e
+ 
+SPARK_MASTER_PORT=7077
+SPARK_MASTER_WEBUI_PORT=8080
+REDIS_PORT=6379
+WORKER_NUM=""
+TOTAL_CORE_NUM=`nproc`
```
jason-dai: We may need to deal with both hyper-threading enabled and disabled scenarios

scripts/cluster-serving/cluster-serving-singlenode-env.sh
Litchilitchy: hard coded to 6379

scripts/cluster-serving/cluster-serving-singlenode-env.sh
Litchilitchy: 1. I think master node down is not an issue, because this spark is prepared for serving.
2. Currently users log into the container to start according to our 0.7 dockerfile. So do we need to change it, or start serving for them in this specific case?
3. I think no need to set, we have code to send value to executors.

scripts/cluster-serving/cluster-serving-singlenode-env.sh
```diff
@@ -0,0 +1,153 @@
...
+  WORKER_CORE_NUM=$((TOTAL_CORE_NUM / WORKER_NUM))
+fi
+
+echo "worker_num: $WORKER_NUM"
+echo "core_start: $CORE_START"
+echo "core_end: $CORE_END"
+
+echo "Check port..."
+if [ $WORKER_NUM -ne 0 ]; then
+  grep_port=`netstat -tlpn | grep "\b$SPARK_MASTER_PORT\b"`
```
qiuxin2012: this may not work as PID's value may be the same with `SPARK_MASTER_PORT`.  Please cut the line `Local Address`

PR #2064, opened 2020-03-15
[WIP] telco solution on analytics zoo

shane-huang:
initial design of the API and submodules. 
TODO:
1. refactor some automl code
     - LSTMSeq2seq._decode_sequence() decouple build inference model and inference to allow wrapping model with tfpark.
     - pipeline.fit_with_fixed_configs(), decouple bulid pipeline & fit 
2. decide whether we should use automl with fixed configurations for models. This way we need to wrap tfpark models inside of automl.


jason-dai:
For model, let's focus on forecast fist, an leave anomaly to the next version

jason-dai:
For forecast, we can fist focus on `LSTMForecaster` and `MTNetForecaster`, which extends `Forecaster` which in turn extends `KerasModel`

shane-huang:
accidentally messed up with original branch. close this one and add a new PR: https://github.com/intel-analytics/analytics-zoo/pull/2081

PR #2073, opened 2020-03-17
fix failed pr validation

yushan111:


yushan111:
fixed automl conftest import issue in pr validation.

PR #2080, opened 2020-03-18
Perfmode with numactl

cyita:
Add performance mode with numactl.

Cluster serving micro-batch summary log lost after merge master.


scripts/cluster-serving/stop-cluster-with-numactl.sh
```diff
@@ -0,0 +1,67 @@
...
+  fi
+}
+
+TOTAL_CORE_NUM=`nproc`
+if ht_enabled; then
+  TOTAL_CORE_NUM=$((TOTAL_CORE_NUM / 2))
+fi
+
+_WORKER_NUM=$1
+if [ $TOTAL_CORE_NUM -lt 24 ] && [ -z "${_WORKER_NUM}" ]; then
```
qiyuangong: Why do you choose `24` as a threshold ?

cyita: Sync with WORKER_NUM calculation.
There are 24 cores on a numa node on ap, and the performance of 4 workers is not satisfactory.
This value may change according to Yabai's test result.

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
...
+    true
+  fi
+}
+
+TOTAL_CORE_NUM=`nproc`
+if ht_enabled; then
+  TOTAL_CORE_NUM=$((TOTAL_CORE_NUM / 2))
+fi
+
+_WORKER_NUM=$1
```
qiyuangong: Give a default value, e.g., 1. Then check if $1 exist. If not exist, use default value.

scripts/cluster-serving/stop-cluster-with-numactl.sh
qiyuangong: Yes. It makes sense for AP, but not a good setting for single socket server, e.g., 6240 single socket only have 18 cores. I think you can remove this check. WORKER_NUM and core number is enough.

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
...
+if [ $TOTAL_CORE_NUM -lt 24 ] && [ -z "${_WORKER_NUM}" ]; then
+  # use local mode
+  echo "local[*]"
+  exit 1
+fi
+
+if [ -z "${SPARK_HOME}" ]; then
+  export SPARK_HOME="$(cd "`dirname "$0"`"/..; pwd)"
```
qiyuangong: I think we can info our user. SPARK_HOME is not set. Just like

https://github.com/intel-analytics/analytics-zoo/blob/master/scripts/spark-submit-scala-with-zoo.sh#L10

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: WORKER_NUM should be calculated according to core num on each numa node.

scripts/cluster-serving/stop-cluster-with-numactl.sh
cyita: This _WORKER_NUM is just for the developer to test performance.

qiyuangong: OK. I got it.

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: ok

scripts/cluster-serving/start-cluster-with-numactl.sh
qiyuangong: OK.

jason-dai:
1) The "performance mode" will simply ignore user specified parameters such as master or driver_memory; I think we should only allow it when local[*] is specified by the user
2) Should should combine the start- and stop-cluster scripts into one file, so as to eliminate code duplication
3) With HT on, each worker should use half of the cores (say 12) but bind to all of them (e.g., 0~23)? Maybe we can add HT support in a follow-up PR as more testing maybe needed.
4) We first check if we can run "performance mode" in the beginning of the start script; and how do we know that we are in "performance mode" when running the stop script?

scripts/cluster-serving/cluster-serving-start
```diff
@@ -98,5 +98,22 @@ if [ -z "${SPARK_REDIS_JAR}" ]; then
     SPARK_REDIS_JAR=spark-redis-2.4.0-jar-with-dependencies.jar
 fi
 
+if [ "${params_performance_mode}" == "ON" ]; then
```
jason-dai: you should check if master is local[*]

scripts/cluster-serving/cluster-serving-stop
```diff
@@ -1,8 +1,32 @@
...
+      }
+   }'
+}
+
 if [ -f "running" ]; then
     rm running
     echo "Stop signal sent to serving."
+    parse_yaml config.yaml
+    eval $(parse_yaml config.yaml)
+    if [ "${params_performance_mode}" == "ON" ]; then
```
jason-dai: how do you make sure it is actually running in "performance mode"?

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+
```
jason-dai: I think you should first check if we can actually run the "performance mode" at the beginning of the script

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
...
+  fi
+}
+
+TOTAL_CORE_NUM=`nproc`
+if ht_enabled; then
+  TOTAL_CORE_NUM=$((TOTAL_CORE_NUM / 2))
+fi
+
+_WORKER_NUM=$1
+if [ $TOTAL_CORE_NUM -lt 24 ] && [ -z "${_WORKER_NUM}" ]; then
```
jason-dai: Define a variable with value 24 (rather than directly using 24)

cyita:
> 1. The "performance mode" will simply ignore user specified parameters such as master or driver_memory; I think we should only allow it when local[*] is specified by the user

Yes, the master specified by the user will be replaced by the output of start-cluster script and the "local[*]" check is needed.

But I'm not sure if we should replace driver_memory or executor memory specified by the user.

> 2. We should combine the start- and stop-cluster scripts into one file, so as to eliminate code duplication

Do you mean start and stop should be two functions of one file?

> 3. With HT on, each worker should use half of the cores (say 12) but bind to all of them (e.g., 0~23), yes? Maybe we can add HT support in a follow-up PR as more testing may be needed.

In the original script, if HT on, they bind the first half of the cores in a numa node, and I will check it with Qiyuan.

> 4. We should first check if we can run "performance mode" in the beginning of the start script; and how do we know that we are in "performance mode" when running the stop script?

Ok, I will move the numactl check to the beginning of the script.
Stop script uses the same logic as the start script. And if the cluster was not set up before we run stop, the stop script cannot stop any master or worker with the given name.

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
...
+  # use local mode
+  echo "local[*]"
+  exit 1
+fi
+
+if [ -z "${SPARK_HOME}" ]; then
```
jason-dai: It seems that we don not check ${SPARK_HOME} in cluster-serving-start? @Litchilitchy Maybe we should also check it there.

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
...
+  shift
+
+  if [ "$SPARK_WORKER_PORT" = "" ]; then
+    PORT_FLAG=
+    PORT_NUM=
+  else
+    PORT_FLAG="--port"
+    PORT_NUM=$(( $SPARK_WORKER_PORT + $WORKER_NUM - 1 ))
+  fi
+  WEBUI_PORT=$(( $SPARK_WORKER_WEBUI_PORT + $WORKER_NUM - 1 ))
```
jason-dai: what if PORT_NUM and WEBUI_PORT are not available?

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
...
+        _WORKER_NUM=$((_LENGTH / 12))
+      fi
+      if [[ $_WORKER_NUM -eq 0 ]]; then
+        _WORKER_NUM=1
+      fi
+      _LENGTH=$((_LENGTH / _WORKER_NUM))
+
+      for ((i = 0; i < $((_WORKER_NUM)); i ++)); do
+        core_start=$(( i * _LENGTH ))
+        _NUMACTL="numactl -m ${_NUMA_NO} -C $(join_by , ${_NUMA_CPUS[@]:${core_start}:${_LENGTH}})"
```
jason-dai: for HT, you should use half of the core (e.g., 12), but bind to 2x cores (e.g., 0~23)?

scripts/cluster-serving/start-cluster-with-numactl.sh
```diff
@@ -0,0 +1,135 @@
...
+  # Load NUMA configurations line-by-line and set `numactl` options
+  for nnode in ${_NUMA_HARDWARE_INFO[@]}; do
+    if [[ ${nnode} =~ ^node\ ([0-9]+)\ cpus:\ (.+)$ ]]; then
+      _NUMA_NO=${BASH_REMATCH[1]}
+      IFS=' ' _NUMA_CPUS=(${BASH_REMATCH[2]})
+      _LENGTH=${#_NUMA_CPUS[@]}
+      if ht_enabled; then _LENGTH=$((_LENGTH / 2)); fi
+      if [[ -z "${_WORKER_NUM}" ]]; then
+        # calculate worker num on this numa node, 12 ~ 23 core/worker
+        _WORKER_NUM=$((_LENGTH / 12))
```
jason-dai: Define a variable of 12 (rather than using 12 directly)

scripts/cluster-serving/cluster-serving-stop
cyita: Stop script uses the same logic as the start script. And if the cluster was not set up before we run stop, the stop script cannot stop any master or worker with the given name.

cyita: Or we should create a file like "running" when starting performance mode?

scripts/cluster-serving/stop-cluster-with-numactl.sh
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
+
+#   Stops all slaves on this worker machine
```
jason-dai: combine this file with stat-cluster

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: fixed

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: Done

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: Done

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: The spark start script will automatically try other ports to start the worker.

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: Done

scripts/cluster-serving/start-cluster-with-numactl.sh
cyita: Will discuss with @qiyuangong

scripts/cluster-serving/stop-cluster-with-numactl.sh
cyita: Done

scripts/cluster-serving/cluster-with-numactl.sh
```diff
@@ -0,0 +1,160 @@
...
+  shift
+  echo "$*"
+}
+
+# Compute memory size for each NUMA node
+IFS=$'\n'; _NUMA_HARDWARE_INFO=(`numactl --hardware`)
+_NUMA_NODE_NUM=`echo ${_NUMA_HARDWARE_INFO[0]} | sed -e "s/^available: \([0-9]*\) nodes .*$/\1/"`
+_TOTAL_MEM=`grep MemTotal /proc/meminfo | awk '{print $2}'`
+# Memory size of each NUMA node = (Total memory size - 1g) / Num of NUMA nodes
+_NUMA_MEM=$((((_TOTAL_MEM - 1048576) / 1048576) / $_NUMA_NODE_NUM))
```
jason-dai: 1) Define a variable for 1048576
2) Shall we set aside some some memory for driver?

cyita: 1. Fixed
2. We will reserve 2g for the driver in standalone mode

scripts/cluster-serving/cluster-with-numactl.sh
```diff
@@ -120,7 +120,8 @@ _NUMA_NODE_NUM=`echo ${_NUMA_HARDWARE_INFO[0]} | sed -e "s/^available: \([0-9]*\
 _TOTAL_MEM=`grep MemTotal /proc/meminfo | awk '{print $2}'`
 # Memory size of each NUMA node = (Total memory size - 1g) / Num of NUMA nodes
 _1G=1048576
-_NUMA_MEM=$((((_TOTAL_MEM - _1G) / _1G) / $_NUMA_NODE_NUM))
+_MEMORY_FOR_DRIVER=2  # reserve 2g memory for the driver
+_NUMA_MEM=$((((_TOTAL_MEM - _1G - (_1G * _MEMORY_FOR_DRIVER)) / _1G) / $_NUMA_NODE_NUM))
```
jason-dai: make sure (_NUMA_MEM / _WORKER_NUM) is larger than 0

scripts/cluster-serving/cluster-with-numactl.sh
```diff
@@ -0,0 +1,165 @@
...
+fi
+
+if ht_enabled; then
+  echo "failed,Please turn off Hyperthreading to activate PERFORMANCE_MODE"
+  exit 1
+fi
+
+_WORKER_CORE_NUM_LESS_THAN=24
+_MODE=$1 # start or stop
+_WORKER_NUM=$2  # worker num on each numa node
```
jason-dai: _WORKER_PER_SOCKET?

scripts/cluster-serving/cluster-with-numactl.sh
```diff
@@ -0,0 +1,172 @@
...
+
+    for ((i = 0; i < $((_WORKER_PER_SOCKET)); i ++)); do
+      if [ "${_MODE}" == "start" ]; then
+        core_start=$(( i * _LENGTH ))
+        _NUMACTL="numactl -m ${_NUMA_NO} -C $(join_by , ${_NUMA_CPUS[@]:${core_start}:${_LENGTH}})"
+        echo ${_NUMACTL}
+
+        # Launch a worker with numactl
+        export SPARK_WORKER_CORES=${_LENGTH} # core num per worker
+        export SPARK_WORKER_MEMORY="$((_NUMA_MEM / _WORKER_PER_SOCKET))g"
```
jason-dai: (_NUMA_MEM / _WORKER_PER_SOCKET) can still potentially be 0?

cyita: If _NUMA_MEM <= 0, script will exit with err msg (Line 122-125)
If _WORKER_PER_SOCKET > _NUMA_MEM, _WORKER_PER_SOCKET=_NUMA_MEM (Line 146-148). And _WORKER_PER_SOCKET >= 1. So SPARK_WORKER_MEMORY will not be 0.

scripts/cluster-serving/cluster-with-numactl.sh
cyita: fixed

jason-dai:
LGTM

PR #2085, opened 2020-03-19
add lstm network traffic use case

yushan111:


shane-huang:
looks good as a basic version. Other features and functionalities can be added in following verisons.

jason-dai:
Should use new  forecaster in `zouwu/model/` instead of `VanillaLSTM`?

PR #2100, opened 2020-03-23
Deprecated TFNet training

yangw1234:
1. Deprecate the training function in TFNet. Users should use TFPark for TF training
2. Remove nnframes SimpleTraining.py


yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3097/

PR #2101, opened 2020-03-23
Add automl docs

yushan111:


shane-huang:
Enough for an initial version. Will update the details later.

PR #2113, opened 2020-03-27
add autots use case

yushan111:
main changes:

- add auto ts use case

- change the name of model use case from "network_traffic_forecasting" to "network_traffic_model_forecasting"

-  fix bug of fill missing data in model use case


shane-huang:
The flow generally looks good to me. Add a README in network_traffic folder describing scenario and the files.

shane-huang:
Got a little bit concern that only 4 trials won't be enough to showcase the capability of autots.. .

pyzoo/zoo/zouwu/use-case/network_traffic/network_traffic_autots_forecasting.ipynb
```diff
@@ -0,0 +1,1768 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Network Traffic Forecasting with _autoTS_"
```
jason-dai: AutoTS

pyzoo/zoo/zouwu/use-case/network_traffic/network_traffic_autots_forecasting.ipynb
```diff
@@ -0,0 +1,1768 @@
...
+    "The general workflow using automated training contains below two steps. \n",
+    "   1. create a ```AutoTSTrainer``` to train a ```TSPipeline```, save it to file to use later or elsewhere if you wish.\n",
+    "   2. use ```TSPipeline``` to do prediction, evaluation, and incremental fitting as well. \n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, you need to initialize RayOnSpark before using auto training (i.e. ```AutoTSTrainer```), and stop it after training finished. Using TSPipeline only does not need RayOnSpark. \n"
```
jason-dai: (Note RayOnSpark is not needed if you just use TSPipeline for inference).

pyzoo/zoo/zouwu/use-case/network_traffic/network_traffic_autots_forecasting.ipynb
```diff
@@ -0,0 +1,1768 @@
...
+      "CPU times: user 14.6 s, sys: 4.26 s, total: 18.9 s\n",
+      "Wall time: 8min 35s\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%time\n",
+    "from zoo.zouwu.autots.forecast import GridRandomRecipe\n",
+    "ts_pipeline = trainer.fit(train_df, val_df, \n",
+    "                          recipe=GridRandomRecipe(look_back=look_back),\n",
```
jason-dai: need to briefly explain GridRandomRecipe

shane-huang: Actually we have a doc in github.io. Maybe we can add a link to the docs?

pyzoo/zoo/zouwu/use-case/network_traffic/network_traffic_autots_forecasting.ipynb
yushan111: @jason-dai  `TSPipeline` can be also used for incremental fitting. How about we change to "(Note RayOnSpark is not needed if you just use TSPipeline)" ?

jason-dai: Note RayOnSpark is not needed if you just use TSPipeline for inference, evaluation or incremental training

shane-huang:
Will change the interface to allow both simple arguments and Recipe.  Merge this PR for now and open a new one to do it.

PR #2114, opened 2020-03-27
Fix bugs for autots

yushan111:


shane-huang:
Does this patch passed pr validaiton?

yushan111:
> Does this patch passed pr validaiton?

yes it has passed.

PR #2116, opened 2020-03-27
Fix scala style and perf on macOS

qiyuangong:


qiyuangong:
Jenkins passed in most jobs except a stacked one. http://10.239.47.210:18888/job/ZOO-PR-Validation/3104/

PR #2120, opened 2020-03-30
Move common logic to datastore

yangw1234:
## Current Problem:
There is common pattern in the implementations of FeatureSet (e.g. TFDataFeatureSet, JEP based FeatureSet): 
1. create and cache a RDD that each of its partition resides on each executor.
2. based on that RDD, using mapPartitions create some stateful object(s) that can output data records as training/evaluation data.

Without a well-defined common interface and repeating the same logic in several places is a burden on maintaining the code and will waste time implementing it again if similar requirements come up in the future.

## Solution:
This PR extracts the common logic in creating and caching the RDD (that each of its partition resides on each executor) into one FeatureSet called `DataStoreFeatureSet`. And it defines an interface, `DataStore` that will maintains the per-partition state and handles the reading and outputting training record.

With this implementation, future development can only focus on how to read read data, maintain state and output training record in each partition by providing a `DataStore` object.

## Implementation

1. Create of DataStoreFeatureSet
 
`DataStoreFeatureSet` will automatically create a RDD with each of its partition in each executor, by providing the preferred locations of each recording.

Users of `DataStoreFeatureSet` should provide a DataStore, which is a serializable object that manage the data reading in each partition.

```scala
abstract class DataStore[T: ClassTag] extends java.io.Serializable {

  def init(upstreamData: Iterator[Any], context: DataStorePartitionContext): Unit

  def makeIterators(train: Boolean): Iterator[T]

  def shuffle(): Unit

  def size(): Int
}
```
2. refactor TFDataFeatureSet to use DataStoreFeatureSet


jason-dai:
Please elaborate what goal you want to achieve here.

yangw1234:
> Please elaborate what goal you want to achieve here.

Hi @jason-dai thanks for pointing out, I described the goal in the PR description.

PR #2133, opened 2020-04-01
[Bug Fix] Fix TFPark control inputs

yangw1234:
In TFPark TFOptimizer, if users use a placehoder to control different behavior between training and evaluation, it does not pass the tenor correctly.

This PR fixes this problem.


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3123/  jenkins pass, the error is irrelevant.

PR #2135, opened 2020-04-02
fix nnframes example inputting params

Le-Zheng:


Le-Zheng:
pr validation http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3152/

PR #2136, opened 2020-04-02
Disable OpenVINO unite test

qiyuangong:


jason-dai:
Why should it be disabled?

qiyuangong:
Network is still unstable. Pulling files from S3 may fail and cause jenkins build fail.

PR #2151, opened 2020-04-06
[Fix issue #2150] Remove waiting in Ray init

yangw1234:
## Current Problem
Currently, `RayContext.init()` will wait at least 13 seconds before start. The current flow is
1. if `task_id == 0`, first sleep 5 seconds, then start ray master, then sleep 8 seconds.
2. if `task_id != 0`, first sleep 8 seconds, then start ray slaves, then sleep 5 seconds.

So, there is a 3 seconds gap to make sure ray slaves is started after ray master is started. 
This 3 seconds gay might be too long for a small cluster and might not be enough for a large cluster. And there are also 10 seconds sleep time does not seem to be useful.

## Proposed Solution
We can put a barrier between master start and slave start, e.g.
```python
def map_func(iter):
    if task_id == 0:
        start_ray_master()
    tc.barrier()
    if task_id != 0:
        start_ray_slave()
```

## Testing
1. existing unit tests
2. manually tested it on yarn cluster

### DIFF
--- a/pyzoo/zoo/ray/util/raycontext.py
+++ b/pyzoo/zoo/ray/util/raycontext.py
@@ -80,7 +80,7 @@ class RayServiceFuncGenerator(object):
         return modified_env
 
     def __init__(self, python_loc, redis_port, ray_node_cpu_cores,
-                 password, object_store_memory, waitting_time_sec=6, verbose=False, env=None,
+                 password, object_store_memory, verbose=False, env=None,
                  extra_params=None):
         """object_store_memory: integer in bytes"""
         self.env = env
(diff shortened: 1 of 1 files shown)


shane-huang:
I think barrier should be okay for this.  And maybe we should add some intergation tests (on cluster) for this test.

hkvision:
LGTM

pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -136,7 +135,7 @@ def _get_raylet_command(redis_address,
                                                        object_store_memory=object_store_memory,
                                                        extra_params=extra_params)
 
-    def _start_ray_node(self, command, tag, wait_before=5, wait_after=5):
+    def _start_ray_node(self, command, tag, wait_before=0, wait_after=0):
```
jason-dai, line 138: just remove wait_before and wait_after?

yangw1234: removed

jason-dai:
LGTM

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3166/ jenkins passes, except one failure caused by network issue.

PR #2164, opened 2020-04-08
fix scala example finish on k8s

Le-Zheng:


Le-Zheng:
http://10.239.47.210:18888/job/ZOO-PR-Validation/3158/

PR #2175, opened 2020-04-10
Refactor RayContext

### LINKED ISSUE #2154: Ray tune applications may get stuck while using RayOnSpark on single node
This affects both zouwu and automl fitting with ray tune. 
While running more than one trials on local, there may be one trial whose status is always `RUNNING` without updating of pid, running time or iteration information. Resource utilizations are low. Killing all ray processes and restart doesn't help. This issue happens in a random fashion. 
image

Note that this only occurs while init with `RayOnSpark`.  `ray.init()` does not trigger this issue.

hkvision:
Fix #2154

1. Use ray.init() and ray.shutdown() directly for spark local. Spark only restricts cpu resources in this case,
2. Some code clean and refactor. Previously the code is too messy. For example, labels and extra-params overlap, API not clear, etc.

### DIFF
--- a/pyzoo/zoo/ray/util/raycontext.py
+++ b/pyzoo/zoo/ray/util/raycontext.py
@@ -16,7 +16,6 @@
 
 import os
 import re
-import time
 import signal
 import random
 import multiprocessing
@@ -25,17 +24,17 @@ from pyspark import BarrierTaskContext
 
 from zoo.ray.util import is_local
 from zoo.ray.util.process import session_execute, ProcessMonitor
-from zoo.ray.util.utils import resourceToBytes
+from zoo.ray.util.utils import resource_to_bytes
 import ray.services as rservices
 
 
-class JVMGuard():
+class JVMGuard:
     """
     The registered pids would be put into the killing list of Spark Executor.
     """
 
     @staticmethod
-    def registerPids(pids):
+    def register_pids(pids):
         import traceback
         try:
             from zoo.common.utils import callZooFunc
@@ -94,6 +93,7 @@ class RayServiceFuncGenerator(object):
         self.verbose = verbose
         # _mxnet_worker and _mxnet_server are resource tags for distributed MXNet training only
         # in order to diff worker from server.
+        # This is useful to allocate workers and servers in the cluster.
         # Leave some reserved custom resources free to avoid unknown crash due to resources.
         self.labels = \
             """--resources='{"_mxnet_worker": %s, "_mxnet_server": %s, "_reserved": %s}' """ \
@@ -132,9 +132,9 @@ class RayServiceFuncGenerator(object):
                             ray_exec,
                             password,
                             ray_node_cpu_cores,
-                            labels,
-                            object_store_memory,
-                            extra_params):
+                            labels="",
+                            object_store_memory=None,
+                            extra_params=None):
         command = "{} start --redis-address {} --redis-password  {} --num-cpus {} {}  ".format(
             ray_exec, redis_address, password, ray_node_cpu_cores, labels)
         return RayServiceFuncGenerator._enrich_command(command=command,
@@ -145,7 +145,7 @@ class RayServiceFuncGenerator(object):
         modified_env = self._prepare_env()
         print("Starting {} by running: {}".format(tag, command))
         process_info = session_execute(command=command, env=modified_env, tag=tag)
-        JVMGuard.registerPids(process_info.pids)
+        JVMGuard.register_pids(process_info.pids)
         process_info.node_ip = rservices.get_node_ip_address()
         return process_info
 
@@ -190,8 +190,7 @@ class RayServiceFuncGenerator(object):
 
 class RayContext(object):
     def __init__(self, sc, redis_port=None, password="123456", object_store_memory=None,
-                 verbose=False, env=None, local_ray_node_num=2,
-                 extra_params=None):
+                 verbose=False, env=None, extra_params=None):
         """
         The RayContext would init a ray cluster on top of the configuration of SparkContext.
         For spark cluster mode: The number of raylets is equal to number of executors.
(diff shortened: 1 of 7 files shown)


pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -267,80 +257,32 @@ def stop(self):
             return
         import ray
         ray.shutdown()
-        if not self.ray_processesMonitor:
-            print("Please start the runner first before closing it")
-        else:
-            self.ray_processesMonitor.clean_fn()
-        self.stopped = True
-
-    def purge(self):
```
jason-dai, line 276: why do you remove `purge`? just do nothing if it's local.

hkvision: Emmm... it's not used anywhere...
OK. I will add it back.

hkvision: Reverted.

hkvision:
Tested manually on yarn clusters for lenet and recommendation examples.

shane-huang:
Need change some of the unittests now to test this. 
For now you can just change this line
https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/test/zoo/zouwu/test_auto_ts.py#L69
from
```
num_rand_samples=1
```
to 
```
num_rand_samples=5
```
Please rebase from master and get my latest changes in unit tests in automl and zouwu.

yushan111:
Kai has also changed one of the automl unit test to multi-trials. 
Specifically, he has changed `num_rand_samples` from 1 to 5 in [ `test_fit_LSTMGridRandomRecipe`](https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/test/zoo/automl/regression/test_time_sequence_predictor.py#L66)

yushan111:
Below tests have passed based on code of this PR.

-  automl and zouwu UTs (including some multi-trials UTs) tested in local

- automl and zouwu app tests in local

- autots notebook (multi-trials) in distributed.

pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -190,8 +190,7 @@ def _start_ray_services(iter):
 
 class RayContext(object):
     def __init__(self, sc, redis_port=None, password="123456", object_store_memory=None,
-                 verbose=False, env=None, local_ray_node_num=2,
```
shane-huang, line 193: We're removing `local_ray_node_num`? What is this parameter? If remove from interface, how do we set this parameter in future?

shane-huang: We should also update RayOnSpark docs if usage or interface changes.

jason-dai: > We're removing `local_ray_node_num`? What is this parameter? If remove from interface, how do we set this parameter in future?

Why do you want to set the parameter? It'll be meaningless after the change.

pyzoo/zoo/ray/util/raycontext.py
```diff
@@ -204,52 +203,44 @@ def __init__(self, sc, redis_port=None, password="123456", object_store_memory=N
...
+        self.object_store_memory = resource_to_bytes(object_store_memory)
+        self.ray_processesMonitor = None
+        self.env = env
+        self.extra_params = extra_params
+        # For Spark local mode, directly call ray.init() and ray.shutdown().
+        # ray.shutdown() would clear up all the ray related processes.
+        # Ray Manager is only needed for Spark cluster mode to monitor ray processes.
+        if not self.is_local:
+            self.num_ray_nodes = int(self.sc.getConf().get("spark.executor.instances"))
+            self.ray_node_cpu_cores = int(self.sc.getConf().get("spark.executor.cores"))
```
jason-dai, line 224: you will still need to set `num_ray_nodes` and `ray_node_cpu_cores` when it is local; it is used in the xShard PR https://github.com/intel-analytics/analytics-zoo/pull/2115 @jenniew

hkvision:
http://10.239.47.210:18888/job/ZOO-PR-Validation/3180/ passed.

pyzoo/zoo/ray/util/raycontext.py
hkvision: This is never used and the API wouldn't change.

pyzoo/zoo/ray/util/raycontext.py
hkvision: For local, can set cores through init_spark_on_local. This change won't affect the local usage.

hkvision: See here: https://github.com/intel-analytics/analytics-zoo/pull/2175/files#diff-bf2c4f9e7d5d7fdbfd5e2a357739e4c9R298

pyzoo/zoo/ray/util/raycontext.py
hkvision: Even before the change, it doesn't make sense as well to start 2 ray nodes locally.

pyzoo/zoo/ray/util/raycontext.py
hkvision: num_ray_nodes is not needed for ray local.

jason-dai: it is used to compute the parallelism by @jenniew at https://github.com/intel-analytics/analytics-zoo/blob/1aa8cecd40e4aa7e3ecde69f32019d60ba563475/pyzoo/zoo/xshard/pandas/preprocessing.py#L110

pyzoo/zoo/ray/util/raycontext.py
shane-huang: Well, actually I don't know what this parameter does in the old interface - the name looks weird. I just want to make sure changing interface doesn't break anything.

pyzoo/zoo/ray/util/raycontext.py
hkvision: So what's num_ray_nodes in local?
I think her logic `num_executors = context.num_ray_nodes` only applies for yarn but not for local.
She should check `if context is local, then num_partitions=local cores; otherwise = num executors * executor cores`

hkvision: Or in spark local I can set `num_ray_nodes = 1` and `ray_node_cpu_cores=local cores` to keep her API aligned, but I don't think it is correct actually.

hkvision:
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3182/ passed
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3183/ passed

pyzoo/zoo/ray/util/raycontext.py
jason-dai: 1 seems 

> Or in spark local I can set `num_ray_nodes = 1` and `ray_node_cpu_cores=local cores` to keep her API aligned, but I don't think it is correct actually.

This seems reasonable

hkvision: Updated.

jason-dai:
LGTM

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3197/

hkvision:
http://10.239.47.210:18888/job/ZOO-PR-Validation/3198/

Issue #2204, opened 2020-04-14
org/tensorflow/util/SaverDefOrBuilder when pip test python examples


glorysdj:
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Pip-Example-Integration/93/consoleFull

py4j.protocol.Py4JJavaError: An error occurred while calling o47.createTFNetFromSavedModel.
: java.lang.NoClassDefFoundError: org/tensorflow/util/SaverDefOrBuilder
	at com.intel.analytics.zoo.pipeline.api.net.TFNetForInference$.fromSavedModel(TFNetForInference.scala:232)
	at com.intel.analytics.zoo.pipeline.api.net.python.PythonZooNet.createTFNetFromSavedModel(PythonZooNet.scala:106)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:357)
	at py4j.Gateway.invoke(Gateway.java:282)
	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
	at py4j.commands.CallCommand.execute(CallCommand.java:79)
	at py4j.GatewayConnection.run(GatewayConnection.java:238)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ClassNotFoundException: org.tensorflow.util.SaverDefOrBuilder
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:338)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 13 more

Clearing up environment. Uninstalling analytics-zoo
Found existing installation: analytics-zoo 0.8.0.dev0
Uninstalling analytics-zoo-0.8.0.dev0:
  Successfully uninstalled analytics-zoo-0.8.0.dev0
Found existing installation: BigDL 0.10.0
[...]

PR #2207, opened 2020-04-14
pip install deleted tensorflow dependency tmp fix

Litchilitchy:
@jason-dai as @yangw1234 meets error in #2204 

Cluster Serving will switch to flink in few days and new summary would be support, so I think we could use this old version, and write new summary support for cluster serving.


Litchilitchy:
delete part of jar...could test if it works

yangw1234:
manually tested

yangw1234:
I'll first merge it, as we need to publish another pip dev version to test this.

PR #2224, opened 2020-04-16
Add torchmodel and torchloss

qiuxin2012:
Add torchmodel and torchloss, run pytorch model in an embedding Cpython interpreter.(Powered by [jep](https://github.com/ninia/jep))

User could use `TorchModel.from_pytorch` to create a TorchModel, `TorchLoss.from_pytorch` to create a TorchLoss.

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala
@@ -51,7 +51,7 @@ class TorchLoss(private val criterionHolder: Array[Byte])
   }
 
   override def updateGradInput(input: Activity, target: Activity): Activity = {
-    //TODO: return a empty result
+    // TODO: return a empty result
     Tensor[Float]()
   }
 
--- a/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala
@@ -20,6 +20,7 @@ import java.util.UUID
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractCriterion, Activity}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.zoo.common.PythonInterpreter
+import jep.NDArray
 
 
 class TorchLoss(private val criterionHolder: Array[Byte])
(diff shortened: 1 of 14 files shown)


qiuxin2012:
If numpy's version is low, like 1.16.5, will get some error like this.
``` jep.JepException: jep.JepException: <class 'TypeError'>: Buffer format 'f' is not valid for a byte[].```

File: zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+    if (dataExisted) {
+      PythonInterpreter.exec("target = data[1]")
+    } else {
+      // TODO: support table target
+      require(target.isTensor, "only support tensor target")
+      // TODO: detect type
+      val t = target.toTensor[Float]
+      PythonInterpreter.set("nd_target",
+        new NDArray[Array[Float]](t.storage().array(), t.size(): _*))
+      PythonInterpreter.exec("target = torch.Tensor(nd_target).long()")
```
qiuxin2012, line 54: Seems we need user to provide this `long()` type.

jason-dai: the target has to be long?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala
```diff
@@ -0,0 +1,77 @@
...
+  extends AbstractCriterion[Activity, Activity, Float]() {
+  import TorchLoss._
+
+  protected lazy val loaded = {
+    PythonInterpreter.set("criterion_bytes", criterionHolder)
+    val loadModelCode =
+      s"""
+         |from pyspark.serializers import CloudPickleSerializer
+         |c_by = bytes(b % 256 for b in criterion_bytes)
+         |${name} = CloudPickleSerializer.loads(CloudPickleSerializer, c_by)
```
jason-dai, line 36: `${getName()}` is used in `TorchModel`?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchModel.scala
```diff
@@ -0,0 +1,207 @@
...
+         |
+         |from pyspark.serializers import CloudPickleSerializer
+         |by = bytes(b % 256 for b in model_bytes)
+         |${getName()} = CloudPickleSerializer.loads(CloudPickleSerializer, by)
+         |""".stripMargin
+    PythonInterpreter.exec(loadModelCode)
+    true
+  }
+
+  val weights: Tensor[Float] = Tensor[Float](Storage[Float](init_weights))
```
jason-dai, line 56: instead of passing `init_weights` from Python, which can be very large, can we get the weights from the loaded model?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala
qiuxin2012: Yes, torch's CrossEntropy require the target is long tensor.  Maybe you need user to tell us the target's type.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchLoss.scala
qiuxin2012: We don't have getName method in Criterion, only Model has name.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/net/TorchModel.scala
```diff
@@ -0,0 +1,207 @@
...
+  val gradients: Tensor[Float] = Tensor[Float](weights.size())
+
+  override def parameters(): (Array[Tensor[Float]], Array[Tensor[Float]]) = {
+    (Array(weights), Array(gradients))
+  }
+
+  val setWeightCode =
+    s"""
+        |w = torch.Tensor(newWeight)
```
qiuxin2012, line 65: Maybe asTensor

qiuxin2012:
jenkins passed

PR #2241, opened 2020-04-22
Cluster serving Flink

Litchilitchy:


zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkRedisSource.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  }
+
+  override def run(sourceContext: SourceFunction.SourceContext[List[(String,String)]]): Unit = while (isRunning){
+    val response = db.xreadGroup(
+      "serving",
+      "cli",
+      512,
+      50,
+      false,
```
glorysdj, line 57: i think we should make these parameters config-able, not hard coded.

zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkRedisSource.scala
```diff
@@ -0,0 +1,80 @@
...
+      new SimpleEntry("image_stream", StreamEntryID.UNRECEIVED_ENTRY))
```
glorysdj, line 58: image_stream should also can be congiured, especially when we need to support multi-models/multi-versions.

zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkRedisSource.scala
```diff
@@ -0,0 +1,80 @@
...
+import com.intel.analytics.zoo.serving.ClusterServing.params
+import com.intel.analytics.zoo.serving.utils.{FileUtils, SerParams}
+
+import scala.collection.JavaConversions._
+
+class FlinkRedisSource(params: SerParams) extends RichSourceFunction[List[(String, String)]] {
+  @volatile var isRunning = true
+
+  var redisPool: JedisPool = null
+  var db: Jedis = null
```
glorysdj, line 35: name db is confusing, actually it's a jedis connection.

zoo/src/main/scala/com/intel/analytics/zoo/serving/pipeline/RedisPool.scala
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.intel.analytics.zoo.serving.pipeline
+
+import redis.clients.jedis.{Jedis, JedisPool}
+
+object RedisPool {
```
glorysdj, line 21: this code never used?

PR #2263, opened 2020-04-26
remove some useless dependency in torch_net.py

qiuxin2012:
remove some useless tf dependency in torch_net.py


pyzoo/zoo/pipeline/api/net/__init__.py
```diff
@@ -16,4 +16,3 @@
 
 from .graph_net import GraphNet
 from .net_load import Net
-from zoo.tfpark.tfnet import TFNet
```
yangw1234, line 19: please make sure you have passed jenkins before merging it.

PR #2272, opened 2020-04-27
Temporarily drop support for val_split

yangw1234:
1. it does not work now
2. it makes more sense to support it in TFDataset if needed


yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3261/

PR #2276, opened 2020-04-27
Upgrade automl for ray 0.8.4

yushan111:


pyzoo/zoo/automl/config/recipe.py
```diff
@@ -67,12 +69,12 @@ def __init__(self):
...
+            "selected_features": json.dumps(all_available_features),
             "model": "LSTM",
-            "lstm_1_units": RandomSample(lambda spec: np.random.choice([32, 64], size=1)[0]),
-            "dropout_1": RandomSample(lambda spec: np.random.uniform(0.2, 0.5)),
-            "lstm_2_units": RandomSample(lambda spec: np.random.choice([32, 64], size=1)[0]),
-            "dropout_2": RandomSample(lambda spec: np.random.uniform(0.2, 0.5)),
+            "lstm_1_units": tune.choice([32, 64]),
+            "dropout_1": tune.uniform(0.2, 0.5),
+            "lstm_2_units": tune.choice([32, 64]),
+            "dropout_2": tune.uniform(0.2, 0.5),
```
shane-huang, line 77: RandomSample is designed to isolate tune config. You can leave it these way now. Later we may need to refactor and isolate all tune related calls (e.g.config, trials, and save/load) into tune backend.

pyzoo/zoo/automl/config/recipe.py
```diff
@@ -248,27 +249,26 @@ def __init__(
...
             # -------- feature related parameters
-            "selected_features": RandomSample(
-                lambda spec: np.random.choice(
-                    all_available_features,
-                    size=np.random.randint(
-                        low=3, high=len(all_available_features), size=1),
-                    replace=False)),
+            "selected_features": tune.sample_from(lambda spec:
+                                                  json.dumps(
+                                                      list(np.random.choice(
```
shane-huang, line 266: tune.choice can not be used here?

pyzoo/zoo/automl/model/time_sequence.py
```diff
@@ -68,6 +68,7 @@ def fit_eval(self, x, y, validation_data=None, mc=False, verbose=0, **config):
         :param validation_data: tuple in format (x_test,y_test), data used for validation.
         If this is specified, validation result will be the optimization target for automl.
         Otherwise, train metric will be the optimization target.
+        :param metric:
```
shane-huang, line 71: add api doc for metric

pyzoo/zoo/automl/regression/time_sequence_predictor.py
```diff
@@ -273,19 +284,25 @@ def _print_config(self, best_config):
...
+    def _make_pipeline(self, analysis, metric_mode, feature_transformers, model, remote_dir):
+        metric = "reward_metric"
+        best_config = analysis.get_best_config(metric=metric, mode=metric_mode)
+        best_logdir = analysis.get_best_logdir(metric=metric, mode=metric_mode)
+        print("best log dir is ", best_logdir)
+        dataframe = analysis.dataframe(metric=metric, mode=metric_mode)
+        # print(dataframe)
+        model_path = os.path.join(best_logdir, dataframe["checkpoint"].iloc[0])
+        config = convert_bayes_configs(best_config).copy()
         self._print_config(config)
```
shane-huang, line 296: I think we don't need _print_config here. There're too many logs currently if run many trials were run.

pyzoo/zoo/automl/search/RayTuneSearchEngine.py
```diff
@@ -295,7 +240,7 @@ def train_func(config, tune_reporter):
             global_input_df = ray.get(input_df_id)
             trial_input_df = deepcopy(global_input_df)
             config = convert_bayes_configs(config).copy()
-            # print("config is ", config)
+            print("config is ", config)
```
shane-huang, line 243: remove this print. We don't need to print config in every trial.

yushan111: Sorry, this is just for debug and I forgot to remove it.

pyzoo/zoo/automl/regression/time_sequence_predictor.py
yushan111: This _print_config is in make_pipeline and is for printing out the best config we used in the result pipeline. Should we still remove it?

pyzoo/zoo/automl/config/recipe.py
yushan111: yes, ray tune 0.8.4 has introduced tensorboard for log. Input a list of string will cause an error in tensorboard (should be a bug of tensorboard) therefore a list of string can not use `tune.choice` directly. However input one string works fine so I encoded the features after `random.choice` into one string with `json` and then passed into the lambda function of `sample_from`. That works.

pyzoo/zoo/automl/config/recipe.py
yushan111: Sure, I will change `RandomSample` back

yushan111: This remains unchanged and will refactor later, corresponding to issue #2289

pyzoo/zoo/automl/search/RayTuneSearchEngine.py
yushan111: @shane-huang  removed in PR #2288

pyzoo/zoo/automl/model/time_sequence.py
yushan111: @shane-huang  added in PR #2288

PR #2292, opened 2020-04-29
update partition size in xshard on spark

dding3:


dding3:
jekins failed. http://10.239.47.210:18888/job/ZOO-PR-Validation/3291/  But the exception urllib.error.URLError: <urlopen error [Errno 113] No route to host> is not related to this PR, merge the pr first.

PR #2312, opened 2020-05-08
Apptest

Le-Zheng:


Le-Zheng:
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Pip-Apps-Integration-Part3/26/ passed

PR #2317, opened 2020-05-11
Orca Pandas partition_by support on Spark

jenniew:
Implement partition_by(cols, num_partitions=None) on Orca Pandas datashard on Spark.
Related issue: https://github.com/analytics-zoo/orca/issues/15


pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -222,3 +224,42 @@ def apply(self, func, *args):
 
     def get_data(self):
         return self.data
+
+
+class SparkPandasDataShards(SparkDataShards):
+    def partition_by(self, cols, num_partitions=None):
```
jason-dai, line 230: I don't think a `SparkPandasDataShards` is needed, as there is no guarantee that after transformation (i.e., `SparkPandasDataShards.tranform_shard`) it will still be a Pandas based shard.

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -222,3 +224,42 @@ def apply(self, func, *args):
...
+        # if partition by a column
+        if isinstance(cols, str):
+            # change data to key value pairs
+            rdd = self.rdd.flatMap(
+                lambda df: df.apply(lambda row: (row[cols], row.values.tolist()), axis=1)
```
jason-dai, line 235: what if `row[cols]` is not valid?

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -222,3 +224,42 @@ def apply(self, func, *args):
...
+                .values.tolist())
+            # partition with key
+            partitioned_rdd = rdd.partitionBy(num_partitions)
+        # partition by column list
+        elif isinstance(cols, list):
```
jason-dai, line 240: only 2-element column if supported below? maybe we can just support a single column for now?

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -222,3 +224,42 @@ def apply(self, func, *args):
...
+class SparkPandasDataShards(SparkDataShards):
+    def partition_by(self, cols, num_partitions=None):
+        # if partition by a column
+        if isinstance(cols, str):
+            # change data to key value pairs
+            rdd = self.rdd.flatMap(
+                lambda df: df.apply(lambda row: (row[cols], row.values.tolist()), axis=1)
+                .values.tolist())
+            # partition with key
+            partitioned_rdd = rdd.partitionBy(num_partitions)
```
jason-dai, line 238: By default `num_partitions` should be the same as the current partition number of the DataShards, yes?

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -222,3 +224,42 @@ def apply(self, func, *args):
...
+                lambda df: df.apply(
+                    lambda row:
+                    (reduce(lambda col1, col2: str(row[col1])+ "_" + str(row[col2]), cols),
+                     row.values.tolist()), axis=1).values.tolist())
+            # partition with key
+            partitioned_rdd = rdd.partitionBy(num_partitions)
+        else:
+            raise Exception("only column name or list of column name are support")
+
+        columns = self.rdd.first().columns
```
jason-dai, line 252: what if `.columns` is not valid?

pyzoo/zoo/orca/data/pandas/preprocessing.py
jenniew: pandas would throw exception if cols are invalid. I will add column check.

pyzoo/zoo/orca/data/pandas/preprocessing.py
jenniew: yes.

pyzoo/zoo/orca/data/pandas/preprocessing.py
jenniew: OK. I will update it and add pandas check.

pyzoo/zoo/orca/data/pandas/preprocessing.py
jenniew: will add value check

pyzoo/zoo/orca/data/pandas/preprocessing.py
jason-dai: but the default value is not used in the implementation?

jenniew: if num_partition is None, rdd.partitionBy will use self._defaultReducePartitions as partition number.

pyzoo/zoo/orca/data/pandas/preprocessing.py
jenniew: what is 2-element column?

jenniew: updated

pyzoo/zoo/orca/data/pandas/preprocessing.py
jenniew: updated

pyzoo/zoo/orca/data/pandas/preprocessing.py
jenniew: updated

pyzoo/zoo/orca/data/pandas/preprocessing.py
jason-dai: but why don't we specify current partition number if `num_partitions` is None?

jenniew: OK. I'll change it.

jenniew: Updated.

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -222,3 +224,42 @@ def apply(self, func, *args):
 
     def get_data(self):
         return self.data
+
+
+class SparkPandasDataShards(SparkDataShards):
```
jason-dai: Do we still need this class? I think we just use `SparkDataShards`, no?

jenniew: it's removed.

jenniew: please check the latest version.

jason-dai:
LGTM

pyzoo/zoo/orca/data/shard.py
```diff
@@ -114,3 +114,39 @@ def collect(self):
     def repartition(self, num_partitions):
         self.rdd = self.rdd.repartition(num_partitions)
         return self
+
+    def partition_by(self, cols, num_partitions=None):
+        import pandas as pd
+        data = self.rdd.first()
```
jason-dai, line 120: I think we don't need get the whole dataframe; just one row in the dataframe, yes?

jenniew: The element in SparkDataShards may not be Pandas Dataframe, it can be numpy or other type. If it is not df, cannnot get a row of that.

jason-dai: what if the dataframe is very large? maybe you can check at remote and then get the columns back?

jenniew: Updated. check the type and return column from remote.

jenniew:
jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3378/

PR #2322, opened 2020-05-12
Update cluster serving Dockerfile & scripts

cyita:


jason-dai:
Why do we need to put all of these files into docker image? Should some of them go to https://github.com/intel-analytics/analytics-zoo/tree/master/pyzoo/test/?

PR #2323, opened 2020-05-12
Orca unique() on DataShards on Spark

jenniew:
Implementation of unique() on DataShards of Pandas Series and Index.
Also implement select operation with [col] because we need to do unique() on df[col].
May add support for DataShards of numpy later. 
Related issue: https://github.com/analytics-zoo/orca/issues/16


jason-dai:
There are two issues that need to be figured out:
1. For Spark based implementation, we need to automatically manage data cache
   ```
   shards = read_csv(..) #need to be automatically cached
   loc_df = shards[location] 
   location = loc_df.unique()
   shards.transform_shard(...) #need to cache the new shards, and delete old ones
   ```
2. For Ray based implementation, we need to manage the backing memory space 
   ```
   shards = read_csv(..) #df stored in Actor
   loc_df = shards[location] #additional df stored in Actor - can consume a lot of memory
                             #only store index - loc_df and shards share the same data
   location = loc_df.unique()
   shards.transform_shard(...) #update to shards also changes loc_df
   ```

jenniew:
Is there any possibility that user do not do aggregation on loc_df,  or they may do loc_df.transform_shard()?

jenniew:
Updated to support shards.unique('location'). Currently only support select column name to do unique operation in DataFrame.

pyzoo/zoo/orca/data/shard.py
```diff
@@ -151,3 +151,23 @@ def merge(iterator):
...
                             " of Pandas DataFrame")
+
+    def unique(self, key):
+        import pandas as pd
+        elem_class, columns = self.rdd.map(
+            lambda data: (type(data), data.columns) if isinstance(data, pd.DataFrame)
+            else (type(data), None)).first()
+        if issubclass(elem_class, pd.DataFrame):
+            import numpy as np
+            if key is None:
```
jason-dai, line 162: why can't key be None?

pyzoo/zoo/orca/data/shard.py
```diff
@@ -151,3 +151,23 @@ def merge(iterator):
...
+                raise Exception("Cannot apply unique operation on Datashards of Pandas Dataframe"
+                                " without column name")
+            if key in columns:
+                rdd = self.rdd.map(lambda df: df[key].unique().tolist())
+                result = rdd.reduce(lambda list1, list2: np.array(list(set(list1 + list2))))
```
jason-dai, line 167: need to perform unique operation in reduce, no?

pyzoo/zoo/orca/data/shard.py
jenniew: if key is None, we may want to apply unique() on DataFrame. But DataFrame has no unique() operation. unique() works on 1-d array like

pyzoo/zoo/orca/data/shard.py
jenniew: set(list1+list2) is to do unique. Or we can also run pd.unique(list1 + list2) in reduce, too.

jenniew: change to pd.unique(list1 + list2) which is faster.

jason-dai:
Is this PR ready for merge?

jenniew:
> Is this PR ready for merge?

Jenkins server always gets some problems in recently days. Will check with Dongjie to see if I can merge.

PR #2329, opened 2020-05-13
Add open_text and open_image in orca

hkvision:
https://github.com/analytics-zoo/orca/issues/8
https://github.com/analytics-zoo/orca/issues/9

Can Popen to execute a cat command to read hdfs text files. Alternatively, https://pypi.org/project/pydoop/ can be used as well.

### DIFF
--- a/pyzoo/zoo/orca/data/utils.py
+++ b/pyzoo/zoo/orca/data/utils.py
@@ -78,12 +78,13 @@ def get_class_name(obj):
     return module + '.' + obj.__class__.__name__
 
 
-def open_file(path):
+def open_text(path):
+    # Return a list of lines
     if path.startswith("hdfs"):  # hdfs://url:port/file_path
         import pyarrow as pa
         fs = pa.hdfs.connect()
         with fs.open(path, 'rb') as f:
-            return f.read().decode("utf-8").split("\n")
+            lines = f.read().decode("utf-8").split("\n")
     elif path.startswith("s3"):  # s3://bucket/file_path
         access_key_id = os.environ["AWS_ACCESS_KEY_ID"]
         secret_access_key = os.environ["AWS_SECRET_ACCESS_KEY"]
@@ -95,9 +96,12 @@ def open_file(path):
         bucket = path_parts.pop(0)
         key = "/".join(path_parts)
         data = s3_client.get_object(Bucket=bucket, Key=key)
-        return data["Body"].read().decode("utf-8").split("\n")
+        lines = data["Body"].read().decode("utf-8").split("\n")
     else:  # Local path
-        return open(path)
+        lines = []
+        for line in open(path):
+            lines.append(line)
+    return [line.strip() for line in lines]
 
 
 def open_image(path):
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/orca/data/utils.py
```diff
@@ -85,3 +85,24 @@ def extract_one_path(file_path, file_type, context):
                           for file in os.listdir(file_path)
                           if os.path.splitext(file)[1] == "." + file_type]
     return file_paths
+
+
+def open_file(path):
+    if path.startswith("hdfs"):  # hdfs://url:port/file_path
+        from subprocess import Popen, PIPE
+        return Popen(["hadoop", "fs", "-cat", path], stdout=PIPE, encoding="utf-8").stdout
+        # Can use pydoop as well.
```
jason-dai, line 94: can we use `pyarrow.HdfsFile.download`?

hkvision: I changed to pyarrow. Take a look? @jason-dai 
Read text file as bytes, decode as string and split by \n; return a list of strings. For HyperSeg code, just need to change import and function name, and no other changes needed.
Built-in `open` function returns a TextIOWrapper, don't know how to have exactly the same return type for hdfs and s3...

hkvision:
@jason-dai Finished the two utils. Take a look. Would add some unit tests afterwards.

pyzoo/zoo/orca/data/utils.py
```diff
@@ -76,3 +76,50 @@ def get_class_name(obj):
...
         return obj.__class__.__name__
     return module + '.' + obj.__class__.__name__
+
+
+def open_file(path):
+    if path.startswith("hdfs"):  # hdfs://url:port/file_path
+        import pyarrow as pa
+        fs = pa.hdfs.connect()
+        with fs.open(path, 'rb') as f:
+            return f.read().decode("utf-8").split("\n")
```
jason-dai, line 86: this only works for text?

pyzoo/zoo/orca/data/utils.py
```diff
@@ -76,3 +76,50 @@ def get_class_name(obj):
...
+
+
+def open_image(path):
+    from PIL import Image
+    if path.startswith("hdfs"):  # hdfs://url:port/file_path
+        import pyarrow as pa
+        from io import BytesIO
+        fs = pa.hdfs.connect()
+        with fs.open(path, 'rb') as f:
+            return Image.open(BytesIO(f.read()))
```
jason-dai, line 110: does `Image.open(f)` work? `f` is supposed to be a File object? https://arrow.apache.org/docs/python/generated/pyarrow.HdfsFile.html

hkvision: Seems Image.open doesn't support HdfsFile...

pyzoo/zoo/orca/data/utils.py
hkvision: Yeah. Any other file formats we need to support?

jason-dai: then we may change the method to `open_text`?

hkvision: okay!

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3405/

hkvision:
tested hdfs and s3 manually

PR #2330, opened 2020-05-13
Add orca tensorflow estimator

yangw1234:
Usage:
```python
    file_path = os.path.join(resource_path, "orca/learn/ncf.csv")
    data_shard = zoo.orca.data.pandas.read_csv(file_path, sc)

    def transform(df):
        result = {
            "x": [df['user'].to_numpy(), df['item'].to_numpy()],
            "y": [df['label']]
        }
        return result

    data_shard.transform_shard(transform)

    est = Estimator.from_pre_built_graph(
        inputs=[user, item],
        labels=[label],
        outputs=[logits],
        loss=loss,
        optimizer=SGD(),
        metrics={"loss": loss})
    est.fit(data_shard=data_shard,
            batch_size=8,
            steps=10,
            validation_data_shard=data_shard)
```


jason-dai:
Several comments:
1) For `Estimator.fit(xshards)`, we may just assume `xshards` is an XShards of dictionary:
```
{
"x": ndarray or list of ndarray
"y": ndarray or list of ndarray
}
``` 
We can add support for `tf.data.Dataset` later.

2) I think we can focus on low level TensorFlow graph first; we can support tf.keras and tf.estmator APIs later.

3) Shall we ask the user to pass `loss` and `optimizer`, or `loss` and `train_op` for the graph??

yangw1234:
@jason-dai 
1. already done
2. already removed model_fn part
3. I think in this stage we first let users pass `loss` and `optimizer` (bigdl's) similar in tfpark?

jason-dai:
> 3. I think in this stage we first let users pass `loss` and `optimizer` (bigdl's) similar in tfpark?

Why it's OK for this PR, we should support native Tensorflow optimizer in a follow-up PR.

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+
+class Estimator(object):
+    def fit(self, data, steps, **kwargs):
+        pass
+
+    def predict(self, data, **kwargs):
+        pass
+
+    @staticmethod
+    def from_pre_built_graph(*, inputs, outputs,
```
jason-dai, line 34: just `from_graph`?

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+                             labels, loss, optimizer,
```
jason-dai, line 35: Only `inputs` is always required. For training, `outputs` can be `None`; and for prediction, `labels`, `loss` and `optimizer` can be None.

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+                                  outputs=outputs,
+                                  labels=labels,
+                                  loss=loss,
+                                  optimizer=optimizer,
+                                  metrics=metrics, updates=updates,
+                                  sess=sess,
+                                  model_dir=model_dir)
+
+
+def _data_shard_to_tf_dataset(data_shard,
```
jason-dai, line 53: maybe add `TFDataset.from_xshards` method?

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+                              batch_size=-1, batch_per_thread=-1,
+                              validation_data_shard=None):
+    # todo data_shard.head ?
+    first_data = data_shard.rdd.first()
```
jason-dai, line 57: `first_data` can be potentially large. We can dynamically fetch `feature_spec` and `label_spec`, similar to https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/data/shard.py#L156

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+                                 features=feature_spec,
+                                 labels=label_spec,
+                                 batch_size=batch_size,
+                                 batch_per_thread=batch_per_thread,
+                                 val_rdd=val_rdd)
+
+    return dataset
+
+
+def _rdd_to_data_shard(rdd):
```
jason-dai, line 100: where is it used?

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+                              validation_data_shard=None):
+    # todo data_shard.head ?
+    first_data = data_shard.rdd.first()
+
+    assert batch_size != -1 or batch_per_thread != -1, \
+        "one of batch_size and batch_per_thread should be specified"
+
+    import tensorflow as tf
+    feature_spec = [(tf.dtypes.as_dtype(feat.dtype), feat.shape[1:])
+                    for feat in first_data["x"]]
```
jason-dai, line 64: `data['x']` should be a ndarray or list of ndarray (or even a dictionary of ndarray). We needs to check it's type, and process it accordingly.

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+    if "y" in first_data:
+        label_spec = [(tf.dtypes.as_dtype(label.dtype), label.shape[1:])
+                      for label in first_data["y"]]
```
jason-dai, line 67: `data['y']` should be a ndarray or list of ndarray (or even a dictionary of ndarray). We needs to check it's type, and process it accordingly.

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+    else:
+        label_spec = None
+
+    # todo this might be very slow
+    def flatten(data):
+        features = data["x"]
+
+        has_label = "y" in data
+        labels = data["y"] if has_label else None
+        length = features[0].shape[0]
```
jason-dai, line 77: this does not work if `features` is simply a ndarray.

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+        self.outputs = outputs
+        self.labels = labels
+        self.loss = loss
+        self.optimizer= optimizer
+        self.metrics = metrics
+        self.updates = updates
+        self.sess = sess
+        self.model_dir = model_dir
+
+    def fit(self, data_shard, steps,
```
jason-dai, line 122: is it efficient to call `fit` multiple times?

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,158 @@
...
+    import tensorflow as tf
+    feature_spec = [(tf.dtypes.as_dtype(feat.dtype), feat.shape[1:])
+                    for feat in first_data["x"]]
+    if "y" in first_data:
+        label_spec = [(tf.dtypes.as_dtype(label.dtype), label.shape[1:])
+                      for label in first_data["y"]]
+    else:
+        label_spec = None
+
+    # todo this might be very slow
```
jason-dai, line 71: Maybe it'll be more efficient if we first convert the RDD of ndarrays to RDD of JTensors, and can perform SampleToMinibatch on the Scala side? We can optimize this later.

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: done

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: done

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: deleted

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: added support for ndarray and list of ndarray

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: added support for a single ndarray

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: Are you suggesting to reuse TFOptimizer? Currently TFOptimizer does not support changing training data. (I think It is possible to support it.)

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: I thought about adding `TFDataset.from_xshards`, but then tfpark package will depend on something in orca package and orca package also depends on something in tfpark package. Feels kind of weird. 
I do not have strong opinions on this though.

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: added support for ndarray and list of ndarray

pyzoo/zoo/orca/learn/tensorflow/estimator.py
jason-dai: > Are you suggesting to reuse TFOptimizer? Currently TFOptimizer does not support changing training data. (I think It is possible to support it.)

I can support this in a separate PR

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -31,10 +31,10 @@ def predict(self, data, **kwargs):
         pass
 
     @staticmethod
-    def from_pre_built_graph(*, inputs, outputs,
-                             labels, loss, optimizer,
-                             metrics=None, updates=None,
-                             sess=None, model_dir=None, backend="spark"):
+    def from_graph(*, inputs, outputs,
+                   labels, loss, optimizer,
```
jason-dai, line 35: you should make `output`, `labels`, `loss` and `optimize`r optional. `output` is not needed for training, while the others are not needed in prediction

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,185 @@
...
+            if isinstance(y, np.ndarray):
+                new_y = [y]
+            elif isinstance(y, list) and all([isinstance(yi, np.ndarray) for yi in y]):
+                new_y = y
+            else:
+                raise ValueError("value of x should be a ndarray or a list of ndarrays")
+            result["y"] = new_y
+        return result
+
+    data_shard = data_shard.transform_shard(to_list)
```
jason-dai, line 82: why do we need to call `transform_shard`? The result is almost the same as the original data, while `transform_shard` can be expensive. Just check if data['x'] is a ndarray or list in `get_spec` and `flatten`

pyzoo/zoo/orca/learn/tensorflow/estimator.py
jason-dai: > I thought about adding `TFDataset.from_xshards`, but then tfpark package will depend on something in orca package and orca package also depends on something in tfpark package. Feels kind of weird.
> I do not have strong opinions on this though.

It might be useful if one wants use it for tfpark (as in @jenniew case)? anyway, the function name should be `_xshards_to_tf_dataset`

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: fixed

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: fixed

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: fixed

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: function name changed to `_xshards_to_tf_dataset`

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,193 @@
...
+        self.model_dir = model_dir
+
+    def fit(self, data_shard, steps,
+            batch_size=32,
+            validation_data_shard=None,
+            feed_dict=None,
+            session_config=None):
+        import bigdl
+
+        assert self.labels is not None, "labels should not be None when used for training"
```
jason-dai, line 156: "labels is None; it should not be None in training"

pyzoo/zoo/orca/learn/tensorflow/estimator.py
```diff
@@ -0,0 +1,193 @@
...
+            updates=self.updates, session=self.sess,
+            tensor_with_value=tensor_with_value,
+            session_config=session_config,
+            model_dir=self.model_dir)
+
+        optimizer.optimize(end_trigger=MaxIteration(steps))
+        return self
+
+    def predict(self, data_shard, batch_size=32):
+        assert self.outputs is not None, "outputs should not be None when used for training"
```
jason-dai, line 186: "outputs is None; it should not be None in prediction"

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: fixed

pyzoo/zoo/orca/learn/tensorflow/estimator.py
yangw1234: fixed

jason-dai:
LGTM

yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3462/

PR #2343, opened 2020-05-18
align tranform_shard api between spark and ray

yangw1234:
In RayDataShard, `transform_shard` requires a function whose first argument is the data and return new data, e.g.
```
def func(data, *args):
    data['count'] += 1
    return data
```
In SparkDataShard, `transform_shard` requires a function that return a single argument function to make the transform.

```
def func(*args):
    def transform(data):
        data['count'] += 1
         return data
    return transform
```

I think the ray version is easier to use.


jason-dai:
LGTM

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3397/ Jenkins passed

Issue #2299, opened 2020-05-06
more user-friendly docs for the time series algorithms in zouwu


shane-huang:
There's no detailed explanation about the models and it's hard for users to understand how to specify the paramters in Forecaster models or how to set the search ranges of parameters in autots. 

We need to add some docs for that. 

Add a new subsection "Algorithms"  in Zouwu and describe LSTM and MTNet in 2 seperate pages.  
https://analytics-zoo.github.io/master/#Zouwu/overview/

shane-huang:
docs should contain 1) an illustration of the model architecture 2) some description of the overall model architecture. 3) what are the parameters and how they affects the model, including what should be paid attention to.

shane-huang:
An user friendly doc example is https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/svm.html

PR #2345, opened 2020-05-18
Change Zouwu model document to make it more user-friendly

TheaperDeng:
Add more detailed param description for the MTNet model document.
Add algorithm introduction.
Working on the LSTM model document changing.


TheaperDeng:
related to issue #2299

PR #2347, opened 2020-05-19
Cluster Serving Sparse tensor and universal config, single stream input

Litchilitchy:
could support any type of input,

specify it in config,

a 3 inputs model, first image, second tensor, third sparse tensor, config should be
```
shape: [[3, 224, 224],[2, 1000], [4, 50000: 100]] # 4* 50000 sparse tensor with 100 non-empty
type: [image, tensor, sparse_tensor]
```


zoo/src/main/scala/com/intel/analytics/zoo/serving/preprocessing/PreProcessing.scala
```diff
@@ -0,0 +1,208 @@
...
+          })
+          tensorBuffer(idx) = Tensor.sparse[Float](indices, valueArray, param.dataShape(idx))
+        }
+      } catch {
+        case ex: IOException =>
+          logger.warn(ex.getMessage)
+      } finally {
+        reader.close()
+      }
+    }
```
cyita, line 129: Merge `decodeTensor` and `decodeSparseTensor`?

jason-dai:
> could support any type of input,
> 
> specify it in config,
> 
> a 3 inputs model, first image, second tensor, third sparse tensor, config should be
> 
> ```
> shape: [[3, 224, 224],[2, 1000], [4, 50000: 100]] # 4* 50000 sparse tensor with 100 non-empty
> type: [image, tensor, sparse_tensor]
> ```

I don't think it's reasonable to ask use to specify how many elements are non-empty in a sparse tensor

Litchilitchy:
> > could support any type of input,
> > specify it in config,
> > a 3 inputs model, first image, second tensor, third sparse tensor, config should be
> > ```
> > shape: [[3, 224, 224],[2, 1000], [4, 50000: 100]] # 4* 50000 sparse tensor with 100 non-empty
> > type: [image, tensor, sparse_tensor]
> > ```
> 
> I don't think it's reasonable to ask use to specify how many elements are non-empty in a sparse tensor

then just input ndarray of [0,0,0,0,*********,1,0,0,0,0,0,0,0,*****,0,0]?

jason-dai:
> > > could support any type of input,
> > > specify it in config,
> > > a 3 inputs model, first image, second tensor, third sparse tensor, config should be
> > > ```
> > > shape: [[3, 224, 224],[2, 1000], [4, 50000: 100]] # 4* 50000 sparse tensor with 100 non-empty
> > > type: [image, tensor, sparse_tensor]
> > > ```
> > 
> > 
> > I don't think it's reasonable to ask use to specify how many elements are non-empty in a sparse tensor
> 
> then just input ndarray of [0,0,0,0,...,1,0,0,0,0,0,0,0,....,0,0]?

Are you talking about the config file, or the input data?

In the configure files, user only needs to specify the type and shape, not the number of non-zero elements.

For in the input data, how is it specified in TF serving? I'd expect two ndarray, one for non-zero values, the other one for the associated indices, e.g.,:
```
#a sparse tensor t with 3 non-zero values: t[93]=3, t[1000]=12, t[30001]=7
value: [3, 12, 7]
index: [93, 1000, 30001]
```

Litchilitchy:
> > > could support any type of input,
> > > specify it in config,
> > > a 3 inputs model, first image, second tensor, third sparse tensor, config should be
> > > ```
> > > shape: [[3, 224, 224],[2, 1000], [4, 50000: 100]] # 4* 50000 sparse tensor with 100 non-empty
> > > type: [image, tensor, sparse_tensor]
> > > ```
> > 
> > 
> > I don't think it's reasonable to ask use to specify how many elements are non-empty in a sparse tensor
> 
> then just input ndarray of [0,0,0,0,****_****,1,0,0,0,0,0,0,0,****_,0,0]?



> > > > could support any type of input,
> > > > specify it in config,
> > > > a 3 inputs model, first image, second tensor, third sparse tensor, config should be
> > > > ```
> > > > shape: [[3, 224, 224],[2, 1000], [4, 50000: 100]] # 4* 50000 sparse tensor with 100 non-empty
> > > > type: [image, tensor, sparse_tensor]
> > > > ```
> > > 
> > > 
> > > I don't think it's reasonable to ask use to specify how many elements are non-empty in a sparse tensor
> > 
> > 
> > then just input ndarray of [0,0,0,0,...,1,0,0,0,0,0,0,0,....,0,0]?
> 
> Are you talking about the config file, or the input data?
> 
> In the configure files, user only needs to specify the type and shape, not the number of non-zero elements.
> 
> For in the input data, how is it specified in TF serving? I'd expect two ndarray, one for non-zero values, the other one for the associated indices, e.g.,:
> 
> ```
> #a sparse tensor t with 3 non-zero values: t[93]=3, t[1000]=12, t[30001]=7
> value: [3, 12, 7]
> index: [93, 1000, 30001]
> ```

okay then we can just do it as TF serving, and number of elements can be inferred from this input instead of provided.

jason-dai:
Please add more unit tests (for both Python and Scala)

pyzoo/zoo/serving/client.py
```diff
@@ -93,15 +96,86 @@ def data_shape_check(self):
...
         """
         :param id: String you use to identify this record
         :param data: Data, ndarray type
         :return:
         """
         if isinstance(img, str):
-            img = cv2.imread(str)
-            if not img:
+            img = cv2.imread(img)
+            if not isinstance(img, np.ndarray):
```
qiyuangong, line 178: `if img.size ==0` should be fine. :)
```
The function imread loads an image from the specified file and returns it. If the image cannot be read (because of missing file, improper permissions, unsupported or invalid format), the function returns an empty matrix ( Mat::data==NULL ).
```

Litchilitchy: noted

PR #2348, opened 2020-05-19
Orca MXNetTrainer XShards support

cyita:


pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -60,16 +60,19 @@ class MXNetTrainer(object):
     :param runner_cores: The number of CPU cores allocated for each MXNet worker and server.
     Default is None. You may need to specify this for better performance.
     """
-    def __init__(self, config, data_creator, model_creator,
+    def __init__(self, config, train_data, model_creator,
                  loss_creator=None, metrics_creator=None,
-                 num_workers=1, num_servers=None, runner_cores=None):
+                 num_workers=1, num_servers=None, runner_cores=None,
+                 test_data=None, train_resize_size=3268):
```
jason-dai, line 66: why do we need `train_resize_size`?

jason-dai, line 66: Is there a good way to determine the `train_resize_size` value? @hkvision

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -51,17 +52,37 @@ def setup_distributed(self, env, config, data_creator, model_creator,
...
                 mx.random.seed(self.config["seed"])
-            data = self.data_creator(self.config, self.kv)
-            if isinstance(data, tuple):
-                assert len(data) == 1 or len(data) == 2, \
-                    "data_creator should return either train_data only or a tuple of " \
-                    "(train_data, val_data), which can be directly fed to model training"
-                if len(data) == 1:
-                    self.train_data, self.val_data = data[0], None
+
+            if isinstance(train_data, tuple):
```
jason-dai, line 56: How does MXNetTrainer support ndarray? Does it makes sense?

jason-dai, line 56: How does MXNetTrainer support ndarray? Does it makes sense? @hkvision

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: It would be difficult to calculate unless we calculate the size for each shard and find the max... Probably we can add an attribute length in every shard?
Also, 3268 is not a good default value. Better to change to None in which case we suppose the data is evenly divided.

hkvision: train_data is an ndarray in this modification???

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
hkvision: why do we only support ndarray not datashard here? All the data can be loaded as an ndarray on the driver?

hkvision: We can choose to support ndarray. But this if branch to use tuple to identify ndarray is strange....

hkvision: The logic here is not correct. If we want to support ndarray, we need to evenly divide them; but not train the whole ndarray on every worker.

jason-dai: What's the previous assumption? `data_creator` will return a different ndarray on each runner?

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
jason-dai: We should assume `train_data` is an XShards of is dictionary:
```
{
"x": ndarray or list of ndarray
"y": ndarray or list of ndarray
} 
```

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -213,3 +256,23 @@ def get_node_ip(self):
...
         return find_free_port()
+
+
+def get_data_label(partition_data):
+    def combine(dict1, dict2):
+        return {key: np.concatenate((value, dict2[key]), axis=0)
+                for (key, value) in dict1.items()}
+
+    data_list = [data['data'] for data in partition_data]
+    label_list = [data['label'] for data in partition_data]
```
jason-dai, line 267: we can assume that train_data is an XShards of dictionary:
```
{
"x": ndarray or list of ndarray
"y": ndarray or list of ndarray
} 
```

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -213,3 +256,23 @@ def get_node_ip(self):
...
+        data = reduce(lambda dict1, dict2: combine(dict1, dict2), data_list)
+    elif isinstance(partition_data[0]['data'], np.ndarray):
+        data = reduce(lambda array1, array2: np.concatenate((array1, array2), axis=0),
+                      data_list)
+    if isinstance(partition_data[0]['label'], dict):
+        label = reduce(lambda dict1, dict2: combine(dict1, dict2), label_list)
+    elif isinstance(partition_data[0]['label'], np.ndarray):
+        label = reduce(lambda array1, array2: np.concatenate((array1, array2), axis=0),
+                       label_list)
+    return {'data': data, 'label': label}
```
jason-dai, line 278: `{'x': data, 'y': label}`

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
hkvision: yeah. Users are supposed to read different data based on `kv.rank` and the function returns different results for different workers.

jason-dai: we can remove ndarray support for now

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -194,6 +215,28 @@ def train(self, nb_epoch=1):
             stats["epoch_time"] = epoch_time
         return stats
 
+    def validate(self):
```
hkvision, line 218: I think we can first move validate for now as it is not tested.

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
cyita: So both `data_creator` and `xshards train_data & test_data` will be supported in this PR?

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
cyita: OK

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
jason-dai: We can remove data_creator, and support xshards only for now

hkvision: But the lenet example uses MNISTIter...
I think we'd better keep data_creator and if user don't use xshards, we can drop back to data_creator. That would be more flexible.

jason-dai: ok

hkvision: You mean input two data_creators?

What about the following logic with arguments data_shards=None, data_creator=None:
- If data_shards is not None, then data_creator is not used. Otherwise, use data_creator instead.
- data_shards can be a shard (train) or a tuple of shards (train and val). data_creator can return an iter (train) or a tuple of iters (train and val) [the current implementation does like this]

This would be clearer for the API.

jason-dai: > You mean input two data_creators?
> 
> What about the following logic with arguments data_shards=None, data_creator=None:
> 
> * If data_shards is not None, then data_creator is not used. Otherwise, use data_creator instead.
> * data_shards can be a shard (train) or a tuple of shards (train and val). data_creator can return an iter (train) or a tuple of iters (train and val) [the current implementation does like this]
> 
> This would be clearer for the API.

I would prefer to have two parameters, `train_data` and `test_data`, each of which can be either a creator function or a xshards.

hkvision: Okay. I'm okay with that. Then if `train_data` is an instance of xshard, we perform other logic; otherwise we require it to be a callable function.

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
cyita: Fixed

pyzoo/test/zoo/orca/learn/mxnet/test_mxnet_gluon.py
```diff
@@ -68,8 +72,8 @@ class TestMXNetGluon(TestCase):
     def test_gluon(self):
         config = create_trainer_config(batch_size=32, log_interval=2, optimizer="adam",
                                        optimizer_params={'learning_rate': 0.02})
-        trainer = MXNetTrainer(config, get_data_iters, get_model, get_loss, get_metrics,
-                               num_workers=2)
+        trainer = MXNetTrainer(config, get_test_data_iter, get_model, get_loss, get_metrics,
```
jason-dai, line 75: should be `get_train_data_iter`?

cyita: Fixed

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -213,3 +235,23 @@ def get_node_ip(self):
...
         """Finds a free port on the current node."""
         return find_free_port()
+
+
+def get_data_label(partition_data):
+    def combine(dict1, dict2):
+        return {key: np.concatenate((value, dict2[key]), axis=0)
+                for (key, value) in dict1.items()}
+
+    data_list = [data['data'] for data in partition_data]
```
jason-dai, line 245: As mentioned below, use 'x' and 'y' instead of 'data' and 'label' when retrieving data;
and we need to check if 'x' and 'y` exists in data.

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -213,3 +235,23 @@ def get_node_ip(self):
...
+    label_list = [data['label'] for data in partition_data]
+    if isinstance(partition_data[0]['data'], dict):
+        data = reduce(lambda dict1, dict2: combine(dict1, dict2), data_list)
+    elif isinstance(partition_data[0]['data'], np.ndarray):
+        data = reduce(lambda array1, array2: np.concatenate((array1, array2), axis=0),
+                      data_list)
```
jason-dai, line 251: it can also be a list of ndarray

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
cyita: Fixed

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
cyita: Fixed

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
cyita: Fixed

cyita:
Jenkis passed http://10.239.47.210:18888/job/ZOO-PR-Validation/3473/

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -38,8 +38,8 @@ class MXNetTrainer(object):
     You can specify "seed" in config to set random seed.
     You can specify "init" in seed to set model initializer.
 
-    :param data_creator: A function that takes config and kv as arguments and returns an MXNet
-    DataIter/DataLoader for training or a tuple of training and validation datasets.
+    :param train_data: A xShards for training or A function that takes config and kv as arguments
```
hkvision, line 41: An instance of `xShards` or a function....

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -50,26 +50,39 @@ class MXNetTrainer(object):
...
+    a list of MXNet metrics or corresponding string representations of metrics, for example,
+    'accuracy'. This is not needed if you don't need evaluation on the training data set.
     """
-    def __init__(self, config, data_creator, model_creator,
-                 loss_creator=None, metrics_creator=None,
-                 num_workers=1, num_servers=None, runner_cores=None):
+    def __init__(self, config, train_data, model_creator,
+                 loss_creator=None, validation_metrics_creator=None,
+                 num_workers=1, num_servers=None, runner_cores=None,
+                 test_data=None, train_resize_size=None, eval_metrics_creator=None):
```
hkvision, line 75: switch the order of train_resize_size and eval_metrics_creator?
Also better put them before num_workers, num_servers and runner_cores

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -50,26 +50,39 @@ class MXNetTrainer(object):
...
     :param num_servers: The number of servers for distributed training. Default is None and in this
     case it would be equal to the number of workers.
     :param runner_cores: The number of CPU cores allocated for each MXNet worker and server.
     Default is None. You may need to specify this for better performance.
+    :param test_data: A xShards for testing or A function that takes config and kv as arguments
+    and returns an MXNet DataIter/DataLoader for testing.
+    You can specify data related configurations for this function in the config argument above.
+    kv is an instance of MXNet distributed key-value store. kv.num_workers and kv.rank
+    can be used in this function to split data for different workers if necessary.
+    :param train_resize_size: The number of batches per epoch to resize to.
```
hkvision, line 67: Is there a better name for this? `resize_size` seems strange...

hkvision: Add more description. `Default is None. You might need to specify this if the size of train_data for each worker varies.`

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: Eval if for train data right? Then it is better to put it before test_data and validation_metrics_creator.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: Keep the order align with the argument orders.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -50,26 +50,45 @@ class MXNetTrainer(object):
     :param loss_creator: A function that takes config as argument and returns an MXNet loss.
     This is not needed for symbolic API where loss is already defined as model output.
 
-    :param metrics_creator: A function that takes config as argument and returns one or a list of
-    MXNet metrics or corresponding string representations of metrics, for example, 'accuracy'.
-    This is not needed if you don't have validation data throughout the training.
+    :param train_resize_batch_num: The number of batches per epoch to resize to. Default is None,
```
hkvision, line 53: , -> .

hkvision:
LGTM except the minor comment above.

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -213,3 +248,33 @@ def get_node_ip(self):
...
+                for index in range(0, len(list1))]
+
+    data_list = [data['x'] for data in partition_data]
+    label_list = [data['y'] for data in partition_data]
+    if isinstance(partition_data[0]['x'], dict):
+        data = reduce(lambda dict1, dict2: combine_dict(dict1, dict2), data_list)
+    elif isinstance(partition_data[0]['x'], np.ndarray):
+        data = reduce(lambda array1, array2: np.concatenate((array1, array2), axis=0),
+                      data_list)
+    elif isinstance(partition_data[0]['x'], list):
```
jason-dai, line 269: it seems tf.data.Dataset only support tuple of ndarray, not list of ndarray according to @yangw1234; we should change this to tuple to make it consistent with https://github.com/intel-analytics/analytics-zoo/pull/2375

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
cyita: Fixed

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
cyita: mx.io.NDArrayIter only support array or list of array or dict of string to array.

cyita:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/3486/

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -213,3 +248,33 @@ def get_node_ip(self):
...
+                for (key, value) in dict1.items()}
+
+    def combine_list(list1, list2):
+        return [np.concatenate((list1[index], list2[index]), axis=0)
+                for index in range(0, len(list1))]
+
+    data_list = [data['x'] for data in partition_data]
+    label_list = [data['y'] for data in partition_data]
+    if isinstance(partition_data[0]['x'], dict):
+        data = reduce(lambda dict1, dict2: combine_dict(dict1, dict2), data_list)
```
hkvision, line 265: This can only join several batches of records together in a larger batch. But it cannot join several single records to a batch, which is the case from RDD.
It is not reasonable for users to add an extra dimension by themselves.
What do you think is the best way to fix the conflicts? @jason-dai

jason-dai: > This can only join several batches of records together in a larger batch. But it cannot join several single records to a batch, which is the case from RDD.
> It is not reasonable for users to add an extra dimension by themselves.
> What do you think is the best way to fix the conflicts? @jason-dai

I don't think I understand the problem. For the following code,
```
train_data = train_data.repartition(self.num_workers).to_ray()
```
Each partition of `train_data` is supposed to contain a dictionary; after `repartition`, each partition should contain several dictionaries. `to_ray` should store a list of dictionary in plasma; then we can apply exactly the same logic to process the data, no?

PR #2349, opened 2020-05-19
Add initial version of PyTorchTrainer in orca.learn

hkvision:


hkvision:
@jason-dai Is this what you expect?

pyzoo/zoo/orca/learn/torch/__init__.py
```diff
@@ -0,0 +1,17 @@
+#
```
jason-dai, line 1: `orca.learn.pytorch`

pyzoo/zoo/orca/learn/torch/torch_trainer.py
```diff
@@ -0,0 +1,51 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from ray.util.sgd.torch import TorchTrainer as RaySGDTorchTrainer
+
+
+class TorchTrainer(object):
```
jason-dai, line 20: PyTorchTrainer

pyzoo/zoo/orca/learn/torch/__init__.py
hkvision: In TorchNet, the package is torch not pytorch, shouldn't we keep aligned?

jason-dai: let's make it explicit that this is for pythorch

hkvision: sure!

pyzoo/zoo/orca/learn/pytorch/pytorch_trainer.py
```diff
@@ -0,0 +1,55 @@
...
+class PyTorchTrainer(object):
+    def __init__(self, model_creator, data_creator, optimizer_creator,
+                 loss_creator=None, scheduler_creator=None, training_operator_cls=None,
+                 initialization_hook=None, config=None, num_workers=1,
+                 use_fp16=False, use_tqdm=False,
+                 scheduler_step_freq="batch", num_replicas=None,
+                 batch_size=None, data_loader_args=None):
+        # Lift TorchTrainer to an Actor so that its local worker would be
+        # created on the cluster as well.
+        RemoteTrainer = ray.remote(TorchTrainer)
```
hkvision, line 30: @jason-dai In our yarn settings, lifting the original TorchTrainer as an Actor would be better. See this issue for more discussion: https://github.com/ray-project/ray/issues/8336

hkvision:
@jason-dai I add a ut. Take a further look. Would this pr be good for an initial version?

pyzoo/zoo/orca/learn/pytorch/pytorch_trainer.py
jason-dai: what if we are running on local?

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3427/

pyzoo/zoo/orca/learn/pytorch/pytorch_trainer.py
hkvision: We can do this, but there's some concerns:
- Normally driver would start with 0 cores, we need to ask users to explicitly specify how many cores for the driver.
- Sometimes the driver may not have enough resources?
- If the driver hardware is worse than workers, the driver will be the bottleneck for the distributed training.

Do you want to keep the local driver?
If so, I would test this. Previously I test the remote way.

hkvision: > what if we are running on local?

Has no problem with ray local. The unit test covers :)

jason-dai:
LGTM

PR #2370, opened 2020-05-26
move pytorch trainer to sub folder

qiuxin2012:
move pytorch trainer to sub folder


qiuxin2012:
@jason-dai Huangkai and I think using a sub folder is better to avoid importing unnessessary dependence.

jason-dai:
> @jason-dai Huangkai and I think using a sub folder is better to avoid importing unnessessary dependence.

Can we just use separate files? And don't create additional module hierarchy like `zoo.orca.learn.pytorch.ray`

hkvision:
> > @jason-dai Huangkai and I think using a sub folder is better to avoid importing unnessessary dependence.
> 
> Can we just use separate files? And don't create additional module hierarchy like `zoo.orca.learn.pytorch.ray`

Separating files and removing the import in init will lead to `from zoo.orca.learn.pytorch.pytorch_trainer import PyTorchTrainer`, this would be inconvenient?

jason-dai:
> > > @jason-dai Huangkai and I think using a sub folder is better to avoid importing unnessessary dependence.
> > 
> > 
> > Can we just use separate files? And don't create additional module hierarchy like `zoo.orca.learn.pytorch.ray`
> 
> Separating files and removing the import in init will lead to `from zoo.orca.learn.pytorch.pytorch_trainer import PyTorchTrainer`, this would be inconvenient?

Maybe we can create separate files under `orca.learn`, such as pythorch.py, mxnet.py, tf.py,, etc., and put the corresponding estimator/trainer classes in the file, and import necessary modules only inside the classes.

PR #2371, opened 2020-05-26
remove long in torchloss

qiuxin2012:
Because TorchLoss's input from scala are Float tensor currently, so for the loss like CrossEntropyLoss who require long-type target, we can use a lossFunction to wrap the Loss:
```
def lossFunc(input, target):
    return nn.CrossEntropyLoss().forward(input, target.flatten().long())
```


qiuxin2012:
jenkins passed

PR #2376, opened 2020-05-26
RayContext init return address info

hkvision:
Return plasma address, etc.
Same as ray.init(), it returns such info.

### DIFF
--- a/pyzoo/zoo/ray/raycontext.py
+++ b/pyzoo/zoo/ray/raycontext.py
@@ -308,12 +308,12 @@ class RayContext(object):
             if self.env:
                 os.environ.update(self.env)
             import ray
-            ray.init(num_cpus=self.ray_node_cpu_cores,
-                     object_store_memory=self.object_store_memory,
-                     resources=self.extra_params)
+            return ray.init(num_cpus=self.ray_node_cpu_cores,
+                            object_store_memory=self.object_store_memory,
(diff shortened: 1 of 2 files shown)


hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3464/

pyzoo/zoo/ray/raycontext.py
```diff
@@ -308,12 +308,12 @@ def init(self, driver_cores=0):
             if self.env:
                 os.environ.update(self.env)
             import ray
-            ray.init(num_cpus=self.ray_node_cpu_cores,
-                     object_store_memory=self.object_store_memory,
-                     resources=self.extra_params)
+            return ray.init(num_cpus=self.ray_node_cpu_cores,
```
jason-dai, line 311: instead return this to user, store the info in RayContext; the user can the call RayContext methods to get relevant information.

hkvision: What about both return and store? This aligns with ray as it ray init does return something...

jason-dai: ok

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3494/

PR #2377, opened 2020-05-26
[WIP] Ray read Spark RDD

hkvision:
https://github.com/analytics-zoo/orca/issues/34

@jason-dai An initial version that works on YARN. Take a look.

### DIFF
--- a/pyzoo/zoo/examples/orca/data/rdd_ray.py
+++ b/pyzoo/zoo/examples/orca/data/rdd_ray.py
@@ -1,0 +1,71 @@
+import ray
+from zoo import init_spark_on_yarn
+from zoo.ray import RayContext
+
+sc = init_spark_on_yarn(
+    hadoop_conf="/opt/work/hadoop-2.7.2/etc/hadoop",
+    conda_name="kai-hyperseg",
+    num_executor=2,
+    executor_cores=44,
+    executor_memory="10g",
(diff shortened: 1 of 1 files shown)


pyzoo/zoo/examples/orca/data/rdd_ray.py
```diff
@@ -0,0 +1,71 @@
...
+    import socket
+    random.seed(splitIndex)
+    res = list(iterator)
+    client = plasma.connect(plasma_address)
+    object_id = client.put(res)
+    # Not directly calling ray.services.get_node_ip_address()
+    # since this would introduce ray overhead.
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    s.connect(("8.8.8.8", 80))
+    node_ip = s.getsockname()[0]
```
jason-dai, line 46: is there a more reliable way to get node ip (which needs to be consistent with Ray workers)?

hkvision: This is exactly the same code in `ray.services.get_node_ip_address()`. Copy it here instead of calling it since ray will introduce overhead.

jason-dai: will it work inside the datacenter w/o internet connection?

PR #2384, opened 2020-05-27
Orca len() operation on XShard on Spark

jenniew:
Add len(key) support for XShards on Spark.
If key is None, return sum of len() on each shard.
If key provided, return sum of len(data[key]) on each shard.

Related issue: https://github.com/analytics-zoo/orca/issues/28


jason-dai:
LGTM

jenniew:
Jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3507/

pyzoo/zoo/orca/data/shard.py
```diff
@@ -208,6 +208,22 @@ def transform(data):
             else:
                 return [self]
 
+    def len(self, key=None):
+        if key is None:
+            return self.rdd.map(lambda data: len(data) if hasattr(data, '__len__') else 1)\
+                .reduce(lambda l1, l2: l1 + l2)
+        else:
+            first = self.rdd.first()
```
jason-dai, line 216: do this check on the task side, instead of collecting the data back to driver

PR #2392, opened 2020-05-30
Orca len on spark change check to task

jenniew:
For xshards.len(key), change key check on each shard instead of checking on driver after collect.


jenniew:
Jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3508/

PR #2396, opened 2020-06-01
Orca add SparkXShard cache management

cyita:


cyita:
Related issue [Manage the caching of SparkXShards #32](https://github.com/analytics-zoo/orca/issues/32)

cyita:
> It seems uncache() and del have same operations. Do we need two operations?

I'm not sure if users need the un-cache method other than del to un-cache xShards.

jason-dai:
> It seems uncache() and del have same operations. Do we need two operations?

Which one is more pythonic `shards.uncache()` or `del shards` @shane-huang ?

pyzoo/zoo/orca/data/shard.py
```diff
@@ -113,18 +113,39 @@ def get_data(self):
...
+        return self.rdd.is_cached
+
     def num_partitions(self):
         return self.rdd.getNumPartitions()
 
     def repartition(self, num_partitions):
-        return SparkXShards(self.rdd.repartition(num_partitions))
+        repartitioned_shard = SparkXShards(self.rdd.repartition(num_partitions))
+        if not self.user_cached:
+            self.uncache()
```
jason-dai, line 147: Wrap this in another method like `_uncache`?

pyzoo/zoo/orca/data/shard.py
```diff
@@ -135,6 +134,10 @@ def uncache(self):
         self.rdd.unpersist()
         return self
 
+    def __uncache(self):
```
jason-dai, line 137: why double underscore instead of single undersore?

jason-dai:
LGTM

cyita:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/3528/

PR #2402, opened 2020-06-02
SparkXShards to RayXShards

hkvision:
@jason-dai What about this implementation? Take a look.
https://github.com/analytics-zoo/orca/issues/34


pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -56,11 +54,12 @@ def setup_distributed(self, env, config, train_data, model_creator, loss_creator
...
-                # XShard input
-                # retrieve train data
-                train_data = train_data.get_partitions()
-                train_partition_data = ray.get(train_data[self.kv.rank].get_data())
+            from zoo.orca.data import RayXShards
+            if isinstance(train_data, RayXShards):
+                # Retrieve train data for RayXShards
+                train_data_partition = train_data.get_partitions()[self.kv.rank]
+                assert self.node_ip == train_data_partition.node_ip
+                train_partition_data = ray.get(train_data_partition.get_data())
```
yangw1234, line 62: ray.get requires an object id, right? if this is the plasma case, `get_data()` will return the actual data. Does it also work in this case?

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -56,11 +54,12 @@ def setup_distributed(self, env, config, train_data, model_creator, loss_creator
...
-            if isinstance(train_data, XShards):
-                # XShard input
-                # retrieve train data
-                train_data = train_data.get_partitions()
-                train_partition_data = ray.get(train_data[self.kv.rank].get_data())
+            from zoo.orca.data import RayXShards
+            if isinstance(train_data, RayXShards):
+                # Retrieve train data for RayXShards
+                train_data_partition = train_data.get_partitions()[self.kv.rank]
+                assert self.node_ip == train_data_partition.node_ip
```
yangw1234, line 61: train_data_partition.node_ip could be a `None`?

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
hkvision: Fixed. Put ray.get in get_data(); get_data will always return the actual data.

hkvision:
Would add unit test and test on RBI code if the API is OK.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -90,6 +90,18 @@ def __init__(self, config, train_data, model_creator,
...
+        if isinstance(train_data, SparkXShards):
+            assert isinstance(test_data, SparkXShards)
+            from zoo.ray import RayContext
+            ray_ctx = RayContext.get()
+            train_data = train_data.repartition(self.num_workers).to_ray(ray_ctx)
+            test_data = test_data.repartition(self.num_workers).to_ray(ray_ctx)
+        if isinstance(train_data, RayXShards):
+            assert isinstance(test_data, RayXShards)
+            assert train_data.get_partitions() == self.num_workers
+            assert test_data.get_partitions() == self.num_workers
```
jason-dai, line 103: I think we will need to call `repartition(self.num_workers)` here too?

pyzoo/zoo/orca/data/shard.py
```diff
@@ -231,3 +239,29 @@ def save_pickle(self, path, batchSize=10):
...
+
+            random.seed(splitIndex)
+            res = list(iterator)
+            client = plasma.connect(object_store_address)
+            object_id = client.put(res)
+            # Not directly calling ray.services.get_node_ip_address()
+            # since this would introduce ray overhead.
+            s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            s.connect(("8.8.8.8", 80))
+            node_ip = s.getsockname()[0]
```
jason-dai, line 259: wrap this in a utility function

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -100,23 +112,30 @@ def __init__(self, config, train_data, model_creator,
...
 
         # Compute URL for initializing distributed setup
-        ips = ray.get(
-            [runner.get_node_ip.remote() for runner in self.runners])
-        ports = ray.get(
-            [runner.find_free_port.remote() for runner in self.runners])
+        worker_ips = ray.get(
+            [runner.get_node_ip.remote() for runner in self.workers])
+        server_ips = ray.get(
+            [runner.get_node_ip.remote() for runner in self.workers])
```
jason-dai, line 128: `for runner in self.servers`
Why do we need `server_ips` anyway?

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -56,11 +54,13 @@ def setup_distributed(self, env, config, train_data, model_creator, loss_creator
...
-                # XShard input
-                # retrieve train data
-                train_data = train_data.get_partitions()
-                train_partition_data = ray.get(train_data[self.kv.rank].get_data())
+            from zoo.orca.data import RayXShards
+            if isinstance(train_data, RayXShards):
+                # Retrieve train data for RayXShards
+                train_data_partition = train_data.get_partitions()[self.kv.rank]
+                if train_data_partition.node_ip:
+                    assert self.node_ip == train_data_partition.node_ip
```
jason-dai, line 62: `self.node_ip` can be potentially None?

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
```diff
@@ -71,10 +71,12 @@ def setup_distributed(self, env, config, train_data, model_creator, loss_creator
...
                 else:
-                    assert isinstance(test_data, XShards), "Test data should be an instance of " \
-                                                           "XShards, please check your input"
-                    test_data = test_data.get_partitions()
-                    val_partition_data = ray.get(test_data[self.kv.rank].get_data())
+                    assert isinstance(test_data, RayXShards), "Test data should be an instance of " \
+                                                           "RayXShards, please check your input"
+                    test_data_partition = test_data.get_partitions()[self.kv.rank]
+                    if test_data_partition.node_ip:
+                        assert self.node_ip == test_data_partition.node_ip
```
jason-dai, line 78: self.node_ip can be potentially None?

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -100,23 +112,30 @@ def __init__(self, config, train_data, model_creator,
...
+        # Sort workers based on their node ips
+        worker_and_ips = list(zip(self.workers, worker_ips))
+        worker_and_ips.sort(key=lambda x: x[1])
+        self.workers = [worker_ip[0] for worker_ip in worker_and_ips]
+        worker_ips = [worker_ip[1] for worker_ip in worker_and_ips]
+        self.runners = self.workers + self.servers
```
jason-dai, line 134: Maybe we can wrap this in a method, such as `self.workers = train_data.colocate_actors(self.workers)` ?

pyzoo/zoo/orca/data/shard.py
```diff
@@ -231,3 +239,29 @@ def save_pickle(self, path, batchSize=10):
...
+            s.close()
+            yield object_id, node_ip
+
+        object_id_node_ips = self.rdd.mapPartitionsWithIndex(put_to_plasma).collect()
```
jason-dai, line 263: shall we call `_uncache()` here?

yangw1234:
does the data in plasma store need to be deleted, such as when the rayxshards goes out of scope?

jason-dai:
> does the data in plasma store need to be deleted, such as when the rayxshards goes out of scope?

Yes; maybe add a `__del__` method

pyzoo/zoo/orca/data/shard.py
```diff
@@ -231,3 +239,29 @@ def save_pickle(self, path, batchSize=10):
     @classmethod
     def load_pickle(cls, path, sc, minPartitions=None):
         return SparkXShards(sc.pickleFile(path, minPartitions))
+
+    def to_ray(self, ray_ctx):
+        object_store_address = ray_ctx.address_info["object_store_address"]
+
+        def put_to_plasma(splitIndex, iterator):
```
yangw1234, line 246: If some of the tasks fail, we could also have a leak, either by retries or cancel of the whole job. Although this might not matters that much.

jason-dai: > If some of the tasks fail, we could also have a leak, either by retries or cancel of the whole job. Although this might not matters that much.

Does plasma support replace? If yes, we can first add a small value to get the object id and store the ids in the `XShards`; then we can update the object in plasma using the actual value. We can add the support in a follow-up PR.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: Put it here to log the server ips and would probably be useful for debugging?

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: So what's the assumption before my modifications? Why there's no repartition previously?

jason-dai: > So what's the assumption before my modifications? Why there's no repartition previously?

There was re-partitioning in @jenniew 's previous code; it may be somehow missed when migrating it over @cyita ?

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
jason-dai: Better not to deal with low level details (such as node ip) in user's code

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: OK. I will add it back.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: Why it's user's code? Users are not see this code, it's just our implementation.
Still you prefer to move server ip?

jason-dai: Better to include all the details in another method, such as `XShards.colocate_actors` as proposed below.

hkvision: Why to put colocate_actor in XShard not in MXNetTrainer? XShard is only about data, why it is related to workers???

pyzoo/zoo/orca/data/shard.py
hkvision: Added. It is reasonable that when we convert the SparkXShard to Ray, then the RDD may not be used.

hkvision:
Added TODOs for deletion and error handling. Would address them in future work.

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
hkvision: self.node_ip is the ip of the worker, it is always not None.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
jason-dai: > Why to put colocate_actor in XShard not in MXNetTrainer? XShard is only about data, why it is related to workers???

This should be used by other estimators on Ray, not only MXNet

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
hkvision: Previously for RayXShards, data locality is not handled and partition.node_ip is None, right? So, here if partition.node_ip is not None, we suppose it comes from plasma directly and do the match of ips. Is it OK?

jason-dai: > Previously for RayXShards, data locality is not handled and partition.node_ip is None, right? So, here if partition.node_ip is not None, we suppose it comes from plasma directly and do the match of ips. Is it OK?

I don't think it's a good idea to makes these type of implicit assumptions, as you have no control of how the client of your code may use your method. Better to call `self.get_node_ip` here and change the function to return `self.node_ip` if it's not None.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: Updated.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
hkvision: Updated.

pyzoo/zoo/orca/learn/mxnet/mxnet_runner.py
hkvision: Updated.

pyzoo/zoo/orca/data/shard.py
```diff
@@ -280,3 +304,29 @@ def load_pickle(cls, path, minPartitions=None):
...
+            random.seed(splitIndex)
+            res = list(iterator)
+            client = plasma.connect(object_store_address)
+            object_id = client.put(res)
+            yield object_id, get_node_ip()
+
+        object_id_node_ips = self.rdd.mapPartitionsWithIndex(put_to_plasma).collect()
+        self.uncache()
+        # Sort the data according to the node_ips.
+        object_id_node_ips.sort(key=lambda x: x[1])
```
jason-dai, line 329: no need to sort here, as it will be sorted in `colocate_actors`?

pyzoo/zoo/orca/data/shard.py
```diff
@@ -99,17 +99,41 @@ def get_partitions(self):
...
+    def colocate_actors(self, actors):
+        """
+        Sort Ray actors and RayPartitions by node_ip so that each actor is colocated
+        with the data partition on the same node.
+        """
+        import ray
+        actor_ips = ray.get([actor.get_node_ip.remote() for actor in actors])
+        actor_zip_ips = list(zip(actors, actor_ips))
+        actor_zip_ips.sort(key=lambda x: x[1])
+        if self.partitions[0].node_ip:
```
jason-dai, line 111: should check `node_ip` at the beginning of the method?

pyzoo/zoo/orca/data/shard.py
hkvision: It should be. But one concern is that we have both `train_data` and `test_data`, I suppose we only colocate for train_data? (Would colocate on both for two times cause some potential error??) Then what about test_data? It is not sorted. So I don't remove the sort here.

pyzoo/zoo/orca/data/shard.py
hkvision: okay. Then if no node_ip, just do nothing and return.

pyzoo/zoo/orca/data/shard.py
jason-dai: > It should be. But one concern is that we have both `train_data` and `test_data`, I suppose we only colocate for train_data? (Would colocate on both for two times cause some potential error??) Then what about test_data? It is not sorted. So I don't remove the sort here.

Then maybe you don't need to re-sort inside `colocate_actors`; if it's not already sorted, then there will also be errors associated with test_data

hkvision: > > It should be. But one concern is that we have both `train_data` and `test_data`, I suppose we only colocate for train_data? (Would colocate on both for two times cause some potential error??) Then what about test_data? It is not sorted. So I don't remove the sort here.
> 
> Then maybe you don't need to re-sort inside `colocate_actors`; if it's not already sorted, then there will also be errors associated with test_data

So, we first of all assume the partitions in RayPartition is sorted and I don't do the re-sort. May handle this better in later PRs.

pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -90,6 +87,24 @@ def __init__(self, config, train_data, model_creator,
         self.num_servers = num_servers if num_servers else self.num_workers
         self.train_resize_batch_num = train_resize_batch_num
 
+        from zoo.orca.data import RayXShards, SparkXShards
+        if isinstance(train_data, SparkXShards):
+            from zoo.ray import RayContext
+            ray_ctx = RayContext.get()
+            train_data = train_data.repartition(self.num_workers).to_ray(ray_ctx)
```
jason-dai, line 94: maybe we can try to get the `ray_ctx` inside `SparkXShards.to_ray`

jason-dai:
LGTM

hkvision:
@jason-dai Finished and add unit tests. Take another look?

jason-dai:
LGTM

hkvision:
Don't know why `pyarrow==0.16.0` would raise such error:
```
E                   py4j.protocol.Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
E                   : org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 7.0 failed 1 times, most recent failure: Lost task 1.0 in stage 7.0 (TID 43, localhost, executor driver): org.apache.spark.SparkException: Python worker exited unexpectedly (crashed)
E                   	at org.apache.spark.api.python.BasePythonRunner$ReaderIterator$$anonfun$3.applyOrElse(PythonRunner.scala:486)

2020-06-11 08:56:57 ERROR Executor:91 - Exception in task 1.0 in stage 7.0 (TID 43)
org.apache.spark.SparkException: Python worker exited unexpectedly (crashed)
	at org.apache.spark.api.python.BasePythonRunner$ReaderIterator$$anonfun$3.applyOrElse(PythonRunner.scala:486)
	at org.apache.spark.api.python.BasePythonRunner$ReaderIterator$$anonfun$3.applyOrElse(PythonRunner.scala:475)
```

But 0.17.0 works. Can we upgrade the jenkins pyarrow version to 0.17.0? @jason-dai @glorysdj

jason-dai:
Yes, we can upgrade the pyarrow version

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3566/

pyzoo/zoo/orca/data/shard.py
yangw1234: It seems we cannot replace an object. But we could first generate the object ids, then use those ids to put, which makes this operation idempotent.

pyzoo/zoo/orca/data/shard.py
```diff
@@ -280,3 +310,39 @@ def load_pickle(cls, path, minPartitions=None):
...
+        ray_ctx = RayContext.get()
+        object_store_address = ray_ctx.address_info["object_store_address"]
+
+        # TODO: Handle failure when doing this?
+        # TODO: delete the data in the plasma?
+        def put_to_plasma(seed):
+            def f(index, iterator):
+                import pyarrow.plasma as plasma
+                from zoo.orca.data.utils import get_node_ip
+                # mapPartition would set the same random seed for each partition?
```
hkvision, line 327: Seems the random seed issue exists in Python 3.6 but NOT in Python 3.7... Weird...

PR #2411, opened 2020-06-04
Get the current RayContext

hkvision:
RayContext.get() to retrieve the current active RayContext.
@jason-dai The OrCreate part is not necessary for now and I don't include it here.
Get the RayContext is enough for usage.


hkvision:
jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/3522/

jason-dai:
LGTM

PR #2420, opened 2020-06-09
Orca remove sc parameter in load_pickle()

jenniew:
Remove sc parameter in load_pickle(), using sc from SparkContext.getOrCreate()


pyzoo/zoo/orca/data/shard.py
```diff
@@ -273,7 +273,9 @@ def save_pickle(self, path, batchSize=10):
         return self
 
     @classmethod
-    def load_pickle(cls, path, sc, minPartitions=None):
+    def load_pickle(cls, path, minPartitions=None):
+        from pyspark.context import SparkContext
+        sc = SparkContext.getOrCreate()
```
jason-dai, line 278: `sc = self.rdd.ctx`?

jenniew: load_pickle() is a class method to create XShards object. Before we call load_pickle, we don't have any XShards object, so we cannot use self.rdd.ctx

hkvision: sc = init_nncontext()
If you just call SparkContext.getOrCreate, then it has no bigdl/zoo configurations, which would cause error in the case where engine.init is needed for example.

jenniew: updated.

jenniew:
Jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3539/

pyzoo/zoo/orca/data/shard.py
```diff
@@ -273,7 +274,13 @@ def save_pickle(self, path, batchSize=10):
         return self
 
     @classmethod
-    def load_pickle(cls, path, sc, minPartitions=None):
+    def load_pickle(cls, path, minPartitions=None):
+        from pyspark.context import SparkContext
+        with SparkContext._lock:
```
jason-dai, line 279: why don't you just call init_nncontext?

jenniew: updated

jason-dai:
LGTM

PR #2421, opened 2020-06-09
fix pytorch featureset, total train sample is incorrect.

qiuxin2012:
fix pytorch featureset, total train sample is incorrect.


qiuxin2012:
jenkins passed

PR #2425, opened 2020-06-09
Orca file add exists, makedirs

jenniew:
Add file utility functions: exists and makedirs, which support local/hdfs/s3 file systems.


jenniew:
Jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3541/

PR #2428, opened 2020-06-10
Orca file add write_text

jenniew:
Add write_text operation in orca.data.file, which supports to write string to local/hdfs/s3 file path.


jason-dai:
LGTM

PR #2430, opened 2020-06-10
support evaluate in orca tf estimator

yangw1234:
Usage:
```python
    data_shard = data_shard.transform_shard(transform)

     est = Estimator.from_graph(
         inputs=[model.user, model.item],
         labels=[model.label],
         loss=model.loss,
         optimizer=tf.train.AdamOptimizer(),
         metrics={"loss": model.loss})
     result = est.evaluate(data_shard)
     assert "loss" in result
     print(result)
```

Fix https://github.com/analytics-zoo/orca/issues/51

### DIFF
--- a/pyzoo/zoo/orca/learn/tf/estimator.py
+++ b/pyzoo/zoo/orca/learn/tf/estimator.py
@@ -170,7 +170,6 @@ class TFOptimizerWrapper(Estimator):
         else:
             self.sess = sess
         self.model_dir = model_dir
-        self.real_batch_size = tf.shape(self.inputs[0])[0]
 
     def fit(self, data, steps,
             batch_size=32,
--- a/pyzoo/zoo/orca/learn/tf/estimator.py
+++ b/pyzoo/zoo/orca/learn/tf/estimator.py
@@ -14,13 +14,15 @@
 # limitations under the License.
 #
 from bigdl.optim.optimizer import MaxIteration, SGD
+
+from zoo.tfpark.utils import evaluate_metrics
 from zoo.orca.data.shard import SparkXShards
 from zoo.orca.data.tf.data import Dataset, TFDataDataset2
-from zoo.tfpark import TFEstimator, TFOptimizer, TFPredictor, TFNet, ZooOptimizer
-import pandas as pd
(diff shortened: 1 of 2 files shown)


yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3560/

jenniew:
Does the evaluation do the shuffling to change the order?

jenniew:
> Does the evaluation do the shuffling to change the order?

Seems no shuffling. If so, we can use in NCF case.

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -154,6 +170,7 @@ def __init__(self, *, inputs, outputs, labels, loss,
         else:
             self.sess = sess
         self.model_dir = model_dir
+        self.real_batch_size = tf.shape(self.inputs[0])[0]
```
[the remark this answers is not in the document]
yangw1234: users do not need to care about that. When averaging metrics, we need to take into account the real_batch_size when averaging.

yangw1234:
No, the evaluation does not change data placement.

jenniew:
LGTM

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -204,16 +210,22 @@ def predict(self, data, batch_size=32):
...
+        dataset = _to_dataset(data, batch_size=-1, batch_per_thread=batch_size)
 
         flat_inputs = nest.flatten(self.inputs)
         flat_outputs = nest.flatten(self.outputs)
         tfnet = TFNet.from_session(sess=self.sess, inputs=flat_inputs, outputs=flat_outputs)
         return tfnet.predict(dataset)
+
+    def evaluate(self, data, batch_size=32):
+        assert self.metrics is not None, \
+            "output is None, it should not be None in prediction"
```
jason-dai, line 222: metrics is None

yangw1234: will open an pr to fix this

Issue #2433, opened 2020-06-11
XShards __del__ gets exception if xshards was not cached before


jenniew:
When xshards.__del__ is called by gc, if xshards was not cached before, it gets such exception:
Exception ignored in: <bound method SparkXShards.__del__ of <zoo.orca.data.shard.SparkXShards object at 0x7fa5ce0c1470>>
Traceback (most recent call last):
  File "/home/jwang/git/analytics-zoo-jennie4/dist/lib/analytics-zoo-bigdl_0.10.0-spark_2.4.3-0.9.0-SNAPSHOT-python-api.zip/zoo/orca/data/shard.py", line 282, in __del__
  File "/home/jwang/git/analytics-zoo-jennie4/dist/lib/analytics-zoo-bigdl_0.10.0-spark_2.4.3-0.9.0-SNAPSHOT-python-api.zip/zoo/orca/data/shard.py", line 148, in uncache
  File "/tools/spark-2.4.3-bin-hadoop2.7/python/lib/pyspark.zip/pyspark/rdd.py", line 255, in unpersist
  File "/tools/spark-2.4.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1257, in __call__
  File "/tools/spark-2.4.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/protocol.py", line 332, in get_return_value
py4j.protocol.Py4JError: An error occurred while calling o81.unpersist. Trace:
py4j.Py4JException: Target Object ID does not exist for this gateway :o81
	at py4j.Gateway.invoke(Gateway.java:279)
	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
	at py4j.commands.CallCommand.execute(CallCommand.java:79)
	at py4j.GatewayConnection.run(GatewayConnection.java:238)
	at java.lang.Thread.run(Thread.java:745)

jason-dai:
we should catch the error in __del__ @cyita

jason-dai:
and we may also need to check if the rdd is cached before calling `unpersist`?

cyita:
> and we may also need to check if the rdd is cached before calling `unpersist`?

Fixed in #2434

PR #2434, opened 2020-06-11
[bugfix] Orca sparkxshards uncache bugfix

cyita:
related issue: [XShards __del__ gets exception if xshards was not cached before #2433](https://github.com/intel-analytics/analytics-zoo/issues/2433)

### DIFF
--- a/pyzoo/zoo/orca/data/shard.py
+++ b/pyzoo/zoo/orca/data/shard.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from py4j.protocol import Py4JError
 
 from zoo.orca.data.utils import *
 from zoo.common.nncontext import init_nncontext
@@ -146,7 +147,10 @@ class SparkXShards(XShards):
     def uncache(self):
         self.user_cached = False
         if self.is_cached():
-            self.rdd.unpersist()
+            try:
+                self.rdd.unpersist()
+            except Py4JError:
+                print("Try to unpersist an uncached rdd")
         return self
 
     def _uncache(self):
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/orca/data/shard.py
```diff
@@ -145,7 +145,8 @@ def cache(self):
 
     def uncache(self):
         self.user_cached = False
-        self.rdd.unpersist()
+        if self.is_cached():
+            self.rdd.unpersist()
```
jason-dai, line 149: try ... catch ...?

jason-dai:
LGTM

jenniew:
Seems self.user_cached is not used anywhere. We just set self.user_cached. When we want to check if cached, we call is_cached() which is to check self.rdd.is_cached.

jason-dai:
> Seems self.user_cached is not used anywhere. We just set self.user_cached. When we want to check if cached, we call is_cached() which is to check self.rdd.is_cached.

it's used in the `_uncache` method

jenniew:
Sorry I'm a little confusing. I think we don't need self.user_cached flag.  We create cache() and uncache()/_uncache() function. cache() would wrap self.rdd.cache(). uncache()/_uncache() is to check self.rdd.is_cache, then unpersist(). When construct xshards, we call cache() to cache. When need to uncache, call uncache()/_uncache(). Why we need to tell user_cached or not user_cached.

cyita:
> Sorry I'm a little confusing. I think we don't need self.user_cached flag. We create cache() and uncache()/_uncache() function. cache() would wrap self.rdd.cache(). uncache()/_uncache() is to check self.rdd.is_cache, then unpersist(). When construct xshards, we call cache() to cache. When need to uncache, call uncache()/_uncache(). Why we need to tell user_cached or not user_cached.

If the user calls `cache()` on a SparkXShard, this shard will not be automatically uncached after `transform_shard`.

PR #2441, opened 2020-06-11
Support redirecting jvm output to python process (when using notebook)

yangw1234:
Usage:
```python
from zoo import ZooContext, init_nncontext
ZooContext.log_output = True
sc = init_nncontext()  # or init_spark_on_local, init_spark_on_yarn
```
Then training logs will be printed to the current notebook.

image

Tested manually:

- [x] init_nncontext
- [x] init_spark_on_local
- [x] init_spark_on_yarn

### DIFF
--- a/pyzoo/zoo/common/nncontext.py
+++ b/pyzoo/zoo/common/nncontext.py
@@ -27,7 +27,7 @@ import sys
 
 
 def init_spark_on_local(cores=2, conf=None, python_location=None, spark_log_level="WARN",
-                        redirect_spark_log=True, capture_output=False):
+                        redirect_spark_log=True):
     """
     Create a SparkContext with Zoo configuration in local machine.
     :param cores: The default value is 2 and you can also set it to *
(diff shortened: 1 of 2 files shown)


yangw1234:
A similar feature will be available in Spark 3.0

jason-dai:
I think this is very useful @shane-huang 

Maybe it is better to set a global flag, instead of changing the function call? Or maybe set an env variable?

yangw1234:
> I think this is very useful @shane-huang
> 
> Maybe it is better to set a global flag, instead of changing the function call? Or maybe set an env variable?

will change to an env variable.

jason-dai:
> > I think this is very useful @shane-huang
> > Maybe it is better to set a global flag, instead of changing the function call? Or maybe set an env variable?
> 
> will change to an env variable.

Maybe add some global flag? such as `ZooContext.log_output`?

yangw1234:
Hi @jason-dai , do you mean letting users set the flag before init_nncontext()? Such as:

```python
ZooContext.log_output = True
sc = init_nncontext()
```

jason-dai:
> Hi @jason-dai , do you mean letting users set the flag before init_nncontext()? Such as:
> 
> ```python
> ZooContext.log_output = True
> sc = init_nncontext()
> ```

Yes

yangw1234:
> > Hi @jason-dai , do you mean letting users set the flag before init_nncontext()? Such as:
> > ```python
> > ZooContext.log_output = True
> > sc = init_nncontext()
> > ```
> 
> Yes

changed to `ZooContext.log_output = True`

jason-dai:
LGTM

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3588/ jenkins passed

PR #2444, opened 2020-06-12
Orca SparkXShard support broadcast variables in transform_shard

cyita:


pyzoo/zoo/orca/data/shard.py
```diff
@@ -130,8 +130,14 @@ def __init__(self, rdd):
         if self.eager:
             self.compute()
 
-    def transform_shard(self, func, *args):
-        transformed_shard = SparkXShards(self.rdd.map(lambda data: func(data, *args)))
+    def transform_shard(self, func, broadcast_args=False, sc=None, *args):
+        if broadcast_args:
+            assert sc is not None, "Spark context should be provided"
+            broadcast_args_value = tuple(map(lambda arg: sc.broadcast(arg).value, args))
```
qiuxin2012, line 136: you should call `.value`  in `rdd.map`

pyzoo/zoo/orca/data/shard.py
```diff
@@ -133,9 +133,11 @@ def __init__(self, rdd):
     def transform_shard(self, func, broadcast_args=False, sc=None, *args):
         if broadcast_args:
             assert sc is not None, "Spark context should be provided"
-            broadcast_args_value = tuple(map(lambda arg: sc.broadcast(arg).value, args))
+            broadcast_args = tuple(map(lambda arg: sc.broadcast(arg), args))
```
qiuxin2012, line 136: Does this work?
```
            broadcast_args = sc.broadcast(args)
            transformed_shard = SparkXShards(self.rdd.map(lambda data: func(data,
                                                                            *broadcast_args.value)))
```

cyita: It works. Fixed.

pyzoo/zoo/orca/data/shard.py
jenniew: do we need to pass sc or just use self.rdd.ctx?

qiuxin2012: Yes, I think we should use rdd's sc.

jason-dai:
I would prefer something like:

```
def func(df, item_set)
   item_set = item_set.value
   ....

item_set= ...
item_set= orca.data.SharedValue(item_set)
full_data.transform_shard(func, item_set)
```

jason-dai:
Alternatively, we can just broadcast the variables by default; not sure about the performance implications for smaller objects.

pyzoo/zoo/orca/data/shard.py
```diff
@@ -280,3 +281,20 @@ def load_pickle(cls, path, minPartitions=None):
 
     def __del__(self):
         self.uncache()
+
+
+class SharedValue(object):
+    def __init__(self, data, sc):
+        if isinstance(sc, SparkContext):
+            self.broadcast_data = sc.broadcast(data)
```
jason-dai, line 289: just use `sc = init_nncontext()`, similar to https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/data/shard.py#L313

cyita:
Jenkins passed: http://10.239.47.210:18888/view/ZOO-PR-PIPELINE/job/ZOO-PR-Validation/3586/

PR #2459, opened 2020-06-15
Add tensorflow version check

### LINKED ISSUE #2456: TF version checking in TFPark?
Should we automatically check TensorFlow version in TFPark?

Say, 
1. if user installed TF 2.x, we raise an error since it is very likely that it is not going to work
2. if user install TF 1.x but not 1.15.0, we print an warning since it might work on other earlier versions.

yangw1234:
It is not uncommon for our users to install a incorrect version of TensorFlow and face some weird error message. This PR tries to solve this problem.

If users install TF 1.x, they will receive an warning such as
```
######################### WARNING ##########################
Analytics Zoo TFPark has only been tested on TensorFlow 1.15.0, but your current TensorFlow installation is 1.14.0.
##############################################################

```
If users install TF 2.x, they will receive an error, e.g.
```
RuntimeError: Analytics Zoo TFPark only supports TensorFlow 1.15.0, but your current TensorFlow installation is 2.2.0
You can export ANALYTICS_ZOO_TF_CHECK=false to disable this check.
```

Users can export ANALYTICS_ZOO_TF_CHECK=false to disable this check

Fix #2456 

manually tested it

### DIFF
--- a/pyzoo/zoo/tfpark/__init__.py
+++ b/pyzoo/zoo/tfpark/__init__.py
@@ -14,6 +14,32 @@
 # limitations under the License.
 #
 
+
+def check_tf_version():
+    import os
+    import logging
+    do_check = True
+    if "ANALYTICS_ZOO_TF_CHECK" in os.environ and os.environ["ANALYTICS_ZOO_TF_CHECK"] == "false":
+        do_check = False
+    if do_check:
+        import tensorflow as tf
+        v_str = tf.__version__
+        major, minor, patch = v_str.split(".")
+        if v_str != "1.15.0":
+            if int(major) == 1:
+                logging.warning("\n######################### WARNING ##########################\n"
+                                "\nAnalytics Zoo TFPark has only been tested on TensorFlow 1.15.0,"
+                                " but your current TensorFlow installation is {}.".format(v_str) +
+                                "\nYou may encounter some version incompatibility issues. "
+                                "\n##############################################################")
+            else:
+                message = "Analytics Zoo TFPark only supports TensorFlow 1.15.0, " + \
+                          "but your current TensorFlow installation is {}".format(v_str) + \
+                          "\nYou can export ANALYTICS_ZOO_TF_CHECK=false to disable this check."
+                raise RuntimeError(message)
+
+check_tf_version()
+
 from .model import KerasModel
 from .estimator import TFEstimator
 from .tf_optimizer import TFOptimizer
--- a/pyzoo/zoo/tfpark/__init__.py
+++ b/pyzoo/zoo/tfpark/__init__.py
@@ -16,27 +16,25 @@
 
 
 def check_tf_version():
-    import os
     import logging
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/tfpark/__init__.py
```diff
@@ -14,6 +14,32 @@
 # limitations under the License.
 #
 
+
+def check_tf_version():
+    import os
+    import logging
+    do_check = True
+    if "ANALYTICS_ZOO_TF_CHECK" in os.environ and os.environ["ANALYTICS_ZOO_TF_CHECK"] == "false":
```
jason-dai, line 22: use `ZooContext._check_tf_version`?

pyzoo/zoo/tfpark/__init__.py
```diff
@@ -14,6 +14,32 @@
...
+        do_check = False
+    if do_check:
+        import tensorflow as tf
```
jason-dai, line 25: what if tf is not installed?

pyzoo/zoo/tfpark/__init__.py
```diff
@@ -14,6 +14,32 @@
...
+        major, minor, patch = v_str.split(".")
+        if v_str != "1.15.0":
+            if int(major) == 1:
+                logging.warning("\n######################### WARNING ##########################\n"
+                                "\nAnalytics Zoo TFPark has only been tested on TensorFlow 1.15.0,"
+                                " but your current TensorFlow installation is {}.".format(v_str) +
+                                "\nYou may encounter some version incompatibility issues. "
+                                "\n##############################################################")
+            else:
+                message = "Analytics Zoo TFPark only supports TensorFlow 1.15.0, " + \
```
jason-dai, line 36: Currently Analytics Zoo TFPark only supports TensorFlow 1.15.0

pyzoo/zoo/tfpark/__init__.py
```diff
@@ -14,6 +14,32 @@
...
+                          "but your current TensorFlow installation is {}".format(v_str) + \
+                          "\nYou can export ANALYTICS_ZOO_TF_CHECK=false to disable this check."
```
jason-dai, line 38: does it make sense to disable check in this case?

yangw1234: Not much. If not 1.x, I think it would definitely fail. Let's remove the "disable check"?

pyzoo/zoo/tfpark/__init__.py
yangw1234: how about report "Importing TensorFlow failed, please install tensorflow 1.15.0." ?

jason-dai:
LGTM

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3605/ jenkins passed except some unrelated error

PR #2466, opened 2020-06-16
Fix num_partitions in orca spark read

hkvision:
When the number of files is smaller than the total cores, the partition num should be equal to the number of files but not the total cores. Otherwise there would be empty partitions.

### DIFF
--- a/pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
+++ b/pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
@@ -89,10 +89,14 @@ class MXNetTrainer(object):
 
         from zoo.orca.data import RayXShards, SparkXShards
         if isinstance(train_data, SparkXShards):
-            train_data = train_data.repartition(self.num_workers).to_ray()
+            if train_data.num_partitions() != self.num_workers:
+                train_data = train_data.repartition(self.num_workers)
+            train_data = train_data.to_ray()
             if test_data:
                 assert isinstance(test_data, SparkXShards)
-                test_data = test_data.repartition(self.num_workers).to_ray()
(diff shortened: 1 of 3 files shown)


pyzoo/zoo/orca/learn/mxnet/mxnet_trainer.py
```diff
@@ -89,10 +89,14 @@ def __init__(self, config, train_data, model_creator,
 
         from zoo.orca.data import RayXShards, SparkXShards
         if isinstance(train_data, SparkXShards):
-            train_data = train_data.repartition(self.num_workers).to_ray()
+            if train_data.num_partitions() != self.num_workers:
+                train_data = train_data.repartition(self.num_workers)
```
jenniew, line 93: If num_elements is less than num_workers, some partition would be empty, what will the runner do?

hkvision: I think the current assumption is the number of files > the number of workers?
Otherwise, there is no need to start so many workers right?
If needed, we can add an assert for num_workers.

jenniew: I think we cannot guarantee no empty partition after repartition, we may add logic to handle empty partition.

jason-dai: > I think we cannot guarantee no empty partition after repartition, we may add logic to handle empty partition.

We may need to add a function, say, `non_empty`, that takes rdd1 and returns rdd2 which contains all the non-empty partitions in rdd1; then we can apply `non_empty` in `SparkXShards` constructor or in read_csv/read_json.

PR #2473, opened 2020-06-17
[Draft] Use spark.read.csv/json

hkvision:
@jason-dai Take a look if we need this change? Seems toPandas() is not needed actually?
Tested for RBI code locally.

### DIFF
--- a/pyzoo/zoo/orca/data/pandas/preprocessing.py
+++ b/pyzoo/zoo/orca/data/pandas/preprocessing.py
@@ -107,63 +107,31 @@ def read_file_spark(context, file_path, file_type, **kwargs):
     if not file_paths:
         raise Exception("The file path is invalid/empty or does not include csv/json files")
 
-    rdd = context.parallelize(file_paths, node_num * core_num)
+    num_files = len(file_paths)
+    total_cores = node_num * core_num
+    num_partitions = num_files if num_files < total_cores else total_cores
 
-    if prefix == "hdfs":
-        def loadFile(iterator):
-            import pandas as pd
-            import pyarrow as pa
-            fs = pa.hdfs.connect()
-
-            for x in iterator:
-                with fs.open(x, 'rb') as f:
-                    if file_type == "csv":
-                        df = pd.read_csv(f, **kwargs)
-                    elif file_type == "json":
-                        df = pd.read_json(f, **kwargs)
-                    else:
-                        raise Exception("Unsupported file type")
-                    yield df
-
-        pd_rdd = rdd.mapPartitions(loadFile)
-    elif prefix == "s3":
-        def loadFile(iterator):
-            access_key_id = os.environ["AWS_ACCESS_KEY_ID"]
-            secret_access_key = os.environ["AWS_SECRET_ACCESS_KEY"]
-            import boto3
-            import pandas as pd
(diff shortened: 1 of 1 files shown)


hkvision:
Also in this way, there would be no empty partitions?

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -107,63 +107,31 @@ def read_file_spark(context, file_path, file_type, **kwargs):
...
+    num_files = len(file_paths)
+    total_cores = node_num * core_num
+    num_partitions = num_files if num_files < total_cores else total_cores
+
+    from pyspark.sql import SQLContext
+    sqlContext = SQLContext.getOrCreate(context)
+    spark = sqlContext.sparkSession
+    # TODO: add S3 confidentials
+    if file_type == "json":
+        df = spark.read.json(file_paths).repartition(num_partitions)
```
jason-dai, line 119: do we need to `repartition`?

jason-dai:
Can you compare the performance with the previous implementation? It might be more efficient to use arrow to do the conversion, similar to `toPandas`, but it may not be critical at this moment.

pyzoo/zoo/orca/data/pandas/preprocessing.py
hkvision: I'm not sure. How many partitions would spark.read.json have for each file? (When I run locally each file has three partitions..)
Or we can let user specify the number of partitions?
Otherwise, if the num partitions is small, then after reading and converting to a pandas df for each partition, there is no way to split the dataframes?

hkvision:
> Can you compare the performance with the previous implementation? It might be more efficient to use arrow to do the conversion, similar to `toPandas`, but it may not be critical at this moment.

The previous implementation means our current implementation? But arrow only reads the files from hdfs, it doesn't do any conversion?

Or you want to compare with toPandas() and pd.DataFrame?

jason-dai:
> > Can you compare the performance with the previous implementation? It might be more efficient to use arrow to do the conversion, similar to `toPandas`, but it may not be critical at this moment.
> 
> The previous implementation means our current implementation? But arrow only reads the files from hdfs, it doesn't do any conversion?
> 
> Or you want to compare with toPandas() and pd.DataFrame?

Compare with current implementations

hkvision:
> > > Can you compare the performance with the previous implementation? It might be more efficient to use arrow to do the conversion, similar to `toPandas`, but it may not be critical at this moment.
> > 
> > 
> > The previous implementation means our current implementation? But arrow only reads the files from hdfs, it doesn't do any conversion?
> > Or you want to compare with toPandas() and pd.DataFrame?
> 
> Compare with current implementations

Then why you say `more efficient to use arrow to do the conversion`? pyarrow only connects and reads from HDFS right? The performance is spark read csv VS pandas read csv + spark parallelism?

pyzoo/zoo/orca/data/pandas/preprocessing.py
hkvision: compare with node number

hkvision: or we need to support splitting dataframes for repartition

PR #2442, opened 2020-06-11
PytorchHorovodEstimator

yangw1234:
Usage:

```python
trainer1 = PytorchHorovodTrainer(
     model_creator=model_creator,
     data_creator=data_creator,
     optimizer_creator=optimizer_creator,
     loss_creator=nn.MSELoss,
     scheduler_creator=scheduler_creator,
     config={
         "lr": 1e-2,  # used in optimizer_creator
         "hidden_size": 1,  # used in model_creator
         "batch_size": 4,  # used in data_creator
     })
 for i in range(5):
     stats = trainer1.train()
     print(stats)
```

Waiting for #2476 

Testing:
- [ ] Unit tests 
- [x] Yarn cluster
- [x] cifar10 converged at 92.2%

PR #2242, opened 2020-04-22
[Bug Fix] Add local file header to fix keras saving in hdfs yarn mode

yangw1234:

PR #2476, opened 2020-06-18
Refactor HorovodRayTrainer (separated from #2242)

yangw1234:
The refactoring part in #2442 is separated as this pr

Here are the changes:

1. Change horovod example directory to orca/learn
2. Change HorovodRayTrainer to HorovodRayRunner since the main function of this is class is to set up the Horovod environment on each worker, will reserve the `train` or `run` method to HorovodEstimator
3. Make some changes to let future HorovodEstimators extends or use HorovodRayRunner.


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3637/ Jenkins passed

PR #2478, opened 2020-06-18
Orca SparkXShard select

cyita:
Related issue [
Projection (selecting columns) for SparkXShards #54 ](https://github.com/analytics-zoo/orca/issues/54)

### DIFF
--- a/pyzoo/zoo/orca/data/shard.py
+++ b/pyzoo/zoo/orca/data/shard.py
@@ -144,13 +144,14 @@ class RayPartition(object):
 
 
 class SparkXShards(XShards):
-    def __init__(self, rdd, eager=None):
+    def __init__(self, rdd, transient=False):
         self.rdd = rdd
         self.user_cached = False
-        self.eager = eager
-        if self.eager is None:
+        if transient:
+            self.eager = False
+        else:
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/orca/data/shard.py
```diff
@@ -144,10 +144,12 @@ def get_data(self):
...
 class SparkXShards(XShards):
-    def __init__(self, rdd):
+    def __init__(self, rdd, eager=None):
         self.rdd = rdd
         self.user_cached = False
-        self.eager = ZooContext._orca_eager_mode
+        self.eager = eager
+        if self.eager is None:
+            self.eager = ZooContext._orca_eager_mode
         self.rdd.cache()
```
jason-dai, line 153: For selection, we shouldn't cache rdd either. Maybe we can have something like:
```
def __init__(self, rdd, transient=False)
    ...
    if transient == False:
        self.eager = ZooContext._orca_eager_mode
        self.rdd.cache()
    else:
        self.eager = False
    ...
```

jason-dai:
LGTM

cyita:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/3627/

PR #2480, opened 2020-06-18
support bn when TorchModel export model to pure pytorch

qiuxin2012:
support bn when TorchModel export model to pure pytorch

As setExtraParameters is final in BigDL's AbstractModule, so I find a way to walk around.
We can remove this walk around when update to bigdl 0.11.0.(https://github.com/intel-analytics/BigDL/pull/3014 remove the final)


jason-dai:
I think this works for BN; however, why can't we just save the Pytorch model at the executor side (say, on HDFS)? We can also send the bits back to the driver and then use PyTroch to load it if needed.

qiuxin2012:
jenkins passed

PR #2482, opened 2020-06-18
Cluster Serving Inference Model Tensorflow String inputs and UT

Litchilitchy:


jason-dai:
I don't think it makes sense to use the key to indicate the type of the value.

jason-dai:
I don't think it makes sense to use the key to indicate the type of the value. @Litchilitchy

PR #2487, opened 2020-06-19
Remove TF imports 

### LINKED ISSUE #2484: check_tf_version()
* `zoo/pipeline/api/net/__init__.py`  imports TFNet
* `zoo/pipeline/api/net/torch_net.py`  imports zoo.tfpark.tfnet.to_sample_rd

As a result, an error is thrown if TensorFlow is not installed or is TF2 due to https://github.com/intel-analytics/analytics-zoo/pull/2459

We should:
1. move `zoo.tfpark.tfnet.to_sample_rdd` to a `util` package
2. remove "import TFNet" from `zoo/pipeline/api/net/__init__.py`  
2. move the check in https://github.com/intel-analytics/analytics-zoo/pull/2459 to where it is used (e.g., in TFNet, TFOptimizer, etc.)

yangw1234:
fix #2484

### DIFF
--- a/pyzoo/zoo/tfpark/__init__.py
+++ b/pyzoo/zoo/tfpark/__init__.py
@@ -14,34 +14,9 @@
 # limitations under the License.
 #
 
-
-def check_tf_version():
-    import logging
-    try:
-        import tensorflow as tf
-    except Exception as e:
-        raise RuntimeError("Importing TensorFlow failed, please install tensorflow 1.15.0.", e)
-
-    v_str = tf.__version__
-    major, minor, patch = v_str.split(".")
-    if v_str != "1.15.0":
-        if int(major) == 1:
-            logging.warning("\n######################### WARNING ##########################\n"
-                            "\nAnalytics Zoo TFPark has only been tested on TensorFlow 1.15.0,"
-                            " but your current TensorFlow installation is {}.".format(v_str) +
-                            "\nYou may encounter some version incompatibility issues. "
-                            "\n##############################################################")
-        else:
-            message = "Currently Analytics Zoo TFPark only supports TensorFlow 1.15.0, " + \
-                      "but your current TensorFlow installation is {}".format(v_str)
-            raise RuntimeError(message)
-
-check_tf_version()
-
 from .model import KerasModel
 from .estimator import TFEstimator
 from .tf_optimizer import TFOptimizer
 from .tf_dataset import TFDataset
-from .zoo_optimizer import ZooOptimizer
 from .tf_predictor import TFPredictor
 from .tfnet import TFNet
--- a/pyzoo/zoo/tfpark/__init__.py
+++ b/pyzoo/zoo/tfpark/__init__.py
@@ -14,9 +14,68 @@
 # limitations under the License.
 #
 
-from .model import KerasModel
-from .estimator import TFEstimator
-from .tf_optimizer import TFOptimizer
-from .tf_dataset import TFDataset
-from .tf_predictor import TFPredictor
-from .tfnet import TFNet
+
+def check_tf_version():
+    import logging
+    try:
+        import tensorflow as tf
+    except Exception as e:
+        return False, RuntimeError("Importing TensorFlow failed, please install tensorflow 1.15.0.", e)
+
+    v_str = tf.__version__
+    major, minor, patch = v_str.split(".")
+    if v_str != "1.15.0":
+        if int(major) == 1:
+            logging.warning("\n######################### WARNING ##########################\n"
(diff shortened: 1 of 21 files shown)


pyzoo/zoo/tfpark/estimator.py
```diff
@@ -95,11 +94,14 @@ def from_model_fn(cls, model_fn, model_dir=None, config=None,
                        warm-started, and it is assumed that vocabularies
                        and `tf.Tensor` names are unchanged.
         """
+        check_tf_version()
```
jason-dai, line 97: no need to check here, as it will be checked in `TFEstimator`?

pyzoo/zoo/tfpark/model.py
```diff
@@ -82,6 +78,8 @@ def load_model(path):
         :param path: String. The path to the pre-defined model.
         :return: KerasModel.
         """
+        check_tf_version()
```
jason-dai, line 81: no need to check here, as it will be checked in `KerasModel`

pyzoo/zoo/tfpark/utils.py
```diff
@@ -17,7 +17,27 @@
...
 from zoo.pipeline.api.keras import metrics as zmetrics
 
-import tensorflow as tf
+
+def check_tf_version():
+    import logging
+    try:
+        import tensorflow as tf
+    except Exception as e:
+        raise RuntimeError("Importing TensorFlow failed, please install tensorflow 1.15.0.", e)
```
jason-dai, line 26: instead of existing, maybe just `log.error` instead

pyzoo/zoo/tfpark/utils.py
```diff
@@ -17,7 +17,27 @@
...
+        if int(major) == 1:
+            logging.warning("\n######################### WARNING ##########################\n"
+                            "\nAnalytics Zoo TFPark has only been tested on TensorFlow 1.15.0,"
+                            " but your current TensorFlow installation is {}.".format(v_str) +
+                            "\nYou may encounter some version incompatibility issues. "
+                            "\n##############################################################")
+        else:
+            message = "Currently Analytics Zoo TFPark only supports TensorFlow 1.15.0, " + \
+                      "but your current TensorFlow installation is {}".format(v_str)
+            raise RuntimeError(message)
```
jason-dai, line 40: instead of existing, maybe just `log.error` instead

pyzoo/zoo/tfpark/tf_dataset.py
```diff
@@ -371,6 +374,7 @@ def from_rdd(*args, **kwargs):
                         when training
         :return: a TFDataset
         """
+        check_tf_version()
```
jason-dai, line 377: instead of checking here, can we check in `TFDataset.__init__` instead?

yangw1234: I think check_tf_version should be called as earlier as possible? If any tf code has been called before check_tf_version, there is a risk of encountering incorrect version issue.

jason-dai: But does it matter? If we will report error eventually, I think it is fine.

yangw1234: User program may fail and exit before `check_tf_version` get called?

pyzoo/zoo/tfpark/utils.py
yangw1234: many logs will be printed at the beginning, I am afraid this will never get noticed.

pyzoo/zoo/tfpark/__init__.py
```diff
@@ -34,14 +34,48 @@ def check_tf_version():
...
+passed, error = check_tf_version()
 
-from .model import KerasModel
-from .estimator import TFEstimator
-from .tf_optimizer import TFOptimizer
-from .tf_dataset import TFDataset
-from .zoo_optimizer import ZooOptimizer
-from .tf_predictor import TFPredictor
-from .tfnet import TFNet
+if passed:
```
yangw1234, line 42: @jason-dai how about this? If tf is not installed or installed incorrectly, error will be thrown when user use the following object to do anything.

jason-dai: LGTM

yangw1234:
jenkins pass http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3676/

yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3686/

PR #2501, opened 2020-06-24
Cluster Serving pub sub schema and doc

Litchilitchy:


docs/docs/ClusterServingGuide/APIGuide.md
```diff
@@ -37,14 +37,23 @@ Import the dependency and create an instance of `InputQueue`
 from zoo.serving.client import InputQueue
 input_api = InputQueue()
 ```
-To enqueue an list of string, specify key as `string*` and pass a list of str objects, list of str type input is usually used in Tensorflow models.
+To enqueue an list of string, pass a list of str objects, list of str type input is usually used in Tensorflow models.
```
qiyuangong, line 40: an -> a

Litchilitchy: noted

PR #2504, opened 2020-06-26
Orca tf estimator with Keras model

jenniew:
Add Keras model support for Orca tf estimator.


pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -229,3 +239,37 @@ def evaluate(self, data, batch_size=32):
         return evaluate_metrics(flat_inputs + flat_labels,
                                 sess=self.sess,
                                 dataset=dataset, metrics=self.metrics)
+
+
+class TFKerasWrapper(Estimator):
+
+    def __init__(self, keras_model, model_dir=None, optim_method=None, session_config=None):
```
jason-dai, line 246: why do we need optim_method here?

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -229,3 +239,37 @@ def evaluate(self, data, batch_size=32):
...
+            epochs=1,
+            validation_data=None,
+            **kwargs):
+
+        train_dataset = _to_dataset(data, batch_size=batch_size, batch_per_thread=-1,
+                                    validation_data=validation_data)
+
+        self.model.fit(train_dataset, batch_size=batch_size, epochs=epochs,
+                       optim_method=self.optim_method, session_config=self.session_config,
+                       **kwargs)
```
jason-dai, line 262: I don't understand - `KerasModel.fit` has no optim_method or session_config arguments?

pyzoo/zoo/orca/learn/tf/estimator.py
jenniew: In the TFOptimizer.from_keras( keras_model, dataset, optim_method=None, session_config=None, model_dir=None), it has optim_method parameter. In the implementation, if user specify this param, we would use this optim_method instead of optimizer in keras model. Should we remove this option in TFOptimizer.from_keras()?

jason-dai: > In the TFOptimizer.from_keras( keras_model, dataset, optim_method=None, session_config=None, model_dir=None), it has optim_method parameter. In the implementation, if user specify this param, we would use this optim_method instead of optimizer in keras model. Should we remove this option in TFOptimizer.from_keras()?

I think we should use built-in TF or Keras optimizers, not BigDL optimizers @yangw1234

pyzoo/zoo/orca/learn/tf/estimator.py
jenniew: KerasModel.fit(x=None, y=None, batch_size=None, epochs=1,validation_data=None, distributed=False,        **kwargs). But KerasModel.fit would call  TFOptimizer.from_keras(self.model, dataset,                                                  model_dir=self.model_dir, **kwargs). And TFOptimizer.from_keras(keras_model, dataset, optim_method=None, session_config=None, model_dir=None) has these two options. If we remove them in estimator, then we don't give option to change optim_method or session_config.

pyzoo/zoo/orca/learn/tf/estimator.py
jenniew: remove optim_method option, and align APIs for graph and keras

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -243,23 +247,23 @@ def evaluate(self, data, batch_size=32):
...
-        self.optim_method = optim_method
         self.session_config = session_config
 
     def fit(self, data,
-            batch_size=None,
+            batch_size=32,
             epochs=1,
             validation_data=None,
-            **kwargs):
+            feed_dict=None
```
jason-dai, line 258: do you need feed_dict here?

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -22,7 +22,6 @@
 import tensorflow as tf
 
 from zoo.tfpark.tf_dataset import TFDataset
-from zoo.tfpark.tf_optimizer import TFModel
 from zoo.tfpark import KerasModel
 from zoo.util import nest
```
jason-dai, line 27: `Estimator.fit` should use epochs instead of steps

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -229,3 +243,37 @@ def evaluate(self, data, batch_size=32):
...
+
+        train_dataset = _to_dataset(data, batch_size=batch_size, batch_per_thread=-1,
+                                    validation_data=validation_data)
+
+        self.model.fit(train_dataset, batch_size=batch_size, epochs=epochs,
+                       session_config=self.session_config
+                       )
+        return self
+
+    def predict(self, data, batch_size=32):
```
jason-dai, line 269: maybe 32 is too large for `batch_per_thread`

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -229,3 +243,37 @@ def evaluate(self, data, batch_size=32):
...
+
+        dataset = _to_dataset(data, batch_size=-1, batch_per_thread=batch_size)
+
+        return self.model.predict(dataset, batch_size)
+
+    def evaluate(self, data, batch_size=32):
```
jason-dai, line 275: maybe 32 is too large for `batch_per_thread`

pyzoo/zoo/orca/learn/tf/estimator.py
jenniew: No, feed_dict in only for graph use

jason-dai: I think it is only needed for `fit` for tf compute graph? Then it makes sense to keep it in `fit` of `TFOptimizerWrapper `. @yangw1234

pyzoo/zoo/orca/learn/tf/estimator.py
yangw1234: I‘ll try to support
1. tf optimizers in KerasModel and TFOptimizer.from_keras
2. keras.Optimizers

pyzoo/zoo/orca/learn/tf/estimator.py
yangw1234: this might be used in evaluate and predict, but there is no implementation yet. Let me think about this.

jason-dai:
BTW, we should make sure the `fit` APIs are also consistent with MXNet estimator. Maybe it's reasonable to make it

`def fit(self, train_data, epochs=1, batch_size=32, val_data=None, ...)`

BTW, why no batch_size is needed for MXNet? @hkvision

hkvision:
> BTW, we should make sure the `fit` APIs are also consistent with MXNet estimator. Maybe it's reasonable to make it
> 
> `def fit(self, train_data, epochs=1, batch_size=32, val_data=None, ...)`
> 
> BTW, why no batch_size is needed for MXNet? @hkvision

Currently batch_size is specified in `config` in MXNet. I can make it a separate argument to align with other estimators if necessary.

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -185,17 +181,14 @@ def __init__(self, *, inputs, outputs, labels, loss,
...
-        else:
-            self.tensor_with_value = None
 
     def fit(self, data,
             epochs=1,
             batch_size=32,
-            validation_data=None
+            validation_data=None,
+            feed_dict=None,
+            session_config=None
```
jason-dai, line 191: session_config before feed_dict?

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -250,21 +248,22 @@ def evaluate(self, data, batch_size=4):
...
         self.model = KerasModel(keras_model, model_dir)
-        self.session_config = session_config
 
     def fit(self, data,
             epochs=1,
             batch_size=32,
-            validation_data=None
+            validation_data=None,
+            feed_dict=None,
+            session_config=None
```
jason-dai, line 259: remove feed_dict for Keras?

pyzoo/zoo/orca/learn/tf/estimator.py
jason-dai, line 191: session_config before feed_dict?

pyzoo/zoo/orca/learn/tf/estimator.py
jenniew: do we keep the same api as fit for graph model?

pyzoo/zoo/orca/learn/tf/estimator.py

pyzoo/zoo/orca/learn/tf/estimator.py
jason-dai: does it make sense to keep it? if not, we don't need to support it for Keras

jason-dai:
> > BTW, we should make sure the `fit` APIs are also consistent with MXNet estimator. Maybe it's reasonable to make it
> > `def fit(self, train_data, epochs=1, batch_size=32, val_data=None, ...)`
> > BTW, why no batch_size is needed for MXNet? @hkvision
> 
> Currently batch_size is specified in `config` in MXNet. I can make it a separate argument to align with other estimators if necessary.

Yes, we should align the interface of all these estmators

pyzoo/zoo/orca/learn/tf/estimator.py
jenniew: No need for fit for Keras model. Removed.

jason-dai:
LGTM

hkvision:
> > > BTW, we should make sure the `fit` APIs are also consistent with MXNet estimator. Maybe it's reasonable to make it
> > > `def fit(self, train_data, epochs=1, batch_size=32, val_data=None, ...)`
> > > BTW, why no batch_size is needed for MXNet? @hkvision
> > 
> > 
> > Currently batch_size is specified in `config` in MXNet. I can make it a separate argument to align with other estimators if necessary.
> 
> Yes, we should align the interface of all these estimators

Where are other parameters such as optim method in tf estimator?

jenniew:
Jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3729/

PR #2507, opened 2020-06-28
Fix UT random fails

pinggao18:
UT random fails due to small tolerance while comparing array values


pinggao18:
jenkins test :  http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3660/

PR #2512, opened 2020-06-29
Generate object ids for plasma error handling

hkvision:
Previously if putting into plasma tasks fail, each spark retrial would create a new object id to put the data. This would lead to memory leak if only some tasks fail.

Generate an ObjectID for each partition beforehand for plasma.
In case tasks fail, spark would retry to put into plasma. If the target id already exists, we suppose that the task on that partition succeeds and does nothing. Otherwise, still use this id to put into plasma.

### DIFF
--- a/pyzoo/zoo/orca/data/shard.py
+++ b/pyzoo/zoo/orca/data/shard.py
@@ -333,14 +333,20 @@ class SparkXShards(XShards):
                 res = list(iterator)
                 client = plasma.connect(object_store_address)
                 target_id = ids[index]
+                # If the ObjectID exists in plasma, we assume a task trial
+                # succeeds and the data is already in the object store.
                 if not client.contains(target_id):
                     object_id = client.put(res, target_id)
-                    assert object_id == target_id
+                    assert object_id == target_id, \
+                        "Error occurred when putting data into plasma object store"
                 client.disconnect()
                 yield target_id, get_node_ip()
             return f
 
-        # Create plasma ObjectIDs beforehand instead of creating a random one every time.
+        # Create plasma ObjectIDs beforehand instead of creating a random one every time to avoid
+        # memory leak in case errors occur when putting data into plasma and Spark would retry.
+        # ObjectIDs in plasma is a byte string of length 20 containing characters and numbers.
+        # The random generation of ObjectIDs is often good enough to ensure unique IDs.
         import pyarrow.plasma as plasma
         object_ids = [plasma.ObjectID.from_random() for i in range(self.rdd.getNumPartitions())]
         object_id_node_ips = self.rdd.mapPartitionsWithIndex(put_to_plasma(object_ids)).collect()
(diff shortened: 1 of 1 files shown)


hkvision:
What about this? @jason-dai @yangw1234

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3683/

pyzoo/zoo/orca/data/shard.py
```diff
@@ -322,33 +322,28 @@ def get_data(data):
...
 
-        # Generate a random string here to make sure that when this method is called twice, the
-        # seeds to generate plasma ObjectID are different.
-        random_str = ''.join(
-            [random.choice(string.ascii_letters + string.digits) for i in range(32)])
-        object_id_node_ips = self.rdd.mapPartitionsWithIndex(put_to_plasma(random_str)).collect()
+        # Create plasma ObjectIDs beforehand instead of creating a random one every time.
+        import pyarrow.plasma as plasma
+        object_ids = [plasma.ObjectID.from_random() for i in range(self.rdd.getNumPartitions())]
+        object_id_node_ips = self.rdd.mapPartitionsWithIndex(put_to_plasma(object_ids)).collect()
```
jason-dai, line 346: Is it possible to generate ID that already exists in the executor side?

hkvision: Can you give more details? Originally on each node, there would be no IDs. If some tasks succeed, there will be IDs and can be retrieved.

jason-dai: How do you make sure the generated ID does not already exist in the plasma?

hkvision: Object ID is a byte string of length 20 that consists of numbers and lower-letter chars. So I suppose there is little probability to generate two same random IDs?

jason-dai: How does each previously plasma id get generated in each partition?

hkvision: Previously if we don't specify the ID, plasma itself would generate a random ID for the data. Now if we specify the ID, plasma would exactly use the id we input.

jason-dai: > Previously if we don't specify the ID, plasma itself would generate a random ID for the data. Now if we specify the ID, plasma would exactly use the id we input.

How does plasma generate the ID?

hkvision: I suppose plasma would also randomly generate the ID of the expected format.

hkvision:
Synced with @jason-dai offline. Would merge it first. If there are further issue, would fix in another PR.

hkvision:
jenkins http://10.239.47.210:18888/job/ZOO-PR-Validation/3705/
One failure in scala is irrelevant.

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3707/

PR #2519, opened 2020-06-30
Orca tf estimator for Spark DataFrame

jenniew:
Add Spark DataFrame support for Orca tf estimator for both graph and keras model.

Related issue: https://github.com/analytics-zoo/orca/issues/57


pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -123,19 +127,38 @@ def flatten(data):
...
+                                 sequential_order=sequential_order,
+                                 shuffle=shuffle)
 
     return dataset
 
 
-def _to_dataset(data, batch_size, batch_per_thread):
+def _to_dataset(data, batch_size, batch_per_thread, validation_data,
+                feature_cols, labels_cols, hard_code_batch_size,
+                sequential_order, shuffle):
```
jason-dai, line 140: need to make sure `data` and `validation_data` have the same type

jenniew: updated

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -123,21 +127,44 @@ def flatten(data):
...
+    if validation_data:
+        assert type(data) == type(validation_data), \
```
jason-dai, line 142: It's possible both data and validation_data are both sub-classes of , say, `SparkXShards`, but have different types?

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -51,7 +52,10 @@ def from_graph(*, inputs, outputs=None,
 
 def _xshards_to_tf_dataset(data_shard,
```
jason-dai, line 53: maybe move `_xshards_to_tf_dataset` and `_to_dataset` to a different file

pyzoo/zoo/orca/learn/tf/estimator.py

pyzoo/zoo/orca/learn/tf/estimator.py

jenniew:
support prediction result as dataframe. 
Related issue: https://github.com/intel-analytics/analytics-zoo/issues/2546

pyzoo/zoo/orca/learn/tf/utils.py
```diff
@@ -0,0 +1,166 @@
...
+def convert_predict_to_dataframe(df, prediction_rdd):
+    from pyspark.sql.types import StructType, StructField, FloatType
+    from pyspark.ml.linalg import VectorUDT, Vectors
+
+    def combine(pair):
+        # scalar
+        if len(pair[1].shape) == 0:
+            return dict(list(pair[0].asDict().items()) + [("prediction", float(pair[1].tolist()))]), FloatType()
+        else:
+            return dict(list(pair[0].asDict().items()) + [("prediction", Vectors.dense(pair[1]))]), VectorUDT()
```
jason-dai, line 159: instead of using dictionary, I think you can simple generate RDD[Row]:

 ```return pair[0]('prediction')=pair[1]```

jenniew: change to Row.

pyzoo/test/zoo/orca/learn/spark/test_estimator_for_spark.py
```diff
@@ -37,184 +41,262 @@ def __init__(self):
...
-        result = {
-            "x": (df['user'].to_numpy(), df['item'].to_numpy()),
-        }
-        return result
-
-    data_shard = data_shard.transform_shard(transform)
-    dataset = Dataset.from_tensor_slices(data_shard)
-    predictions = est.predict(dataset).collect()
-    assert len(predictions) == 10
+class TestEstimatorForGraph(TestCase):
```
yangw1234, line 44: Any reasons to change these? Subclassing does not work well with fixtures.

jenniew: I want to use Testcase.assertRaises(Exception) to assert exception. And we don't need to pass fixture anymore, because zoo.orca.data.pandas.read_csv(file_path, sc) is changed to zoo.orca.data.pandas.read_csv(file_path) now and no need sc explicitly.

jenniew:
Jenkins: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3774/
Only some examples are not passed but not related with this PR.

jenniew:
@jason-dai @yangw1234 What else do we need to change for this PR?

jason-dai:
LGTM

PR #2525, opened 2020-07-02
Fix Orca read file path

### LINKED ISSUE #2521: Orca read file sometimew would report path not exist if don't specify file:/// for local file
When I run the Orca pandas read file code on Almaren node, for local file path, I must specify `file:////opt/work/...`, otherwise the error is thrown:
```
2020-07-01 16:53:32 WARN  Utils$:111 - /opt/work/client/kai/rbi_data/sample_data/train doesn't exist!
Traceback (most recent call last):
  File "pandas_backend.py", line 23, in <module>
    orient='columns', lines=True)
  File "/opt/work/client/anaconda3/envs/mxnet-0.8/lib/python3.6/site-packages/zoo/orca/data/pandas/preprocessing.py", line 57, in read_json
    return read_file_spark(context, file_path, "json", **kwargs)
  File "/opt/work/client/anaconda3/envs/mxnet-0.8/lib/python3.6/site-packages/zoo/orca/data/pandas/preprocessing.py", line 109, in read_file_spark
    raise Exception("The file path is invalid/empty or does not include csv/json files")
Exception: The file path is invalid/empty or does not include csv/json files
```
I suppose for most cases just inputting the normal local file path works.
But for the cases that need to specify file:/// to work, the error message is misleading. Shall we handle this? Or we can write some documentations for this.

hkvision:
- Fix #2521 do not use Scala HadoopFS API to list local file paths.
- Do not filter files by its suffix as neither Spark or Pandas does this. We just pass all the files to Spark or Pandas and let them handle or raise an error. For pandas during read time error would be raised if the file format is wrong. For spark seems it can read anything, but when you do transformation on the resulting DataFrame, the error would be raised if the file format is not as expected.

### DIFF
--- a/pyzoo/test/zoo/orca/data/test_pandas_backend.py
+++ b/pyzoo/test/zoo/orca/data/test_pandas_backend.py
@@ -40,7 +40,7 @@ class TestSparkXShards(TestCase):
         file_path = os.path.join(self.resource_path, "abc")
         with self.assertRaises(Exception) as context:
             xshards = zoo.orca.data.pandas.read_csv(file_path)
-        self.assertTrue('The file path is invalid/empty' in str(context.exception))
+        self.assertTrue('No such file or directory' in str(context.exception))
 
     def test_read_local_json(self):
         ZooContext.orca_pandas_read_backend = "pandas"
--- a/pyzoo/test/zoo/orca/data/test_pandas_backend.py
+++ b/pyzoo/test/zoo/orca/data/test_pandas_backend.py
@@ -41,6 +41,11 @@ class TestSparkXShards(TestCase):
         with self.assertRaises(Exception) as context:
             xshards = zoo.orca.data.pandas.read_csv(file_path)
         self.assertTrue('No such file or directory' in str(context.exception))
+        file_path = os.path.join(self.resource_path, "image3d")
+        with self.assertRaises(Exception) as context:
+            xshards = zoo.orca.data.pandas.read_csv(file_path)
+        # This error is raised by pandas.errors.ParserError
+        self.assertTrue('Error tokenizing data' in str(context.exception))
 
     def test_read_local_json(self):
         ZooContext.orca_pandas_read_backend = "pandas"
--- a/pyzoo/zoo/orca/data/pandas/preprocessing.py
+++ b/pyzoo/zoo/orca/data/pandas/preprocessing.py
@@ -46,11 +46,11 @@ def read_json(file_path, **kwargs):
 
 def read_file_spark(file_path, file_type, **kwargs):
     sc = init_nncontext()
+    node_num, core_num = get_node_and_core_number()
 
     if ZooContext.orca_pandas_read_backend == "pandas":
         file_url_splits = file_path.split("://")
         prefix = file_url_splits[0]
-        node_num, core_num = get_node_and_core_number()
 
         file_paths = []
         if isinstance(file_path, list):
@@ -141,12 +141,12 @@ def read_file_spark(file_path, file_type, **kwargs):
                 comment = kwargs["comment"]
                 if not isinstance(comment, str) or len(comment) != 1:
                     raise ValueError("Only length-1 comment characters supported")
-            df = spark.read.csv(file_paths, **kwargs)
+            df = spark.read.csv(file_path, **kwargs)
             if header is None:
                 df = df.selectExpr(
                     *["`%s` as `%s`" % (field.name, i) for i, field in enumerate(df.schema)])
         else:
-            df = spark.read.json(file_paths, **kwargs)
+            df = spark.read.json(file_path, **kwargs)
 
         # Handle pandas-compatible postprocessing arguments
         if isinstance(names, list):
--- a/pyzoo/zoo/orca/data/pandas/preprocessing.py
+++ b/pyzoo/zoo/orca/data/pandas/preprocessing.py
@@ -46,20 +46,21 @@ def read_json(file_path, **kwargs):
 
 def read_file_spark(file_path, file_type, **kwargs):
     sc = init_nncontext()
-    file_url_splits = file_path.split("://")
-    prefix = file_url_splits[0]
-    node_num, core_num = get_node_and_core_number()
-
-    file_paths = []
-    if isinstance(file_path, list):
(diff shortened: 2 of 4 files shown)


pyzoo/zoo/orca/data/utils.py
```diff
@@ -59,35 +59,33 @@ def list_s3_file(file_path, file_type, env):
...
-            # only get json/csv files
-            file_paths = [file for file in file_paths
-                          if os.path.splitext(file)[1] == "." + file_type]
+            from os import listdir
+            from os.path import isfile, abspath, join, splitext
+            # An error would be already raised here if the path is invalid.
+            file_paths = [abspath(join(file_path, file)) for file in listdir(file_path)]
+    # Only get json/csv files.
+    file_paths = [file for file in file_paths
+                  if os.path.isfile(file) and os.path.splitext(file)[1] == "." + file_type]
```
jenniew, line 88: This would not work on mastercard case. The input is one *.data file, so my original logic is not to check suffix if path is a file. For directory, I filter suffix to remove unrelated files. Actually pandas does not check file suffix. We may change logic to not filter but skip the pandas read result if pandas return null or throw exception. But this may cause some partition empty.

hkvision: But your previous logic is wrong, you are only focusing on the mastercard's single case but not general. What about users have  a directory of .dat files? Then you still filter all of them and get nothing.

hkvision: Since the function name is read_csv, it is reasonable that all the input files should be .csv. We can add read_dat if necessary. What do you think? @jason-dai

jenniew: I know your concern. My original design is assuming the file suffix is csv.  But it cannot apply to mastercard case, so I make this workaround.  Pandas has no constraint for file name.  Maybe let us check what spark do. I think read_dat is not necessary, as data file can store different formats.

jason-dai: what's the format of the .dat file?

jenniew: .dat file can store text, image, video, binary, etc. No fix format.

jason-dai: > .dat file can store text, image, video, binary, etc. No fix format.

If the file is not cvs format, what's the behavior of `pd.read_csv`?

jenniew: It would throw exception. Do we remove the filter for filenames and throw exception if pandas fails?

hkvision: OK. I checked mastercard code, it uses the ml-1m dataset which are `.dat` files.
Seems pd.read_csv can only support reading one file; if I input a directory it directly raises an exception. If it can't parse the file, it also raises an exception.
But we are supporting multiple files, just throwing an exception if one file fails may not be reasonable...

jenniew: Or maybe we can do something like ImageProcessing, provide a flag to ignore reading error. If user choose to ignore, we may only give warning message. If not choose ignore, throw exception. There may be possible that user won't like to continue if the some data is incorrect. So give the option may satisfy different requirements.

hkvision: But as you mentioned before, how to handle possible empty partitions? We still need to add a remove empty method?

jenniew: For the functionality, I think our code can work even there are empty partitions. I checked SparkXShards operations, only partition_by() and to_ray() use rdd.mapPartitions(). Others use rdd.map(). rdd.map() can work even there are empty partitions. In partition_by() operation, it first calls rdd.partitionBy(), which rearranges the elements almost evenly to partitions so the empty partitions would be removed, then calls rdd.mapPartitions(). For to_ray() operation, I think we may often call xshards.repartition() before to_ray(), repartition would  re-dispatch elements too, so to_ray() may not store empty partition data to plasma. For the performance, there may have some influence.

hkvision:
I tested that `spark.read.csv` can accept a folder and will read all the files under the folder. But surprisingly it can read any file, including images, but the result could be wrong or messy.
My opinion is to filter csv files only to play safe since seems both pandas and spark have no guarantee that the result is correct given any file type. @jason-dai @jenniew

hkvision:
@jenniew Take a look. Removed the logic of filtering files. Let pandas or spark to handle and report the error. As spark.read.csv/json already accepts multiple files or a folder, the file extraction logic only applies for pandas backend.

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -46,20 +46,21 @@ def read_json(file_path, **kwargs):
...
+        file_paths = []
+        if isinstance(file_path, list):
+            [file_paths.extend(extract_one_path(path, os.environ)) for path in file_path]
+        else:
+            file_paths = extract_one_path(file_path, os.environ)
 
-    if not file_paths:
-        raise Exception("The file path is invalid/empty or does not include csv/json files")
+        if not file_paths:
+            raise Exception("The file path is invalid/empty or does not include csv/json files")
```
jenniew, line 62: raise Exception("The file path is invalid/empty). As we don't filter csv/json files, we cannot tell if the folder include csv/json files.

pyzoo/test/zoo/orca/data/test_pandas_backend.py
```diff
@@ -40,7 +40,7 @@ def test_read_local_csv(self):
         file_path = os.path.join(self.resource_path, "abc")
         with self.assertRaises(Exception) as context:
             xshards = zoo.orca.data.pandas.read_csv(file_path)
-        self.assertTrue('The file path is invalid/empty' in str(context.exception))
+        self.assertTrue('No such file or directory' in str(context.exception))
```
jenniew, line 43: Can you add negative test of file/folder is/contains invalid csv/json file?

hkvision: Added. Take a look.

jenniew:
LGTM.

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3786/

hkvision:
Thanks for your review @qiyuangong @jenniew

PR #2527, opened 2020-07-02
Remove reading file to RayXShards

hkvision:
Fix https://github.com/analytics-zoo/orca/issues/55 the second issue.

Now RayXShards is only a connector for Ray-based Estimators or Trainers.
read_csv/json only Spark only and thus I remove the context argument, get the sc inside the function instead.
All preprocessing/transformation steps should be operated on Spark.
RayXShards should be transparent to users.


hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3709/

hkvision:
> Does RayXShards.colocate_actors(self, actors) only work for MXNetRunner actor? If we want to support for general actors, we may assert len(actor) == len(self.partitions). And actor.get_node_ip() may not exist, we may use ray.services.get_node_ip_address()?

This is supposed to be general but currently only tested on MXNetRunner. Would add support if other ray estimators are ready.
ray.services.get_node_ip_address only works on the current node, how can you get the ip of the remote actor? We should suppose each actor has a get ip function, otherwise we are not able to do this.

Anyway, this is not related to this PR. You can raise a new issue if necessary.

jenniew:
LGTM

hkvision:
Thanks for your review @jenniew

Issue #2522, opened 2020-07-01
spark.read.json on Almaren node would throw java.lang.IllegalArgumentException: Illegal pattern component: XXX


hkvision:
I'm testing spark backend for reading json files. It works properly on my desktop. But when I run on Almaren node, I get the following error:
```
py4j.protocol.Py4JJavaError: An error occurred while calling o60.json.
: java.lang.IllegalArgumentException: Illegal pattern component: XXX
	at org.apache.commons.lang3.time.FastDatePrinter.parsePattern(FastDatePrinter.java:282)
	at org.apache.commons.lang3.time.FastDatePrinter.init(FastDatePrinter.java:149)
	at org.apache.commons.lang3.time.FastDatePrinter.<init>(FastDatePrinter.java:142)
	at org.apache.commons.lang3.time.FastDateFormat.<init>(FastDateFormat.java:384)
	at org.apache.commons.lang3.time.FastDateFormat.<init>(FastDateFormat.java:369)
	at org.apache.commons.lang3.time.FastDateFormat$1.createInstance(FastDateFormat.java:91)
	at org.apache.commons.lang3.time.FastDateFormat$1.createInstance(FastDateFormat.java:88)
	at org.apache.commons.lang3.time.FormatCache.getInstance(FormatCache.java:82)
	at org.apache.commons.lang3.time.FastDateFormat.getInstance(FastDateFormat.java:165)
	at org.apache.spark.sql.catalyst.json.JSONOptions.<init>(JSONOptions.scala:87)
	at org.apache.spark.sql.catalyst.json.JSONOptionsInRead.<init>(JSONOptions.scala:133)
	at org.apache.spark.sql.catalyst.json.JSONOptionsInRead.<init>(JSONOptions.scala:139)
	at org.apache.spark.sql.execution.datasources.json.JsonFileFormat.inferSchema(JsonFileFormat.scala:55)
	at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$6.apply(DataSource.scala:180)
	at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$6.apply(DataSource.scala:180)
	at scala.Option.orElse(Option.scala:289)
	at org.apache.spark.sql.execution.datasources.DataSource.getOrInferFileFormatSchema(DataSource.scala:179)
	at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:373)
[...]

PR #2530, opened 2020-07-02
fix driver classpath issue in init_spark_on_local

jenniew:
Change drive classpath from sparkconf to submit option.

Related issue: https://github.com/intel-analytics/analytics-zoo/issues/2522

### DIFF
--- a/pyzoo/zoo/util/spark.py
+++ b/pyzoo/zoo/util/spark.py
@@ -17,7 +17,7 @@
 import os
 import glob
 
-from pyspark import SparkContext
+from bigdl.util.common import *
 from zoo.common.nncontext import init_spark_conf
 from zoo import init_nncontext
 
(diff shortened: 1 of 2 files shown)


hkvision:
In this way, only `init_spark_on_local` works but `init_nncontext` doesn't work.

jenniew:
> In this way, only `init_spark_on_local` works but `init_nncontext` doesn't work.

Move set submit opts to init_spark_conf

pyzoo/zoo/util/spark.py
```diff
@@ -17,7 +17,7 @@
 import os
 import glob
 
-from pyspark import SparkContext
+from bigdl.util.common import *
```
hkvision, line 20: Then we don't need to change this file.

hkvision:
Need to manually test on server before merging this.

jenniew:
> Need to manually test on server before merging this.

I already tested it on almaren with init_nncontext and init_spark_on_local.

pyzoo/zoo/util/spark.py
jenniew: updated

jenniew:
Jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3713/

PR #2534, opened 2020-07-05
TFOptimizer.from_keras using keras optimizers

yangw1234:


pyzoo/zoo/tfpark/zoo_optimizer.py
```diff
@@ -27,6 +27,44 @@ def __init__(self):
...
+        grads = gradients.gradients(loss, params)
+
+        all_reduced_grads = []
+        for grad, param in zip(grads, params):
+            if grad is None:
+                raise ValueError("Variable {} has `None` for gradient. "
+                                 "Please make sure that all of your ops have a "
+                                 "gradient defined (i.e. are differentiable). "
+                                 "Common ops without gradient: "
+                                 "K.argmax, K.round, K.eval.".format(param))
```
jason-dai, line 48: Is it reasonable for users to use these op?

yangw1234: This error message is copied from `keras.optimizers.Optimizer.get_gradients()`. I think it is to remind user **not** to use ops without gradients, such `argmax`, `round`.

yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3725/

PR #2535, opened 2020-07-06
Cluster Serving dynamic loading and fix

Litchilitchy:
temporary solution for dynamic loading and some other fix for nightly version

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkRedisSource.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkRedisSource.scala
@@ -76,7 +76,8 @@ class FlinkRedisSource(params: SerParams) extends RichSourceFunction[List[(Strin
       RedisUtils.checkMemory(jedis, 0.6, 0.5)
 
     }
-    if (FileUtils.checkStop()) {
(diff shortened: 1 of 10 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkRedisSource.scala
```diff
@@ -76,7 +76,8 @@ class FlinkRedisSource(params: SerParams) extends RichSourceFunction[List[(Strin
       RedisUtils.checkMemory(jedis, 0.6, 0.5)
 
     }
-    if (FileUtils.checkStop()) {
+    val lastModified = FileUtils.getLastModified(params.modelDir)
+    if (FileUtils.checkStop() || FileUtils.checkModified(params.modelDir, lastModified)) {
```
jason-dai, line 80: shouldn't `lastModified` set to the time when the model was last loaded?

Litchilitchy: this is a bug fixed in new pr #2544

Litchilitchy: Nobody would use this feature for now thus I could merge it in next pr?

PR #2538, opened 2020-07-07
Add Koalas argument support for spark.read.csv/json

hkvision:
Referred to: https://koalas.readthedocs.io/en/latest/_modules/databricks/koalas/namespace.html#read_csv
https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.read_csv.html
https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.read_json.html read_json doesn't support arguments except `index_col`

Direct mapping arguments are as follows. There is no need to test them.
- sep -> sep
- quotechar -> quote
- escapechar -> escape
- comment -> comment

The following arguments only support a single value, which are meaningless actually. Just put it in our api to align with koalas.
- mangle_dupe_cols: only support True
- parse_dates: only support False

The following arguments are covered by unit tests:
- header
- names
- usecols
- dtype
- squeeze
- index_col

For read_json, `seq, quote, escape, comment, header` are not compatible. Only the remaining arguments are supported.


hkvision:
@jason-dai This PR is finished. Please take a look.
@jenniew Could you please also have a review for the implementation and argument support versus Koalas?

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -78,29 +78,135 @@ def loadFile(iterator):
...
             def f(iter):
                 import pandas as pd
                 data = list(iter)
-                yield pd.DataFrame(data, columns=columns)
+                pd_df = pd.DataFrame(data, columns=columns)
+                if squeeze and len(pd_df.columns) == 1:
+                    pd_df = pd_df.iloc[:, 0]
+                if index_col:
+                    pd_df = pd_df.set_index(index_col)
+                yield pd_df
```
jenniew, line 205: Seems we don't need yield as we don't generate iterables. Maybe we just return [pd_df]

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -78,29 +78,135 @@ def loadFile(iterator):
...
 
             return f
 
-        pd_rdd = df.rdd.mapPartitions(to_pandas(df.columns))
+        pd_rdd = df.rdd.mapPartitions(to_pandas(df.columns, squeeze, index_col))
 
     data_shards = SparkXShards(pd_rdd)
     return data_shards
```
jenniew, line 212: what will function do if use give pandas options but we do not support? provide warning?

hkvision: In this case, the arguments would be passed to spark.read.csv and that function don't recognize them and would raise an error.

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -78,29 +78,135 @@ def loadFile(iterator):
...
+                comment = kwargs["comment"]
+                if not isinstance(comment, str) or len(comment) != 1:
+                    raise ValueError("Only length-1 comment characters supported")
             df = spark.read.csv(file_paths, **kwargs)
+            if header is None:
+                df = df.selectExpr(
+                    *["`%s` as `%s`" % (field.name, i) for i, field in enumerate(df.schema)])
         else:
-            raise Exception("Unsupported file type")
+            df = spark.read.json(file_paths, **kwargs)
```
jason-dai, line 144: no need to handle pandas-compatible keyword arguments for json?

pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -78,29 +78,135 @@ def loadFile(iterator):
...
+        assert file_type == "json" or file_type == "csv", \
+            "Unsupported file type: %s. Only csv and json files are supported for now" % file_type
         from pyspark.sql import SQLContext
         sqlContext = SQLContext.getOrCreate(sc)
         spark = sqlContext.sparkSession
         # TODO: add S3 confidentials
-        if file_type == "json":
-            df = spark.read.json(file_paths, **kwargs)
-        elif file_type == "csv":
+
```
jason-dai, line 88: need to add license disclaimer, similar to https://github.com/apache/spark/blob/master/core/src/main/java/org/apache/spark/util/collection/TimSort.java and https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/examples/ray/parameter_server/model.py

pyzoo/zoo/orca/data/pandas/preprocessing.py
hkvision: For the arguments such as header; comment; escapechar, etc; spark.read.json doesn't support them, they are csv only. For the arguments that operate on the resulting dataframe, such as usecols, dtype, etc; they are shared by json and csv, even though koalas don't support them for json.
Do you mean report error for header, etc when reading json?

pyzoo/zoo/orca/data/pandas/preprocessing.py
hkvision: Okay!

pyzoo/zoo/orca/data/pandas/preprocessing.py
jason-dai: So we currently support arguments that operate on the resulting dataframe for json?

hkvision: Yeah, more specifically, we support names, usecols, index_col, dtype, squeeze.

pyzoo/zoo/orca/data/pandas/preprocessing.py
hkvision: Take a look? @jason-dai

jason-dai:
LGTM

hkvision:
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3761/
Several failures are unrelated.

PR #2540, opened 2020-07-07
TCMF add xshards support

cyita:


pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -20,6 +20,7 @@
 from zoo.automl.model.VanillaLSTM import VanillaLSTM as LSTMKerasModel
 from zoo.tfpark import KerasModel as TFParkKerasModel
 from zoo.automl.model import TCMF
+from zoo.orca.data.shard import SparkXShards, SharedValue
```
hkvision, line 23: from zoo.orca.data import ...

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -62,7 +63,8 @@ def __init__(self,
                  init_FX_epoch=100,
                  max_FX_epoch=300,
                  max_TCN_epoch=300,
-                 alt_iters=10):
+                 alt_iters=10,
+                 loaded_model=None):
```
shane-huang, line 67: do not add loaded_model to the constructor. Users should only use load to load models.

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,7 +210,30 @@ def predict(self,
...
+
+    def save(self, path):
+        if isinstance(self.internal, SparkXShards):
+            self.internal.save_pickle(path)
+        # TODO: single node
+
+    @classmethod
+    def load(cls, path, minPartitions=None):
+        loaded_model_shards = SparkXShards.load_pickle(path, minPartitions=minPartitions)
+        return TCMFForecaster(loaded_model=loaded_model_shards)
```
shane-huang, line 234: do not use the constructor to set the internal model. You can just use TCMFForcaster._internal=loaded_model

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,7 +210,30 @@ def predict(self,
...
+
+        if isinstance(self.internal, SparkXShards):
+            return self.internal.transform_shard(orca_predict, horizon)
+        else:
+            return self.internal.predict(x=x, horizon=horizon)
+
+    def save(self, path):
+        if isinstance(self.internal, SparkXShards):
+            self.internal.save_pickle(path)
+        # TODO: single node
```
shane-huang, line 229: we do not have single node saving method? @yushan111 I thought you have added one already.

shane-huang:
Add documentation about :param x: the input for fit. Let user know what kind of input he can send into the model.

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -157,10 +162,24 @@ def fit(self,
...
+            if incremental:
+                tcmf.fit_incremental(train_data)
+            else:
+                tcmf.fit_eval(train_data)
+            return [cid_arr, tcmf]
+
+        if isinstance(x, SparkXShards):
+            model_shared_value = SharedValue(self.internal)
+            self.internal = x.transform_shard(orca_train_model, model_shared_value, incremental)
         else:
```
shane-huang, line 178: also test whether x is numpy array. These are the only data formats we support. XShards for disributed and numpy for single node,  Otherwise  raise error.

shane-huang:
Maybe we can define two model wrappers inside of TCMFForecaster, one for single node, e.g. SingleNodeModelWrapper and the other for distributed, e.g. DistributedModelWrapper. They all have fit, predict, save and load. Just they accept different parameters in fit and load. 
In this case, in fit and load, if input is XShard, you can set internal to be DistributedModelWrapper, if numpy array you can set internal to be SingleNodeModelWrapper. Then in predict and save you don't have to do if/else, and just call xxWrapper.predict and save.

shane-huang:
When adding unittest, turn input from a dataframe to a numpy array and in post processing turn back to dataframe. That could serve as an example for users.

pyzoo/zoo/zouwu/model/forecast.py
cyita: Single node save & load added.

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -151,16 +153,35 @@ def fit(self,
...
+            else:
+                tcmf.fit_eval(train_data)
+            return [cid_arr, tcmf]
+
+        if isinstance(x, SparkXShards):
+            if x._get_class_name() == "numpy.ndarray":
+                model_shared_value = SharedValue(self.internal)
+                self.internal = x.transform_shard(orca_train_model, model_shared_value, incremental)
+            else:
+                raise ValueError("value of x should be a ndarray or a xShards of ndarray")
```
jason-dai, line 177: "x should be ..., but is a xShards of ..."

hkvision: a xShards -> an

pyzoo/test/zoo/zouwu/test_model_forecast.py
```diff
@@ -106,6 +106,49 @@ def test_forecast_tcmf(self):
...
+        from zoo.zouwu.model.forecast import TCMFForecaster
+        import zoo.orca.data.pandas
+        from zoo.orca.data import SharedValue
+        import os.path
+
+        resource_path = os.path.join(os.path.split(__file__)[0],
+                                     "../resources/zouwu/tcmf_random_data.csv")
+
+        def preprocessing(df):
+            idx = df.index.values
```
jason-dai, line 119: Does this work? All the partition on the same node will share the the same copy of the pytorch model. I think you need to clone the model for each partition.

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -151,16 +153,35 @@ def fit(self,
...
+        :param x: the input for fit. Only ndarray and SparkXShards of ndarray are supported
         :param covariates: the global covariates
         :param lr: learning rate
         :param incremental: if the fit is incremental
         :return:
         """
-        if incremental:
-            self.internal.fit_incremental(x)
+        def orca_train_model(np, shared_model, incremental):
+            tcmf = shared_model.value
```
jason-dai, line 163: you should keep two variable, one `internal` for the model, one `model_shards` for the xShards of the models, and only one of them should be available; otherwise there can be errors when

jason-dai, line 163: Does this work in your testing? The partitions on the same node will share the same copy of the model, which is not correct. You should clone the model for each partition.

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -151,16 +153,35 @@ def fit(self,
...
+
+        if isinstance(x, SparkXShards):
+            if x._get_class_name() == "numpy.ndarray":
+                model_shared_value = SharedValue(self.internal)
+                self.internal = x.transform_shard(orca_train_model, model_shared_value, incremental)
+            else:
+                raise ValueError("value of x should be a ndarray or a xShards of ndarray")
+        elif isinstance(x, np.ndarray):
+            if incremental:
+                self.internal.fit_incremental(x)
```
jason-dai, line 180: what if internal is an xShards here?

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,7 +212,33 @@ def predict(self,
...
+            cid_arr = data[0]
+            tcmf = data[1]
+            predict_results = tcmf.predict(x=None, horizon=horizon)
+            results = np.concatenate([np.expand_dims(cid_arr, axis=1), predict_results], axis=1)
+            return results
+
+        if isinstance(self.internal, SparkXShards):
+            return self.internal.transform_shard(orca_predict, horizon)
+        else:
+            return self.internal.predict(x=x, horizon=horizon)
```
jason-dai, line 225: what if internal is an xShards here?

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,7 +212,33 @@ def predict(self,
...
-        return self.internal.predict(x=x, horizon=horizon)
+        def orca_predict(data, horizon=24):
+            cid_arr = data[0]
+            tcmf = data[1]
+            predict_results = tcmf.predict(x=None, horizon=horizon)
+            results = np.concatenate([np.expand_dims(cid_arr, axis=1), predict_results], axis=1)
+            return results
+
+        if isinstance(self.internal, SparkXShards):
+            return self.internal.transform_shard(orca_predict, horizon)
```
jason-dai, line 223: what if internal is a PyTorch model here?

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -151,16 +153,35 @@ def fit(self,
...
+            train_data = np[:, 1:]
+            if incremental:
+                tcmf.fit_incremental(train_data)
+            else:
+                tcmf.fit_eval(train_data)
+            return [cid_arr, tcmf]
+
+        if isinstance(x, SparkXShards):
+            if x._get_class_name() == "numpy.ndarray":
+                model_shared_value = SharedValue(self.internal)
```
jason-dai, line 174: what if internal is an xShards here?

jason-dai:
I think there are two issues:

1) The TCMF model can be either a local one (when `internal` is a PyTorch model), or a distributed one (when `internal` is an xShards of models); and the user can invoke `fit`, `predict` and `evaluate` multiple times (with either ndarray or xShards as inputs). The current implementation can easily go wrong if the user pass different input data types in different invocations.

2) When the TCMF model is a distributed one, the assumption is that each partition contains a model trained for specific time series; then during prediction and evaluation, you need to map the input time series to appropriate partitions.

pyzoo/test/zoo/zouwu/test_model_forecast.py
```diff
@@ -106,6 +106,49 @@ def test_forecast_tcmf(self):
...
         model.evaluate(x=None, target_value=target_value, metric=['mse'])
 
+    def test_forecast_tcmf_xshards(self):
+        from zoo.zouwu.model.forecast import TCMFForecaster
+        import zoo.orca.data.pandas
+        from zoo.orca.data import SharedValue
+        import os.path
+
+        resource_path = os.path.join(os.path.split(__file__)[0],
+                                     "../resources/zouwu/tcmf_random_data.csv")
```
shane-huang, line 116: if this is a random data you can just generate them instead of loading them from a csv.

pyzoo/zoo/zouwu/model/forecast.py
shane-huang: The only parameter used is the horizon. We make this interace to accept x to make it compatible with abstract Forcaster interface but x is never actually used. We might need to check whether x is None or give a warning to users if it is not. 

But it is true that if we do incremental fitting/rolling prediction we will need to do id matching.  That will be in handled in the fit function.

shane-huang:
> I think there are two issues:
> 
> 1. The TCMF model can be either a local one (when `internal` is a PyTorch model), or a distributed one (when `internal` is an xShards of models); and the user can invoke `fit`, `predict` and `evaluate` multiple times (with either ndarray or xShards as inputs). The current implementation can easily go wrong if the user pass different input data types in different invocations.

Agree. We need to add a few sanitiy checks in these interfaces. Bascially we would want fit (full-train mode) to be called only once and fit_incremental to be called many times for adding new observations for rolling predictions. It should be properly tested. We can currently disable the incremental option (raise an error if enter that path). And if we call fit (full-train mode) more than once an error should be raised. 

I think it's better to use two wrappers as I mentioned, so that the conditional checking won't spread everywhere in code, and later we may change the implementation of the distributed version. 

> 2. When the TCMF model is a distributed one, the assumption is that each partition contains a model trained for specific time series; then during prediction and evaluation, you need to map the input time series to appropriate partitions.

We don't need to map the input time series for prediction but we do need it for evaluation. And the input of evaluation should be partitioned in the same way as input x shard by id. I'm not sure if that is even possible. Maybe we should just disable the distributed version of the evaluate method.

jason-dai:
> We don't need to map the input time series for prediction but we do need it for evaluation. And the input of evaluation should be partitioned in the same way as input x shard by id. I'm not sure if that is even possible. Maybe we should just disable the distributed version of the evaluate method. - if distrbuted version we just collect all prediction results and compare with evaluation input.

OK - I guess you don't actually need input in prediction. For evaluation, the true `y` is only needed after prediction, so as to compute the metrics with the predicted `y_hat`, yes? We can compute the metrics using `y` and `y_hat` by co-partitioning them together after the model prediction then.

shane-huang:
> eeded after predictio

Does Xshard support co partitioning?

shane-huang:
> > We don't need to map the input time series for prediction but we do need it for evaluation. And the input of evaluation should be partitioned in the same way as input x shard by id. I'm not sure if that is even possible. Maybe we should just disable the distributed version of the evaluate method. - if distrbuted version we just collect all prediction results and compare with evaluation input.
> 
> OK - I guess you don't actually need input in prediction. For evaluation, the true `y` is only needed after prediction, so as to compute the metrics with the predicted `y_hat`, yes? We can compute the metrics using `y` and `y_hat` by co-partitioning them together after the model prediction then.

Yes I think we can first obtain the prediction results (partitioned by id's) and then co-parition y and y_hat in case the two does not have the same id partitions.

shane-huang:
@cyita 

To summarize what we need to change in the code:

1.  create two wrapper classes to wrap single-node and distributed version for future extension and to avoid conditional checks spreading everywhere. 

1. in fit, 
     - diable incremental_fit for now (raise error and shan will enable it in rolling prediction and do further checks)
    -  only allow full traininng once
    - in fit add a flag id_as_firstcol
        -  if true, distributed hehave as usual, single-node will need to handle the 1st col as id, attach it in prediction. 
       -   if false, distributed raise error (in case of repartition), single-node will just predict
1. add a function to check whether model is distributed to replace `instanceof` checks
1. for evaluate, handle them case by case:
       - `y` ndarray, `yhat` xshard -> collect and do evaluation locally? or distributed the y and evalute in distributed fashion? seems former is faster. 
       - `y` xshard, `yhat` xshard -> co-partition and evaluate on each paritition and aggregate
       - `y` ndarray, `yhat` ndarray -> local evaluation 
      - `y` xshard, `yhat` ndarray -> raise error, probabaly an error in this case. 
           

-  if yes, same behaviro in ditributed

jason-dai:
> @cyita
> 
> To summarize what we need to change in the code:
> 
> 1. create two wrapper classes to wrap single-node and distributed version for future extension and to avoid conditional checks spreading everywhere.
> 2. in fit,
>    
>    * diable incremental_fit for now (raise error and shan will enable it in rolling prediction and do further checks)
>    * only allow full traininng once
>    * in fit add a flag id_as_firstcol
>      
>      * if true, distributed hehave as usual, single-node will need to handle the 1st col as id, attach it in prediction.
>      * if false, distributed raise error (in case of repartition), single-node will just predict
> 3. add a function to check whether model is distributed to replace `instanceof` checks
> 4. for evaluate, handle them case by case:
>    - `y` ndarray, `yhat` xshard -> collect and do evaluation locally? or distributed the y and evalute in distributed fashion? seems former is faster.
>    - `y` xshard, `yhat` xshard -> co-partition and evaluate on each paritition and aggregate
>    - `y` ndarray, `yhat` ndarray -> local evaluation
>    - `y` xshard, `yhat` ndarray -> raise error, probabaly an error in this case.

I think we can disable evaluation for the distribtued model for now, and add it in another PR.

jason-dai:
> 1. create two wrapper classes to wrap single-node and distributed version for future extension and to avoid conditional checks spreading everywhere.

Does the user need to use different interface to create a local vs. distributed model?

cyita:
> > 1. create two wrapper classes to wrap single-node and distributed version for future extension and to avoid conditional checks spreading everywhere.
> 
> Does the user need to use different interface to create a local vs. distributed model?

No. The fit method will check the input type and assign the corresponding wrapper to self.internal.

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -140,27 +142,35 @@ def __init__(self,
...
-        self.model = self._build()
-
-    def _build(self):
-        self.internal = TCMF()
-        return self.internal._build(**self.config)
 
     def fit(self,
-            x,
-            incremental=False):
+            x, id_as_first_col=True, incremental=False):
```
jason-dai, line 147: why do we need `id_as_first_col` anyway? when will the user pass false?

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -140,27 +142,35 @@ def __init__(self,
...
+                raise ValueError("value of x should be a ndarray or an xShards of ndarray")
+
+            try:
+                self.internal.fit(x, id_as_first_col, incremental)
+            except Exception as inst:
+                self.internal = None
+                raise inst
         else:
-            self.internal.fit_eval(x)
+            raise Exception("This model has been full trained, "
```
jason-dai, line 172: This model has already been fully trained

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,7 +201,35 @@ def predict(self,
         :param covariates: the global covariates
         :return:
         """
-        return self.internal.predict(x=x, horizon=horizon)
+        if self.internal is None:
+            raise Exception("You should run fit before call predict()")
```
jason-dai, line 205: before calling

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,7 +201,35 @@ def predict(self,
...
+        else:
+            return self.internal.predict(x, horizon)
+
+    def save(self, path):
+        if self.internal is None:
+            raise Exception("You should run fit before call save()")
```
jason-dai, line 211: before calling

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,7 +201,35 @@ def predict(self,
...
+        else:
+            self.internal.save(path)
+
+    def is_distributed(self):
+        if self.internal is None:
+            raise ValueError("You should run fit before call is_distributed()")
+        elif isinstance(self.internal, TCMFDistributedModelWrapper):
```
jason-dai, line 218: `return internal.is_distributed()`

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -183,3 +186,169 @@ def _get_optional_parameters(self):
...
+            else:
+                tcmf.fit_eval(train_data)
+            return [cid_arr, tcmf]
+
+        if not id_as_first_col:
+            raise Exception("id should be the first column of ndarray in xShards")
+
+        if isinstance(x, SparkXShards):
+            if x._get_class_name() == "numpy.ndarray":
+                config_shared_value = SharedValue(self.config)
```
jason-dai, line 237: config is small, and there is no need to use `SharedValue`

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -183,3 +186,169 @@ def _get_optional_parameters(self):
...
+    def load(self, model_path, minPartitions=None):
+        """
+        restore model from model file and config.
+        :param model_path: the model file
+        :return: the restored model
+        """
+        self.internal = SparkXShards.load_pickle(model_path, minPartitions=minPartitions)
+
+
+class TCMFSingleNodeModelWrapper(ModelWrapper):
```
jason-dai, line 288: TCMFLocalModelWrapper

pyzoo/zoo/zouwu/model/forecast.py
cyita: Since single-node training won't change the order of input, users can choose to attach id as the first column or not. If the user passes False in single node training, the index will be assigned as the first column of predict result to ensure the same behavior as distributed training.

shane-huang:
LGTM. Once it passed pr validation it can be merged I think. What do you think ? @jason-dai

cyita:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/3794/

pyzoo/zoo/zouwu/model/forecast.py
jason-dai: I think it is better to pass a dataframe (or xShards of dataframe) in this case; the dataframe can contain a column of 'id', and another column of 'y' or 'target'. Then `fit` can check the schema of the input, and `predict` can also return dataframe (or xShards of dataframe).

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,20 +194,26 @@ def evaluate(self,
 
     def predict(self,
                 x=None,
+                output_col_name=None,
```
jason-dai, line 197: change to `prediction_col` with default value 'prediction'

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -144,11 +144,13 @@ def __init__(self,
         }
 
     def fit(self,
-            x, id_as_first_col=True, incremental=False):
+            x, id_col_name, target_col_name, incremental=False):
```
jason-dai, line 147: make it consistent with https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/tf/estimator.py#L205; maybe change to `data, id_col, target_col`, and need to provide reasonable default value for `id_col` and `target_col`

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -191,20 +194,26 @@ def evaluate(self,
...
+        :param output_col_name: the column name list of predict result, type: list
         :param horizon: horizon length to look forward.
         :param covariates: the global covariates
         :return:
         """
         if self.internal is None:
             raise Exception("You should run fit before calling predict()")
         else:
-            return self.internal.predict(x, horizon)
+            if isinstance(output_col_name, list) and len(output_col_name) == horizon:
```
jason-dai, line 212: if `output_col_name` is None, we should use some reasonable default value for `output_col_name`

jason-dai:
I think we should first make sure we have consistent APIs for the models:
1) For distributed model 
     - The input to `fit` should be an xShards of Dataframes which contains both 'id' and 'target'
     - There is no input to `predict`, which should return an xShards of Dataframe which contains 'id' and 'prediction'
     - The input to `evaluate` should be an xShards of Dataframes which contains both 'id' and 'target'; we can disable it for now and support it in a follow-up PR
2) For local model
     - The input to `fit` can be either an NDArray (without 'id'), or a Dataframe which contains 'target' (and optionally 'id'); do we need to support both @shane-huang @yushan111 ?
     - There should be no input to `predict`; depending one whether Dataframe or NDArray is used as input in `fit`, it should also return Dataframe or NDArray
     - Similarly, the input to `evaluate` can be a Dataframe or NDArray

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -183,3 +187,189 @@ def _get_optional_parameters(self):
...
+    def predict(self, x, output_col_name, horizon=24):
+        """
+        Prediction.
+        :param x: input
+        :return: result
+        """
+        pred = self.internal.predict(x=x, horizon=horizon)
+        results = np.concatenate([np.expand_dims(self.id_arr, axis=1), pred], axis=1)
+        final_df = pd.DataFrame(results, columns=["id"] + output_col_name)
+        final_df.id = final_df.id.astype("int")
```
yushan111, line 349: Do we support input non-numerical id, like string?

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -183,3 +187,189 @@ def _get_optional_parameters(self):
...
+        final_df = final_df.set_index("id")
```
yushan111, line 350: Output prediction df may need to be consist with the input df of fit. 
1. The name of id column should be the same.
2. If id is a column of input df, it shouldn't be an index of output prediction df.

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -183,3 +187,189 @@ def _get_optional_parameters(self):
...
+class TCMFDistributedModelWrapper(ModelWrapper):
+
+    def __init__(self, config):
+        self.internal = None
+        self.config = config
+
+    def fit(self, x, id_col_name, target_col_name, incremental=False):
+        def orca_train_model(df, config, incremental):
+            tcmf = TCMF()
+            tcmf._build(**config)
```
yushan111, line 227: If input dataframe, some of the configs should be able to be inferred, such as "start_date", "freq". Maybe we should also change the interface, like moving "start_date", "freq", "dti" from constructor to fit.

shane-huang:
> I think we should first make sure we have consistent APIs for the models:
> 
> 1. For distributed model
>    
>    * The input to `fit` should be an xShards of Dataframes which contains both 'id' and 'target'
>    * There is no input to `predict`, which should return an xShards of Dataframe which contains 'id' and 'prediction'
>    * The input to `evaluate` should be an xShards of Dataframes which contains both 'id' and 'target'; we can disable it for now and support it in a follow-up PR
> 2. For local model
>    
>    * The input to `fit` can be either an NDArray (without 'id'), or a Dataframe which contains 'target' (and optionally 'id'); do we need to support both @shane-huang @yushan111 ?
>    * There should be no input to `predict`; depending one whether Dataframe or NDArray is used as input in `fit`, it should also return Dataframe or NDArray
>    * Similarly, the input to `evaluate` can be a Dataframe or NDArray

jason-dai:
> I think we should first make sure we have consistent APIs for the models:
> 
> 1. For distributed model
>    
>    * The input to `fit` should be an xShards of Dataframes which contains both 'id' and 'target'
>    * There is no input to `predict`, which should return an xShards of Dataframe which contains 'id' and 'prediction'
>    * The input to `evaluate` should be an xShards of Dataframes which contains both 'id' and 'target'; we can disable it for now and support it in a follow-up PR
> 2. For local model
>    
>    * The input to `fit` can be either an NDArray (without 'id'), or a Dataframe which contains 'target' (and optionally 'id'); do we need to support both @shane-huang @yushan111 ?
>    * There should be no input to `predict`; depending one whether Dataframe or NDArray is used as input in `fit`, it should also return Dataframe or NDArray
>    * Similarly, the input to `evaluate` can be a Dataframe or NDArray

@shane-huang @cyita @yushan111 To be clear, I think there are two options for the input to distributed `fit`:
1. It can be an xShards of Dataframe where each row is a univariate time series; each row has two column: 'id' is the ID value, and 'y' is a one-dimensional vector for the univariate time series

2. Alternatively it can be an xShards of dictionary with two keys: 'id' is a one-dimensional vector for all the IDs, and 'y' is a NxT ndarray for all the time series data.

The local `fit` also takes either a Dataframe or dictionary, with 'id' being optional.

pyzoo/test/zoo/zouwu/test_model_forecast.py
```diff
@@ -94,19 +94,163 @@ def test_forecast_mtnet(self):
...
+        horizon = np.random.randint(1, 50)
+        # construct data
+        id = np.arange(200)
+        data = np.random.rand(300, 480)
+        input = dict({'cid': id, 'data': data})
+        with self.assertRaises(Exception) as context:
+            model.fit(input, id_col="cid", target_col="data")
+        self.assertTrue("the length of the id_col should be equal to the number of"
+                        in str(context.exception))
+        model.fit(input, id_col=None, target_col="data")
```
yushan111, line 157: what if user doesn't pass id_col in local

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -183,3 +187,219 @@ def _get_optional_parameters(self):
...
+        :param x: input
+        :return: result
+        """
+        def orca_predict(data, id_col):
+            id_arr = data[0]
+            tcmf = data[1]
+            predict_results = tcmf.predict(x=None, horizon=horizon)
+            result = dict()
+            result[id_col] = id_arr
+            result["prediction"] = predict_results
```
yushan111, line 273: should we allow user to pass a name for prediction or just use the fixed name?

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -371,5 +379,27 @@ def load(self, model_path):
...
         self.internal.restore(model_path + "/model")
+
+
+def split_id_and_train_data(d, id_col, target_col):
+    if target_col in d:
+        train_data = d[target_col]
+        if not isinstance(train_data, np.ndarray):
+            raise ValueError("the value of target_col should be an ndarray")
+    else:
+        raise ValueError("target_col doesn't exist in x")
```
jason-dai, line 395: show the value of target_col  in the error message

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -371,5 +379,27 @@ def load(self, model_path):
...
+    id_arr = None
+    if id_col is not None:
+        if id_col in d:
+            id_arr = d[id_col]
+            if len(id_arr) != train_data.shape[0]:
+                raise ValueError("the length of the id_col should be equal to the number of "
+                                 "rows in the target_col")
+        else:
+            raise ValueError("id_col doesn't exist in x")
```
jason-dai, line 404: show the value of id_col in the error message

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -371,5 +379,27 @@ def load(self, model_path):
...
+        if not isinstance(train_data, np.ndarray):
+            raise ValueError("the value of target_col should be an ndarray")
+    else:
+        raise ValueError("target_col doesn't exist in x")
+    id_arr = None
+    if id_col is not None:
+        if id_col in d:
+            id_arr = d[id_col]
+            if len(id_arr) != train_data.shape[0]:
+                raise ValueError("the length of the id_col should be equal to the number of "
```
jason-dai, line 401: show the value of id_col and target_col in the error message

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -144,13 +144,13 @@ def __init__(self,
         }
 
     def fit(self,
-            x, id_col_name, target_col_name, incremental=False):
+            x, id_col='id', target_col='y', incremental=False):
```
jason-dai, line 147: instead of `id_col` and `target_col`, just `id` and `y`. Alternatively, we can always use `id` and `y` instead of making these two additional parameters

pyzoo/zoo/automl/model/tcmf_model.py
jason-dai: we can make `id_col` optional for local?

jason-dai:
I think we can simply the logic by (1) always ask users to pass `id` and `y` in the dictionary to `fit` (2) for local `fit`, the user may omit `id` (3) always return `prediction` (and optionally `id`) in the dictionary by `predict`

pyzoo/zoo/automl/model/tcmf_model.py
cyita: non-numerical id has been supported in latest commit

pyzoo/test/zoo/zouwu/test_model_forecast.py
cyita: The default value of id has been removed.

pyzoo/zoo/zouwu/model/forecast.py
```diff
@@ -144,13 +144,13 @@ def __init__(self,
         }
 
     def fit(self,
-            x, id_col='id', target_col='y', incremental=False):
+            x, id, y, incremental=False):
```
jason-dai, line 147: no need for `id` or `y` as parameters; just always use 'id' and 'y'

pyzoo/zoo/automl/model/tcmf_model.py
```diff
@@ -381,25 +382,25 @@ def load(self, model_path):
...
+        if id in d:
+            id_arr = d[id]
             if len(id_arr) != train_data.shape[0]:
-                raise ValueError("the length of the id_col should be equal to the number of "
-                                 "rows in the target_col")
+                raise ValueError("the length of the id array should be equal to the number of "
+                                 "rows in the y")
         else:
-            raise ValueError("id_col doesn't exist in x")
+            raise ValueError("id " + id + " doesn't exist in x")
```
jason-dai, line 405: just return None; report error in distributed `fit`, but it's OK for local `fit`

jason-dai:
LGTM

cyita:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/3827/

PR #2541, opened 2020-07-07
Fix AUC changing output and target

hkvision:
Same as https://github.com/intel-analytics/BigDL/pull/3011
AUC would change the output and target, which would influence other validation metrics.


hkvision:
jenkins passed: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3734/

dding3:
LGTM

hkvision:
Thanks so much for your review! @dding3

Issue #2528, opened 2020-07-02
Orca data tests are not added in Jenkins


hkvision:
https://github.com/intel-analytics/analytics-zoo/tree/master/pyzoo/zoo/examples/orca/data

ray_pandas.py is not needed as it is going to be removed.
Need to add spark_pandas.py to Jenkins integration (both pip and non-pip).

PR #2542, opened 2020-07-08
Add orca tf estimator tests and orca data example to Jenkins

jenniew:
Add orca tf estimator tests and orca data example to Jenkins.

Related issue: https://github.com/intel-analytics/analytics-zoo/issues/2528


jenniew:
Jenkins pass： https://github.com/analytics-zoo/orca/issues/57

PR #2543, opened 2020-07-08
fix tcmf memory leak

yushan111:


shane-huang:
seems changes is only using torch.no_grad to avoid constructing computational graphs and change prediction batch size. LGTM for now.

yushan111:
jenkins passed: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3737

PR #2566, opened 2020-07-14
init_spark_standalone

hkvision:
Initial implementation of running Analytics Zoo on Spark standalone cluster.
Users don't need to handle anything with spark, just call init_spark_standalone() and it would do the environment packaging and cluster setup automatically.

Current assumption:
- JDK8 is installed on every node beforehand.
- Better that the master can ssh all workers without password. Otherwise users would be required to input password from time to time.


hkvision:
@jason-dai This initial version can work. Take a look and give some comments :)

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3839/

PR #2570, opened 2020-07-15
update bigdl version to 0.11.0

Le-Zheng:


zoo/pom.xml
```diff
@@ -27,7 +27,7 @@
         <scala.macros.version>2.1.0</scala.macros.version>
         <scalatest.version>2.2.4</scalatest.version>
         <spark.version>2.4.3</spark.version>
-        <bigdl.version>0.10.0</bigdl.version>
+        <bigdl.version>0.11.0</bigdl.version>
```
qiyuangong, line 30: Need to update version in apps, docker and doc

hkvision: Also need to change python dependency and related doc.

Le-Zheng: Done

zoo/pom.xml
```diff
@@ -27,7 +27,7 @@
         <scala.macros.version>2.1.0</scala.macros.version>
         <scalatest.version>2.2.4</scalatest.version>
         <spark.version>2.4.3</spark.version>
-        <bigdl.version>0.10.0</bigdl.version>
```
qiyuangong, line 30: 0.11 or 0.12?

qiyuangong, line 30: Does it mean we skipped 0.11?

PR #2572, opened 2020-07-15
Simplify horovod start process

yangw1234:
Simplify horovod start process. 

It also works around the cloudpickle conflict problem.

Instead of finding a common network interface among workers, just use the interface corresponding to the ip address returned by ray.

Tested on:
local
almaren cluster
yarn integration test cluster


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3784/ jenkins passed except the mxnet download error

PR #2573, opened 2020-07-16
Initial support for tf ray estimator

yangw1234:
Usage:

```python
trainer = TFRayEstimator(
    model_creator=simple_model,
    compile_args=compile_args,
    verbose=True,
    config=create_config(32))
trainer.fit(train_data_creator, epochs=5)
trainer.evaluate(val_data_creator, steps=10)
```
Here the `model_creator` requires returning a model without `compile`, and users need to pass compile arguments through the constructor. The reason is if we are going to support horovod, we need to wrap `optimizer`  with `hvd.DistributedOptimizer` before compiling the model.

Todos:

- [x] support setting inter/inter_op_parallelism. 
- [x] verify callback works

will add ut after jenkins has enabled horovod


yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3994/

PR #2590, opened 2020-07-20
Predict result as xshard

jenniew:
Support predict result as XShards if input data is XShards.

Related issue: https://github.com/analytics-zoo/orca/issues/58


pyzoo/zoo/orca/learn/tf/utils.py
```diff
@@ -124,6 +130,21 @@ def combine(pair):
...
+            predictions = np.array(predictions).T.tolist()
+            result = [np.array(predict) for predict in predictions]
+            return [result]
+        # np array
+        else:
+            return [np.array(predictions)]
+
+    predict_rdd = prediction_rdd.mapPartitions(transform_predict)
+    return data_shard.rdd.zip(predict_rdd)\
+        .map(lambda x_predict: {'x': x_predict[0]['x'], 'prediction': x_predict[1]})
```
jason-dai, line 147: I don't think we need to include x in the output, as x is just a vector with no id.

What's the format of the input xShards for predict?

jenniew: input format is {'x': numpy array or list of numpy arrays}

jenniew: remove 'x' in prediction result.

jason-dai:
LGTM

jenniew:
Jenkins pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/3981/

PR #2591, opened 2020-07-21
Orca tf estimator checkpoint and resume training

jenniew:
Support Orca tf estimator with checkpoint and resume training.

Related issue: https://github.com/analytics-zoo/orca/issues/56


pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -107,6 +119,7 @@ def __init__(self, *, inputs, outputs, labels, loss,
         else:
```
jason-dai, line 119: should be `"optimizer is of type {}, ".format(type(optimizer))` at line 89

jenniew: updated

jenniew:
Jenkins pass: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/3896/

PR #2593, opened 2020-07-21
Add OrcaContext and make spark default read file backend

hkvision:
- Move properties to OrcaContext.
@jason-dai Since log_output is used in init_nncontext, it can't be moved to OrcaContext?
- Make spark the default backend for reading csv/json files.
- When error occurs, add a message to hint users to switch backend.

### DIFF
--- a/pyzoo/zoo/common/nncontext.py
+++ b/pyzoo/zoo/common/nncontext.py
@@ -117,30 +117,17 @@ def init_spark_on_yarn(hadoop_conf,
     return sc
 
 
-class ZooContextMeta(type):
-
-    _log_output = False
+class ZooContext:
 
     @property
-    def log_output(cls):
-        """
-        Whether to redirect Spark driver JVM's stdout and stderr to the current
-        python process. This is useful when running Analytics Zoo in jupyter notebook.
-        Default to False. Needs to be set before initializing SparkContext.
-        """
-        return cls._log_output
+    def log_output(self):
+        from zoo.orca import OrcaContext
+        return OrcaContext.log_output
 
-    @log_output.setter
-    def log_output(cls, value):
-        if SparkContext._active_spark_context is not None:
-            raise AttributeError("log_output cannot be set after SparkContext is created."
-                                 " Please set it before init_nncontext, init_spark_on_local"
-                                 "or init_spark_on_yarn")
-        cls._log_output = value
-
-
-class ZooContext(metaclass=ZooContextMeta):
-    pass
+    @staticmethod
+    def log_output(value):
+        from zoo.orca import OrcaContext
+        OrcaContext.log_output = value
 
 
 # The following function copied from
--- a/pyzoo/zoo/common/nncontext.py
+++ b/pyzoo/zoo/common/nncontext.py
@@ -120,8 +120,6 @@ def init_spark_on_yarn(hadoop_conf,
 class ZooContextMeta(type):
 
     _log_output = False
-    __orca_eager_mode = True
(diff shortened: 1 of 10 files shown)


pyzoo/zoo/orca/common.py
```diff
@@ -0,0 +1,54 @@
...
+#
+
+
+class OrcaContextMeta(type):
+
+    _orca_eager_mode = True
+    _orca_pandas_read_backend = "spark"
+
+    @property
+    def orca_eager_mode(cls):
```
jason-dai, line 24: change to `_eager_mode`, as this is supposed to be an internal property

pyzoo/zoo/orca/common.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        return cls._orca_eager_mode
+
+    @orca_eager_mode.setter
+    def orca_eager_mode(cls, value):
+        assert isinstance(value, bool), "orca_eager_mode should either be True or False"
+        cls._orca_eager_mode = value
+
+    @property
+    def orca_pandas_read_backend(cls):
```
jason-dai, line 37: change to `pandas_read_backend`

pyzoo/zoo/orca/common.py
hkvision: OK. So eager_mode is internal (generally not to be called by users) and pandas_read_backend is external?

pyzoo/zoo/common/nncontext.py
```diff
@@ -115,8 +115,6 @@ def init_spark_on_yarn(hadoop_conf,
 class ZooContextMeta(type):
 
     _log_output = False
```
jason-dai, line 117: we should either move `log_output` to `OrcaContext`, or have `OrcaContext.log_output` to call `ZooContext.log_output`

pyzoo/zoo/orca/common.py
jason-dai: yes

pyzoo/zoo/common/nncontext.py
hkvision: All review comments addressed, take another look? @jason-dai

pyzoo/zoo/common/nncontext.py
```diff
@@ -112,30 +112,17 @@ def init_spark_on_yarn(hadoop_conf,
...
-        if SparkContext._active_spark_context is not None:
-            raise AttributeError("log_output cannot be set after SparkContext is created."
-                                 " Please set it before init_nncontext, init_spark_on_local"
-                                 "or init_spark_on_yarn")
-        cls._log_output = value
-
-
-class ZooContext(metaclass=ZooContextMeta):
-    pass
+    def log_output(self):
```
jason-dai, line 118: if you have moved `log_output` to `OrcaContext`, then we can just eliminate `ZooContext`

hkvision: Then is it strange to call OrcaContext outside Orca?

jason-dai: I think there are two options:
1) Move everything to OrcaContext and remove ZooContext
2) Keep the property in ZooContext, and have OrcaContext call ZooContext

hkvision: Updated. Take another look @jason-dai ?

jason-dai:
LGTM

hkvision:
jenkins: http://10.239.47.210:18888/job/ZOO-PR-Validation/3838/
One failure is due to node memory, not related to this PR.

PR #2606, opened 2020-07-22
support ray tune automl xgbregressor

dding3:


pyzoo/zoo/automl/model/XGBoostRegressor.py
```diff
@@ -0,0 +1,163 @@
...
+            self.metric = 'rmse'
+
+    def _build(self, **config):
+        """
+        build the models and initialize.
+        :param config: hyper parameters for building the model
+        :return:
+        """
+        self.set_params(**config)
+        self.model = XGBRegressor(n_estimators=self.num_rounds, max_depth=self.max_depth,
```
jason-dai, line 81: why n_estimators instead of num_rounds? we should make the parameters consistent with XGBoost

pyzoo/zoo/automl/search/RayTuneSearchEngine.py
```diff
@@ -233,8 +235,12 @@ def train_func(config):
...
             trial_ft = deepcopy(global_ft)
             # trial_model = deepcopy(global_model)
-            trial_model = TimeSequenceModel(check_optional_config=False,
-                                            future_seq_len=future_seq_len)
+            # print("config is ", config)
+            if 'model' in config.keys() and config['model'] == 'XGBRegressor':
+                trial_model = XGBoostRegressor()
+            else:
+                trial_model = TimeSequenceModel(check_optional_config=False,
+                                                future_seq_len=future_seq_len)
```
jason-dai, line 243: instead of doing this, pass a `model_fn()` that creates the trial model to RayTuneSearchEngine (either constructor function or `compile` function) @shane-huang @yushan111

jason-dai:
1)  Create something like `orca.automl.AutoXGBoost.regressor` 

2) Need an example for AutoXGBoost

pyzoo/zoo/automl/search/RayTuneSearchEngine.py
shane-huang: I agree that RayTuneSearchEngine should not contain any specific code to create a particular model type instance.  In the initial design, feature transformer and model are created outside of RayTuneSearchEngine and passed into RayTuneSearchEngine via RayTuneSearchEngine.compile(). Before trial run, we use ray.put to put transformer and model (only one instance of each) in plasma, and each ray task retrieves the transformer and model instances from plasma and continue using hyper parameter configuration to construct internal pipeline constructution (including the model building) and train execution. 

The reason why it ends up like this is that we found there're some serialization problem when putting the model to plasma. So we needs to create the function into it. I think we need to fix this. I'll submit an issue. We may not need to fix it in this pr but we may look at it later.

shane-huang: In this PR we can just pass in a function 
in `_hp_search` in xgboostpredictor:
```python
future_seq_len=...
def model_create_func():
      model = TimeSequenceModel(check_optional_config=False,
                                            future_seq_len=future_seq_len)
      return model
```
In prepare_train_func, replease the model creation part with below code, 
```python
def _prepare_train_func(model_create_func,
                          ...)
            ...
           trial_model = model_create_func()
```

shane-huang: also remember to change timesequencepredictor as well

jason-dai: > I agree that RayTuneSearchEngine should not contain any specific code to create a particular model type instance. In the initial design, feature transformer and model are created outside of RayTuneSearchEngine and passed into RayTuneSearchEngine via RayTuneSearchEngine.compile(). Before trial run, we use ray.put to put transformer and model (only one instance of each) in plasma, and each ray task retrieves the transformer and model instances from plasma and continue using hyper parameter configuration to construct internal pipeline (including building the actual model) and execute the flow (training or transformation).
> 
> The reason why it ends up like this is that we found there're some serialization problem when putting the model to plasma. So we create the model inside trial task instead of creating it ouside.

Why put it into plasma? if the object is small, just directly use it in the `train_func` and then it will be automatically pickled and sent to remote

pyzoo/zoo/automl/pipeline/time_sequence.py
```diff
@@ -219,3 +219,17 @@ def load_ts_pipeline(file):
                                        config=all_config)
     print("Restore pipeline from", file)
     return ts_pipeline
+
+
+def load_xgboost_pipeline(file):
```
jason-dai, line 224: where is this used?

pyzoo/zoo/automl/feature/echo_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+from sklearn.preprocessing import MinMaxScaler, StandardScaler
+import pandas as pd
+import numpy as np
+import featuretools as ft
+from featuretools.primitives import make_agg_primitive, make_trans_primitive
+from featuretools.variable_types import Text, Numeric, DatetimeTimeIndex
+import json
+
+
+class EchoTransformer(BaseFeatureTransformer):
```
jason-dai, line 30: IdentityTransformer

pyzoo/zoo/automl/search/RayTuneSearchEngine.py
```diff
@@ -19,9 +19,11 @@
 from copy import copy, deepcopy
 
 from zoo.automl.model import TimeSequenceModel
+from zoo.automl.model import XGBoostRegressor
```
jason-dai, line 22: no longer needed for these two?

pyzoo/zoo/examples/orca/automl/AutoXGBoost/regressor.py
```diff
@@ -0,0 +1,74 @@
...
+    ray_ctx.init()
+
+    import pandas as pd
+    df = pd.read_csv(opt.path)
+    feature_cols = ["FIPS", "Lower 95% Confidence Interval", "Upper 95% Confidence Interval",
+                    "Average Annual Count", "Recent 5-Year Trend"]
+    target_col = "Age-Adjusted Incidence Rate"
+    train_df, val_df = train_test_split(df, test_size=0.2, random_state=2)
+
+    tsp = Regressor(feature_cols=feature_cols, target_col=target_col)
```
jason-dai, line 60: estimator = AutoXBGBoost.Regressor(...)

pyzoo/zoo/automl/pipeline/time_sequence.py
dding3: currently it's not used. but it can be used to restore the saved pipeline and use the pipeline to do the inference etc.

pyzoo/zoo/examples/orca/automl/AutoXGBoost/regressor.py
dding3: > `estimator = AutoXBGBoost.regressor(...)`, where AutoXBGBoost is a class and regressor is a constructor function

so we would like a single node AutoXGBoost training example here other than ray tune autoXGBoost example?

jason-dai: `estimator = AutoXBGBoost.regressor(...)` should return `XgbRegressorPredictor` or a wrapper of `XgbRegressorPredictor`

pyzoo/zoo/orca/automl/AutoXGBoost.py
```diff
@@ -18,15 +18,16 @@
...
 
 
-class Regressor(object):
-    def __init__(self,
-                 feature_cols,
-                 target_col,
-                 name="automl",
-                 logs_dir="~/zoo_automl_logs"):
+class AutoXGBoost(object):
+    def regressor(self,
```
jason-dai, line 22: this should be a static function

pyzoo/zoo/orca/automl/AutoXGBoost.py
```diff
@@ -18,15 +18,16 @@
...
+                  feature_cols,
+                  target_col,
+                  name="automl",
+                  logs_dir="~/zoo_automl_logs"):
         # models
         self.tsp = XgbRegressorPredictor(feature_cols, target_col,
                                          name, logs_dir)
+        return self.tsp
 
     def fit(self,
```
jason-dai, line 32: where is this used?

jason-dai:
How does the user control the threads? Otherwise LGTM.

dding3:
> How does the user control the threads? Otherwise LGTM.

user can pass a config to regressor to set the n_jobs to control the thread.

jason-dai:
> How does the user control the threads? Otherwise LGTM.

The default value is -1, which use all the cores; will it cause conflicts when running multiple trials in parallel in the same node? Maybe we should set it to a more reasonable value based on core_num.

pyzoo/zoo/automl/model/XGBoostRegressor.py
```diff
@@ -0,0 +1,166 @@
...
+        :param future_seq_len:
+        """
+        # models
+        if not config:
+            config = {}
+        self.n_estimators = config.get('n_estimators', 1000)
+        self.max_depth = config.get('max_depth', 5)
+        self.tree_method = config.get('tree_method', 'hist')
+        _, core_num = get_node_and_core_number()
+        self.n_jobs = config.get('n_jobs', core_num)
```
jason-dai, line 41: `core_num` will be the cpu resource allocated to each raylet @hkvision ? And how to we determine how many trials will run in parallel?

jason-dai, line 41: `core_num` will be the cpu resource allocated to each raylet @hkvision ? And how to we determine how many trials will run in parallel @yushan111 ?

hkvision: Each raylet would be given cores equal to the executor cores. This core_num is the available cores of a node, not necessarily equal to executor cores, right?

jason-dai: > Each raylet would be given cores equal to the executor cores. This core_num is the available cores of a node, not necessarily equal to executor cores if there are multiple executors on one node, right?

how do you determine the number of executor cores?

hkvision: From spark.executor.cores.

jason-dai: > From spark.executor.cores.

that should the same as core_num, no?

yushan111: The number of trials running in parallel will be: num_executors * executor_cores // cpus_per_trial. We have `resources_per_trial` where user can define how many resources (number of cpus, gpus) to allocate per trial.

dding3: @jason-dai what about if we set the default njobs=1 in xgbregressor and if xgbregressor is created from ray tune search engine, we set the value to the cpus_per_trial.

jason-dai: > @jason-dai what about if we set the default njobs=1 in xgbregressor and if xgbregressor is created from ray tune search engine, we set the value to the cpus_per_trial.

May we should
1) Set default njobs to -1
2) Set njobs to cpus_per_trial in fit
3) Use self.njobs value in predict and evaluate

dding3:
jekins pass. http://10.239.47.210:18888/job/ZOO-PR-Validation/4006/

jason-dai:
what `n_jobs` value will be used in predict and evaluate?

dding3:
> what `n_jobs` value will be used in predict and evaluate?

will use configed n_jobs, fixed in https://github.com/intel-analytics/analytics-zoo/pull/2739

PR #2613, opened 2020-07-24
Py kerasNet top 1 accuracy name switch

Litchilitchy:


pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -118,6 +118,8 @@ def get_validation_summary(self, tag=None):
         # Arguments
         tag: The string variable represents the scalar wanted
         """
+        if tag == 'Accuracy':
+            tag = 'Top1Accuracy'
         validation_set = set(('AUC', 'Accuracy', 'BinaryAccuracy', 'CategoricalAccuracy',
```
qiuxin2012, line 123: This set is also incorrect. Maybe user use a metric written by himself.

PR #2622, opened 2020-07-26
Update PythonFeatureSet.scala

GavinGu07:
Assign a fixed PythonFeatureSet's Serial Version UID to fix the incompatible UID exception


qiuxin2012:
jenkins passed

PR #2626, opened 2020-07-27
Cluster Serving add secure params

Litchilitchy:


zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -77,6 +77,12 @@ class ClusterServingHelper(_configPath: String = "config.yaml") {
   var weightPath: String = null
   var defPath: String = null
   var modelDir: String = null
+  /**
+   * secure related
+   */
+  var redisSecureEnabled: Boolean = true
```
glorysdj, line 83: change to 
redisSecureEnabled: Boolean = true,
    redissTrustStorePath: String = getClass.getClassLoader.getResource("keys/keystore.jks").getPath,
    redissTrustStorePassword: String = "1234qwer" 
for default values?

PR #2627, opened 2020-07-27
Fix quick-start.py

cwpeng-cn:
Fix quick-start.py


Litchilitchy:
close this PR and move the contents to #2629 if not there

PR #2629, opened 2020-07-27
Add cluster serving zip

cwpeng-cn:
For convenience, we have added a zip package for cluster serving, which contains the required sh script, python script, image data and serving.jar.


Litchilitchy:
1. rename to `zoo.jar` 

2. add `xxx-python-api.zip` to this zip

Litchilitchy:
create another sh to rename and export env var, could name it `cluster-serving-prepare.sh`

and demo script could call that script

demo is used to show usage, it is better to use another script to get ready for the environment needed.

scripts/cluster-serving/cluster-serving-demo.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+./cluster-serving-prepare.sh
+./cluster-serving-start -c ./config.yaml & sleep 20
+python3 ./quick_start.py --image_path ../test_image
+
+sleep 20
+./cluster-serving-stop
+
+echo "Demo end, please check at flink Dashboard"
```
qiyuangong, line 10: Add empty line at the end.

cwpeng-cn: ok

scripts/cluster-serving/cluster-serving-demo.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+./cluster-serving-prepare.sh
```
Litchilitchy, line 3: I think it should be `source ./cluster-serving-prepare.sh` here otherwise export var can not be valid at new shell?

cwpeng-cn: Thanks, the bug is fixed

PR #2630, opened 2020-07-28
support create featureset from pytorch dataloader creator functions.

qiuxin2012:
Support create FeatureSet from pytorch dataloader creator functions.

### DIFF
--- a/pyzoo/zoo/feature/common.py
+++ b/pyzoo/zoo/feature/common.py
@@ -380,7 +380,19 @@ class FeatureSet(DataSet):
             warnings.warn(warning_msg)
 
         bys = CloudPickleSerializer.dumps(CloudPickleSerializer, dataloader)
-        jvalue = callZooFunc(bigdl_type, "createFeatureSetFromPyTorch", bys)
+        jvalue = callZooFunc(bigdl_type, "createFeatureSetFromPyTorch", bys, False)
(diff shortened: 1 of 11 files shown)


pyzoo/zoo/feature/common.py
```diff
@@ -380,7 +380,19 @@ def pytorch_dataloader(cls, dataloader, bigdl_type="float"):
             warnings.warn(warning_msg)
 
         bys = CloudPickleSerializer.dumps(CloudPickleSerializer, dataloader)
-        jvalue = callZooFunc(bigdl_type, "createFeatureSetFromPyTorch", bys)
+        jvalue = callZooFunc(bigdl_type, "createFeatureSetFromPyTorch", bys, False)
+        return cls(jvalue=jvalue)
+
+    @classmethod
+    def pytorch_dataloader_creator(cls, creator, bigdl_type="float"):
```
jason-dai, line 387: combine `pytorch_dataloader_creator` with `pytorch_dataloader` - you may check the type of `data` argument

pyzoo/zoo/pipeline/api/torch/utils.py
```diff
@@ -17,6 +17,14 @@
 import math
 
 
+def training_param(model):
```
jason-dai, line 20: trainable_param

PR #2638, opened 2020-07-30
Change transform_shard using mapPartitions

jenniew:
Change transform_shard using mapPartitions instead of map.
Related issue: https://github.com/analytics-zoo/orca/issues/61


pyzoo/zoo/orca/data/shard.py
```diff
@@ -144,7 +144,12 @@ def __init__(self, rdd, transient=False):
         self.type = {}
 
     def transform_shard(self, func, *args):
-        transformed_shard = SparkXShards(self.rdd.map(lambda data: func(data, *args)))
+        def transform(iter, func, *args):
+            for x in iter:
+                yield func(x, *args)
```
jason-dai, line 149: you should not use a loop; the point is that `func` should be executed only once.

cyita:
If data in XShard isn't a dataframe, transform_shard can't be called after repartition?

pyzoo/zoo/orca/data/shard.py
```diff
@@ -180,7 +189,19 @@ def num_partitions(self):
...
+        if self._get_class_name() == 'pandas.core.frame.DataFrame':
+            import pandas as pd
+
+            def combine(iter):
+                dfs = list(iter)
+                if len(dfs) > 0:
+                    return [pd.concat(dfs)]
+                else:
+                    return iter
+            new_rdd = self.rdd.repartition(num_partitions).mapPartitions(combine)
```
jason-dai, line 201: If each partition only contains one pandas dataframe, does `rdd.repartition` work?

jenniew: It can work. But if repartition from small partitions to large partitions, some partitions would be empty.

jenniew:
> If data in XShard isn't a dataframe, transform_shard can't be called after repartition?

I think so since there may be multiple elements in one partition.

jenniew:
Jenkins pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/4234/

PR #2644, opened 2020-07-30
Revert "Add TxT import"

hkvision:
Reverts intel-analytics/analytics-zoo#2643


hkvision:
Jenkins doesn't install mxnet; and thus can't put it in init...

Issue #2663, opened 2020-08-03
Add serialize_data_creation flag in PyTorchTrainer after upgraded to ray 0.8.6+


hkvision:
Currently when loading data for 0.8.4, ray would default add a file lock and thus if we are not downloading data, most likely the data loading on one machine is not parallel.
Ray 0.8.6+ has fixed this issue; make this optional.

hkvision:
Better upgrade to 0.8.7 as num_cpus of each runner can be set.

hkvision:
We have aborted the dependency of RaySGD. Closing this issue.

PR #2669, opened 2020-08-04
Orca PyTorch Estimator load data once

cyita:


pyzoo/zoo/examples/orca/learn/horovod/pytorch_estimator.py
```diff
@@ -78,6 +78,8 @@ def validation_data_creator(config):
 
 
 def train_example():
+    import horovod.torch as hvd
+    hvd.init()
```
jason-dai, line 82: we don't want to ask the user to import horovod explicitly.

pyzoo/zoo/orca/learn/pytorch/pytorch_horovod_estimator.py
```diff
@@ -215,6 +218,7 @@ def setup_horovod():
 
     def train(self,
               data_creator,
+              epochs=1,
               num_steps=None,
```
jason-dai, line 222: why do we support both epochs and steps? which one is supported in tf.estimator?

cyita: use epochs instead of num_steps.

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
```diff
@@ -156,9 +156,26 @@ def should_wrap_dataloader(loader):
         return (isinstance(loader, DataLoader)
                 and not isinstance(loader.dataset, IterableDataset))
 
+    def train_epochs(self, data_creator, epochs=1, profile=False, info=None):
+        should_wrap = False
+        with FileLock(
+                os.path.join(tempfile.gettempdir(), ".orcadata.lock")):
```
hkvision, line 162: We'd better add an option to use this file lock, refer to the implementation of ray and this issue: https://github.com/intel-analytics/analytics-zoo/issues/2663

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
```diff
@@ -156,9 +156,26 @@ def should_wrap_dataloader(loader):
...
+            loader = data_creator(self.config)
+        if TorchRunner.should_wrap_dataloader(loader):
+            loader = self.with_sampler(loader)
+            should_wrap = True
```
hkvision, line 166: why not put `iter(loader)` here? is should_wrap necessary?

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
```diff
@@ -156,9 +156,20 @@ def should_wrap_dataloader(loader):
         return (isinstance(loader, DataLoader)
                 and not isinstance(loader.dataset, IterableDataset))
 
+    def train_epochs(self, data_creator, epochs=1, profile=False, info=None):
+        with FileLock(
+                os.path.join(tempfile.gettempdir(), ".orcadata.lock")):
+            loader = data_creator(self.config)
+        if TorchRunner.should_wrap_dataloader(loader):
+            loader = self.with_sampler(loader)
```
hkvision, line 164: Shouldn't remove iter? I mean combine the iter here?

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
cyita: If I use iter(), a new iterator from the start should be used in different epochs.

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
cyita: Do you mean adding a flag in OrcaContext?

jason-dai: Shall we simply upgrade to Ray 0.8.7?

jason-dai:
change `from_model_creator` to `from_torch`

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
hkvision: @cyita No, just an argument in Estimator. Check here: https://github.com/ray-project/ray/blob/master/python/ray/util/sgd/torch/torch_runner.py#L92

hkvision: @jason-dai But this file is copied from ray, we still need to modify this file even we upgrade ray.

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
hkvision: OK. So what's the behavoir without using iter()?

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
jason-dai: > @cyita No, just an argument in Estimator. Check here: https://github.com/ray-project/ray/blob/master/python/ray/util/sgd/torch/torch_runner.py#L92

I don't think the user will want to pass this argument? Maybe we can just include it in OrcaContext?

hkvision: Then we would make default not to add the file lock. Users would hardly download all the data in parallel.

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
cyita: Since train_operator uses `for batch_idx, batch in enumerate(iterator):`, the behavior will be the same as using iter()

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
yushan111: Maybe iter(loader) should be inside the for loop of train epochs, since each epoch exhausts the iterator.

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
yushan111: should wrap is necessary since it should not wrap a DistributedSampler for the dataloader if its dataset is an IterableDataset.  Sampler is only compatible with may-style datasets.

hkvision: Isn't TorchRunner.should_wrap_dataloader(loader) == should_wrap? Why do we still need a flag?

yushan111: I misunderstood. The flag is useless.

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
hkvision: So we can put with_sampler and iter in train_epoch?

cyita: Yes.

yushan111: I think with_sampler is a way to set up loader so maybe it is not appropriate to call every epoch.

jason-dai: I think with_sampler should be called only once @yangw1234

cyita: Move iter in train_epoch and with_sampler outside the loop.

PR #2672, opened 2020-08-05
Support metric tensors in keras model

yangw1234:


yangw1234:
jenkins passed and manually tested it, ut will be added later

Issue #2673, opened 2020-08-05
add support page in doc website with support info


helenlly:
e.g. support  Colab  or not?  which pytorch/ ray/..... version  does AZ support? 

refer to https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!searchin/bigdl-user-group/jiahaoyu04@gmail.com%7Csort:date/bigdl-user-group/b5yCGWIBjpk/PhCF56d-BAAJ

helenlly:
@Le-Zheng , can we close the issue?

Le-Zheng:
> @Le-Zheng , can we close the issue?

Sure

PR #2684, opened 2020-08-06
add python support library version page in doc

Le-Zheng:
To solve https://github.com/intel-analytics/analytics-zoo/issues/2673


docs/docs/ProgrammingGuide/python-lib-version.md
```diff
@@ -0,0 +1,195 @@
+This page documents and describes Python library versions that Analytics Zoo supports. The Python versions we have tested are Python 3.6 and Python 3.7.
+
+```
+Name & version
+
+absl-py==0.9.0
```
hkvision, line 6: Do we need to list all the packages? Actually for many packages there are no version restricts and those versions are only the versions on your machines.
What about list only the versions of the key packages; for example, tf, torch, ray, etc...

docs/docs/ProgrammingGuide/run-notebook-colab.md
```diff
@@ -57,3 +57,7 @@ If you would like to open Analytics Zoo Notebook in a GitHub repo directly, the
   to [https://colab.research.google.com/github/intel-analytics/zoo-tutorials/blob/master/keras/2.1-a-first-look-at-a-neural-network.ipynb](https://colab.research.google.com/github/intel-analytics/zoo-tutorials/blob/master/keras/2.1-a-first-look-at-a-neural-network.ipynb).
 
 Then, prepare the environment of Java8 and Analytics Zoo as described [above](#prepare-environment) at the beginning of the colab notebook. If you would like to save the changes, you can make a copy to drive and run it within the instructions.
+
+**Note**
+
+As Colab engine provides your application with some built-in Python libraries, you should check if the library versions are compatible with the application. You may refer this [document](./python-lib-version.md) to specify the python library version that Analytics Zoo supports.
```
qiyuangong, line 63: `provides some built-in` is enough.
`compatible with your application` maybe better.

PR #2685, opened 2020-08-06
Add init_spark_standalone for local node

hkvision:
If directly call `start-all.sh`, even for localhost, ssh is needed. So I use `start-master.sh` and `start-worker.sh` instead to avoid ssh commands.


pyzoo/zoo/util/spark.py
```diff
@@ -219,3 +228,86 @@ def _submit_opt():
...
+        submit_args = " --master " + master
+        submit_args = submit_args + " --driver-cores {} --driver-memory {} --num-executors {}" \
+                                    " --executor-cores {} --executor-memory {}"\
+            .format(driver_cores, driver_memory, num_executors, executor_cores, executor_memory)
+        if extra_python_lib:
+            submit_args = submit_args + " --py-files {}".format(extra_python_lib)
+        if jars:
+            submit_args = submit_args + " --jars {}".format(jars)
+        submit_args = submit_args + " pyspark-shell"
+        os.environ['PYSPARK_SUBMIT_ARGS'] = submit_args
```
jason-dai, line 280: how is 'PYSPARK_SUBMIT_ARGS' used?

scripts/standalone/sbin/start-worker.sh
```diff
@@ -0,0 +1,92 @@
...
+  WEBUI_PORT=$(( $SPARK_WORKER_WEBUI_PORT + $WORKER_NUM - 1 ))
+
+  "${ZOO_STANDALONE_HOME}/sbin"/spark-daemon.sh start $CLASS $WORKER_NUM \
+     --webui-port "$WEBUI_PORT" $PORT_FLAG $PORT_NUM $MASTER "$@"
+}
+
+if [ "$SPARK_WORKER_INSTANCES" = "" ]; then
+  start_instance 1 "$@"
+else
+  for ((i=0; i<$SPARK_WORKER_INSTANCES; i++)); do
```
jason-dai, line 89: how is this value set? this should be set according to `num_executors`

pyzoo/zoo/util/spark.py
```diff
@@ -219,3 +228,86 @@ def _submit_opt():
             if conda_name and penv_archive and pack_env:
                 os.remove(penv_archive)
         return sc
+
+    def init_spark_standalone(self,
+                              num_executors,
```
jason-dai, line 233: I think this needs to be set to $SPARK_WORKER_INSTANCES, yes?

pyzoo/zoo/util/spark.py
```diff
@@ -219,3 +228,86 @@ def _submit_opt():
...
+                              executor_cores,
```
jason-dai, line 234: I think this value needs to be set to $SPARK_WORKER_CORES, yes?

pyzoo/zoo/util/spark.py
```diff
@@ -219,3 +228,86 @@ def _submit_opt():
...
+                              executor_memory="10g",
+                              driver_memory="1g",
```
jason-dai, line 236: how to set $SPARK_WORKER_MEMORY?

jason-dai:
Instead of copying the spark scripts, can we use the ones inside pyspark package?

hkvision:
> Instead of copying the spark scripts, can we use the ones inside pyspark package?

The scripts are under `sbin`, but pyspark doesn't have that folder. They are only available if download the full spark package.

pyzoo/zoo/util/spark.py
hkvision: This is the environment variable, I suppose Spark would read this when creating the sc. This implementation is the same as init_spark_on_yarn.

scripts/standalone/sbin/start-worker.sh
hkvision: I think the worker instance and executors are two things. For each node, we only start a worker instance, but within one worker instance, we can start multiple executors, right? So here on a single node, worker is always 1 and executors can be more than 1.

pyzoo/zoo/util/spark.py
hkvision: https://spark.apache.org/docs/latest/spark-standalone.html#starting-a-cluster-manually
The default configurations allocates all available cores and (total mem - 1GB) for the worker. But actually how many cores and memory are used is determined by driver and executor settings, isn't that reasonable?

hkvision:
I have an extra concerns about this: If there's some error in user's code, the program exits and the standalone cluster may not be shutdown. Do we need to provide a shutdown command for users to run like what we do in cluster serving? Or a shutdown hook like what we do in ray?

Also do we need to extract a method for users that already have a standalone cluster? This may not be necessary for tianchi, but would probably be useful for other users.

@jason-dai

scripts/standalone/sbin/start-worker.sh
jason-dai: We want to run each executor in a separate JVM; does it work with only one worker?

hkvision: I will check this.

hkvision:
```
root      45843  45785 72 14:04 pts/0    00:00:12 /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -cp /opt/anaconda3/envs/standalone/lib/python3.6/site-packages/pyspark/conf:/opt/anaconda3/envs/standalone/lib/python3.6/site-packages/pyspark/jars/* -Xmx1g org.apache.spark.deploy.SparkSubmit --conf spark.master spark://172.168.2.180:7077 --conf spark.executor.cores=4 --conf spark.app.name=pi --conf spark.cores.max=16 pyspark-shell
root      46073  44433 47 14:04 pts/0    00:00:07 /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -cp /opt/kai/spark-2.4.3-bin-hadoop2.7/conf/:/opt/kai/spark-2.4.3-bin-hadoop2.7/jars/* -Xmx1024M -Dspark.driver.port=46116 org.apache.spark.executor.CoarseGrainedExecutorBackend --driver-url spark://CoarseGrainedScheduler@Gondolin-Node-080:46116 --executor-id 1 --hostname 172.168.2.180 --cores 4 --app-id app-20200807140444-0001 --worker-url spark://Worker@172.168.2.180:36312
root      46077  44433 48 14:04 pts/0    00:00:07 /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -cp /opt/kai/spark-2.4.3-bin-hadoop2.7/conf/:/opt/kai/spark-2.4.3-bin-hadoop2.7/jars/* -Xmx1024M -Dspark.driver.port=46116 org.apache.spark.executor.CoarseGrainedExecutorBackend --driver-url spark://CoarseGrainedScheduler@Gondolin-Node-080:46116 --executor-id 0 --hostname 172.168.2.180 --cores 4 --app-id app-20200807140444-0001 --worker-url spark://Worker@172.168.2.180:36312
root      46083  44433 51 14:04 pts/0    00:00:07 /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -cp /opt/kai/spark-2.4.3-bin-hadoop2.7/conf/:/opt/kai/spark-2.4.3-bin-hadoop2.7/jars/* -Xmx1024M -Dspark.driver.port=46116 org.apache.spark.executor.CoarseGrainedExecutorBackend --driver-url spark://CoarseGrainedScheduler@Gondolin-Node-080:46116 --executor-id 2 --hostname 172.168.2.180 --cores 4 --app-id app-20200807140444-0001 --worker-url spark://Worker@172.168.2.180:36312
root      46202  44433 56 14:04 pts/0    00:00:08 /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -cp /opt/kai/spark-2.4.3-bin-hadoop2.7/conf/:/opt/kai/spark-2.4.3-bin-hadoop2.7/jars/* -Xmx1024M -Dspark.driver.port=46116 org.apache.spark.executor.CoarseGrainedExecutorBackend --driver-url spark://CoarseGrainedScheduler@Gondolin-Node-080:46116 --executor-id 3 --hostname 172.168.2.180 --cores 4 --app-id app-20200807140444-0001 --worker-url spark://Worker@172.168.2.180:36312
```
4 separate JVMs right? @jason-dai

pyzoo/zoo/ray/raycontext.py
```diff
@@ -55,9 +55,9 @@ class RayServiceFuncGenerator(object):
     """
     def _prepare_env(self):
         modified_env = os.environ.copy()
-        cwd = os.getcwd()
-        modified_env["PATH"] = "{}/{}:{}".format(cwd, "/".join(self.python_loc.split("/")[:-1]),
-                                                 os.environ["PATH"])
+        if "PATH" in os.environ:
```
hkvision, line 58: "PATH" is removed by Spark in os.environ for standalone (probably also for k8s)?
```
def check_env(x):
   import os
   return "PATH" in os.environ


print("PATH" in os.environ)  # True
result = sc.parallelize(range(0, 4)).map(check_env).collect()
print(result)  # [False, False, False, False]
```

hkvision: cc @jason-dai

hkvision: Seems this only happens on the same node. Previously for multiple nodes, there is no such question.

hkvision:
```
This system supports the C.UTF-8 locale which is recommended. You might be able to resolve your issue by exporting the following environment variables:

    export LC_ALL=C.UTF-8
    export LANG=C.UTF-8
```
This error occurred when initializing ray. Need to add to env when creating SparkContext. Don't know why...
Also this only happens on the single node. Previously for multiple nodes, there is no such issue.

scripts/standalone/sbin/start-worker.sh
jason-dai: I think it's OK for now. If we want to use numactl in future, we will need to launch multiple workers using numactl in the scripts.

jason-dai:
> ```
> This system supports the C.UTF-8 locale which is recommended. You might be able to resolve your issue by exporting the following environment variables:
> 
>     export LC_ALL=C.UTF-8
>     export LANG=C.UTF-8
> ```
> 
> This error occurred when initializing ray. Need to add to env when creating SparkContext. Don't know why...
> Also this only happens on the single node. Previously for multiple nodes, there is no such issue.
> 
> Do we need to add this to our code by default?

We will need to support this as we do want to support RayOnSpark.

hkvision:
> do

But actually I don't know the root cause of this error. It only happens on local standalone (both reproduced on my desktop and Gondolin server). Any ideas?

hkvision:
> > ```
> > This system supports the C.UTF-8 locale which is recommended. You might be able to resolve your issue by exporting the following environment variables:
> > 
> >     export LC_ALL=C.UTF-8
> >     export LANG=C.UTF-8
> > ```
> > 
> > 
> > This error occurred when initializing ray. Need to add to env when creating SparkContext. Don't know why...
> > Also this only happens on the single node. Previously for multiple nodes, there is no such issue.
> > Do we need to add this to our code by default?
> 
> We will need to support this as we do want to support RayOnSpark.

But actually I don't know the root cause of this error. It only happens on local standalone (both reproduced on my desktop and Gondolin server). Any ideas?

scripts/standalone/sbin/start-worker.sh
hkvision: OK. Then keep this for the first version. Would update later for support numactl.

pyzoo/zoo/ray/raycontext.py
```diff
@@ -55,9 +55,19 @@ class RayServiceFuncGenerator(object):
...
+            executor_python_path = "{}/{}".format(
+                os.getcwd(), "/".join(self.python_loc.split("/")[:-1]))
+        else:
+            executor_python_path = "/".join(self.python_loc.split("/")[:-1])
+        if "PATH" in os.environ:
+            modified_env["PATH"] = "{}:{}".format(executor_python_path, os.environ["PATH"])
+        else:
+            modified_env["PATH"] = executor_python_path
+        modified_env["LC_ALL"] = "C.UTF-8"
+        modified_env["LANG"] = "C.UTF-8"
```
hkvision, line 70: I add the two environment variables here. Is it OK? I would test whether it impacts yarn later. @jason-dai

jason-dai: > I add the two environment variables here. Is it OK? I would test whether it impacts yarn later. @jason-dai

ok

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/3962/
I'll merge it first. If there are further issues, would fix in a following PR.

PR #2686, opened 2020-08-06
Support dropping remainder batch in TFPark

yangw1234:
Some TensorFlow models require the batch dimension to be set to a specific value. In such cases, the last batch may not have enough samples to form a full batch in validation and we have to drop it.


yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/3936/

PR #2687, opened 2020-08-07
set default nthread for xgboost

dding3:


dding3:
jekins passed. http://10.239.47.210:18888/job/ZOO-PR-Validation/3997/

PR #2716, opened 2020-08-12
fix to pytorch throws exception when model is frozen.

qiuxin2012:
fix to pytorch throws exception when model is frozen.


qiuxin2012:
jenkins passed.

PR #2719, opened 2020-08-13
Horovod Estimator validate check serialize_data_creation

### LINKED ISSUE #2714: serialize_data_creation is not used in validate of Horovod Estimator
Need to make the Lock optional like train.

cyita:
Fix #2714

### DIFF
--- a/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
+++ b/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
@@ -200,8 +200,11 @@ class TorchRunner:
         self._toggle_profiling(profile=profile)
 
         with self.timers.record("validation"):
-            with FileLock(
-                    os.path.join(tempfile.gettempdir(), ".orcadata.lock")):
+            if OrcaContext.serialize_data_creation:
+                with FileLock(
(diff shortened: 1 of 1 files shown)


pyzoo/zoo/orca/learn/pytorch/torch_runner.py
```diff
@@ -200,8 +200,11 @@ def validate(self, data_creator, num_steps=None, profile=False, info=None):
         self._toggle_profiling(profile=profile)
 
         with self.timers.record("validation"):
```
hkvision, line 202: Shall we put the timer include data loading or just validation?

cyita: Maybe just validation is better since the timer in train_epoch just includes validation.

hkvision: But now the data loading time is also recorded here, shall we move it out?
For training timer, it only calculates training time excluding data time.

cyita:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/4004/

PR #2724, opened 2020-08-13
Orca tf estimator save model

jenniew:
Add save() method in Orca tf estimator. 
For graph model, it would save current session to tf checkpoint.
For keras model, it would save wrapped keras model to file system.

Also change load_checkpoint() to load_orca_checkpoint().
Change load_latest_checkpoint() to load_latest_orca_checkpoint()

### DIFF
--- a/pyzoo/zoo/orca/learn/tf/estimator.py
+++ b/pyzoo/zoo/orca/learn/tf/estimator.py
@@ -169,15 +169,17 @@ class Estimator(object):
 
     def save_tf_checkpoint(self, path):
         """
-        Save tensorflow checkpoint model in this estimator.
-        For tensorflow graph model, it would save the checkpoint.
-        For tensorflow keras model, it would save keras model.
-        :param path: save path.
+        Save tensorflow checkpoint in this estimator.
+        :param path: tensorflow checkpoint path.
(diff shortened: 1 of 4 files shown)


jason-dai:
maybe separate to two different APIs: save_keras_model and save_tf_checkpoint?

jenniew:
> maybe separate to two different APIs: save_keras_model and save_tf_checkpoint?

updated

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -401,9 +404,12 @@ def evaluate(self, data, batch_size=32,
                                 sess=self.sess,
                                 dataset=dataset, metrics=self.metrics)
 
-    def save(self, path):
+    def save_tf_checkpoint(self, path):
         save_tf_checkpoint(self.sess, path)
 
+    def save_keras_model(self, path):
+        raise Exception("No keras model in this estimator")
```
jason-dai, line 411: NotImplementedError, maybe in parent class?

jenniew:
Jenkins pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/4049/

PR #2728, opened 2020-08-14
Cluster Serving flink distributed to remote and load local model

Litchilitchy:
Whether OpenVINO model could be broadcast successfully does not matter now.

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
@@ -19,7 +19,7 @@ package com.intel.analytics.zoo.serving
 
 
 import com.intel.analytics.zoo.serving.engine.{FlinkInference, FlinkRedisSink, FlinkRedisSource}
-import com.intel.analytics.zoo.serving.utils.{ClusterServingHelper, ClusterServingManager, SerParams}
+import com.intel.analytics.zoo.serving.utils.{ClusterServingHelper, ClusterServingManager, Conventions, SerParams}
 import org.apache.flink.streaming.api.scala.{StreamExecutionEnvironment, _}
 import org.apache.log4j.{Level, Logger}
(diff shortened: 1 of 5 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -47,7 +47,7 @@ object ClusterServing {
     helper.initArgs()
     params = new SerParams(helper)
     val serving = StreamExecutionEnvironment.getExecutionEnvironment
-
+    serving.registerCachedFile(params.modelDir, "cluster-serving-model")
```
glorysdj, line 50: put cluster-serving-model to your Conventions?

zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkInference.scala
```diff
@@ -20,22 +20,26 @@ import com.intel.analytics.bigdl.tensor.Tensor
...
+
   var t: Tensor[Float] = null
   var logger: Logger = null
   var inferenceCnt: Int = 0
   var pre: PreProcessing = null
   var post: PostProcessing = null
 
   override def open(parameters: Configuration): Unit = {
+    val myDir = getRuntimeContext.getDistributedCache.getFile("cluster-serving-model").getPath
+    println(s"Model is distributed to executor path ${myDir}")
```
glorysdj, line 41: change println to logger debug?

glorysdj:
need to test on cluster mode

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
Litchilitchy: noted

Litchilitchy:
@glorysdj test passed

PR #2747, opened 2020-08-18
update http to add pass secret and salt to cluster serving

glorysdj:
update http to add pass secret and salt to cluster serving
user can put secret and salt to redis without redis-cli

---
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/analytics-zoo/2747)

### DIFF
--- a/zoo/src/main/scala/com/intel/analytics/zoo/serving/http/FrontEndApp.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/serving/http/FrontEndApp.scala
@@ -137,14 +137,14 @@ object FrontEndApp extends Supportive with EncryptSupportive {
             }).toList
             complete(jacksonJsonSerializer.serialize(servingMetrics))
           }
-        } ~ (path("model-secure") & parameters('secret.as[String], 'salt.as[String])) {
-          (secret, salt) => {
+        } ~ (post & path("model-secure") &
+          extract(_.request.entity.contentType) & entity(as[String])) {
+          (contentType, content) => {
             try {
-              val dycryptSecret = decryptWithAES256(secret, Conventions.INTERNAL_SECRET,
-                Conventions.INTERNAL_SALT)
-              val dycryptSalt = decryptWithAES256(salt, Conventions.INTERNAL_SECRET,
-                Conventions.INTERNAL_SALT)
-              val message = SecuredModelSecretSaltMessage(dycryptSecret, dycryptSalt)
+              val secrets = content.split("&")
+              val secret = secrets(0).split("=")(1)
+              val salt = secrets(1).split("=")(1)
+              val message = SecuredModelSecretSaltMessage(secret, salt)
               val result = Await.result(redisPutter ? message, timeout.duration)
                 .asInstanceOf[Boolean]
               result match {
@@ -154,7 +154,8 @@ object FrontEndApp extends Supportive with EncryptSupportive {
             } catch {
               case e: Exception =>
                 e.printStackTrace()
-                val error = ServingError(e.getMessage)
+                val error = ServingError(e.getMessage + "\n please post a content like " +
+                  "secret=xxx&salt=xxxx")
                 complete(500, error.toString)
             }
 
--- a/zoo/src/main/scala/com/intel/analytics/zoo/serving/http/FrontEndApp.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/serving/http/FrontEndApp.scala
@@ -137,9 +137,8 @@ object FrontEndApp extends Supportive with EncryptSupportive {
             }).toList
             complete(jacksonJsonSerializer.serialize(servingMetrics))
           }
-        } ~(path("model-secure") & parameters('secret.as[String], 'salt.as[String])) {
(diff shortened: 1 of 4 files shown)


Litchilitchy:
@qiyuangong Please double review the behavior of 
```
              val dycryptSecret = decryptWithAES256(secret, Conventions.INTERNAL_SECRET,
                Conventions.INTERNAL_SALT)
              val dycryptSalt = decryptWithAES256(salt, Conventions.INTERNAL_SECRET,
                Conventions.INTERNAL_SALT)
```
`Conventions` is an opened variable

qiyuangong:
> @qiyuangong Please double review the behavior of
> 
> ```
>               val dycryptSecret = decryptWithAES256(secret, Conventions.INTERNAL_SECRET,
>                 Conventions.INTERNAL_SALT)
>               val dycryptSalt = decryptWithAES256(salt, Conventions.INTERNAL_SECRET,
>                 Conventions.INTERNAL_SALT)
> ```
> 
> `Conventions` is an opened variable

OK. Will check details.

zoo/src/main/scala/com/intel/analytics/zoo/serving/http/actors.scala
```diff
@@ -96,6 +97,12 @@ class RedisPutActor(
           }
         }
       }
+    case message: SecuredModelSecretSaltMessage =>
+      silent(s"$actorName put secret and salt in redis")() {
+        jedis.hset(Conventions.MODEL_SECURED_KEY, Conventions.MODEL_SECURED_SECRET, message.secret)
```
qiyuangong, line 102: If we add multi-model support, we also need a model ID to avoid conflict.

zoo/src/main/scala/com/intel/analytics/zoo/serving/http/FrontEndApp.scala
```diff
@@ -136,6 +137,28 @@ object FrontEndApp extends Supportive {
             }).toList
             complete(jacksonJsonSerializer.serialize(servingMetrics))
           }
+        } ~ (path("model-secure") & parameters('secret.as[String], 'salt.as[String])) {
+          (secret, salt) => {
+            try {
+              val dycryptSecret = decryptWithAES256(secret, Conventions.INTERNAL_SECRET,
```
qiyuangong, line 143: I don't think we need this encryption and decryption. Because our key and salt are public in code.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/Conventions.scala
```diff
@@ -31,4 +31,7 @@ object Conventions {
   val MODEL_SECURED_KEY = "model_secured"
   val MODEL_SECURED_SECRET = "secret"
   val MODEL_SECURED_SALT = "salt"
+
+  val INTERNAL_SECRET = "1234qwer"
```
qiyuangong, line 35: I think we can make it configurable (from file) or just remove it. :)

zoo/src/main/scala/com/intel/analytics/zoo/serving/http/FrontEndApp.scala
glorysdj: we need to pass the secret and salt to redis, same as model, we can not put plain text in config file or in the command to start the program. we just need the secret and salt to be encrypted.

zoo/src/main/scala/com/intel/analytics/zoo/serving/http/actors.scala
glorysdj: in a easier way, we required the models are encrypted with same secret and salt.

glorysdj: and for now, each cluster serving job is deployed with one single model. in the future, we can config each model or each version of model with meta data including encrypted info.

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/Conventions.scala
glorysdj: we need a secret/salt, even the info is put in config file, it should be encrypted. this internal secret/salt is used to decrypt the encrypted secret/salt...

qiyuangong: > we need a secret/salt, even the info is put in config file, it should be encrypted. this internal secret/salt is used to decrypt the encrypted secret/salt...

Yes. These key and salt are necessary. But, it should be allowed to change. Just like SSL keystore password or Spark RPC secret (https://spark.apache.org/docs/latest/security.html#spark-rpc-communication-protocol-between-spark-processes). We can make it as property, and change it if necessary (configure file or keyfactory).

glorysdj: ok. i will change the implementation.

qiyuangong (approved):
LGTM

glorysdj:
jenkins passed http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/4043/

PR #2757, opened 2020-08-20
[Horovod] Support multiple workers in one spark executor

### LINKED ISSUE #2703: Horovod for Spark local can only start one worker
Better make num_workers configurable. Otherwise running in local doesn't make any sense.

yangw1234:
Fix #2703

### DIFF
--- a/pyzoo/zoo/orca/learn/horovod/horovod_ray_runner.py
+++ b/pyzoo/zoo/orca/learn/horovod/horovod_ray_runner.py
@@ -64,11 +64,11 @@ def _hosts_to_hosts_spec(hosts):
 class HorovodRayRunner:
 
     # todo check whether horovod is built with gloo
-    def __init__(self, ray_ctx, worker_cls=None, worker_param=None):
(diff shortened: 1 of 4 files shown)


hkvision:
What about just specifying num_workers and worker_cores / worker_cpus? Sometimes users don't necessarily need to use all the yarn cluster resources?

yangw1234:
> What about just specifying num_workers and worker_cores / worker_cpus? Sometimes users don't necessarily need to use all the yarn cluster resources?
If we do not use all the yarn cluster resources, some workers may not have tasks at all. Does it work with xshards in this case?

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/4052/ jenkins passed

also tested manually on resnet50

yangw1234:
@hkvision  I'll first merge it. Let's discuss the num_workers issues later.

hkvision:
LGTM. For XShards, I think it requires there are workers on each node. But if users don't use XShards, it is still possible to have fewer workers?

PR #2759, opened 2020-08-21
Cluster Serving Flink operator inference model to singleton class

Litchilitchy:


zoo/src/main/scala/com/intel/analytics/zoo/serving/engine/FlinkInference.scala
```diff
@@ -37,30 +37,41 @@ class FlinkInference(params: SerParams)
...
     logger = Logger.getLogger(getClass)
 
-    val localModelDir = getRuntimeContext.getDistributedCache
-      .getFile(Conventions.SERVING_MODEL_TMP_DIR).getPath
-    val localConfPath = getRuntimeContext.getDistributedCache
-      .getFile(Conventions.SERVING_CONF_TMP_PATH).getPath
-    logger.info(s"Config parameters loaded at executor at path ${localConfPath}, " +
-      s"Model loaded at executor at path ${localModelDir}")
-    params.model = ClusterServingHelper.loadModelfromDir(localConfPath, localModelDir)
+    if (!FlinkInference.initialized) {
```
glorysdj, line 40: this is not safe, two thread may run into this line at the same time. please refer to a standard singleton implementation.

Litchilitchy: switched to double-check synchronized

PR #2776, opened 2020-08-26
tf distributed behavior consistent with horovod

yangw1234:


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/4129/ jenkins passed except on irrelevant error
horovod ut tested locally

PR #2784, opened 2020-08-27
Add columns and repartition parameter in read_parquet

jenniew:
Add columns options in read_parquet if user only needs a few columns among so many columns.  Load all columns and cache to xshards may have memory issue.

Add repartition option if user does not want to repartition when num_partition on xshards < node_number. For big dataset, repartition may take a lot effort if users just do prediction/evaluation.

### DIFF
--- a/pyzoo/zoo/orca/data/pandas/preprocessing.py
+++ b/pyzoo/zoo/orca/data/pandas/preprocessing.py
@@ -265,28 +265,22 @@ def read_file_spark(file_path, file_type, **kwargs):
     return data_shards
 
 
-def read_parquet(file_path, repartition=True, columns=None, **kwargs):
+def read_parquet(file_path, columns=None, **kwargs):
     """
     Read parquet files to SparkXShards of pandas DataFrames.
 
     :param file_path: Parquet file path, a list of multiple parquet file paths, or a directory
     containing parquet files. Local file system, HDFS, and AWS S3 are supported.
-    :param repartition: Whether to repartition if number of partitions in this XShards is less than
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/orca/data/pandas/preprocessing.py
```diff
@@ -265,12 +265,17 @@ def f(iter):
     return data_shards
 
 
-def read_parquet(file_path,):
+def read_parquet(file_path, repartition=True, columns=None, **kwargs):
```
jason-dai, line 268: how does the user know if repartition is needed?

jenniew: If he wants to balance load to spark executors?

jason-dai: I think either we do the repartitioning automatically, or the user can just call repartition() explicitly?

jack1981: The repartition is a very expensive operation in real PROD , big data in big cluster.
We have spark code check list , the repartition have to be be called explicitly.
The current API will do it internally and automatically , we hope there is a way to control the behavior by user.
The repartition=True is ok and user can change repartition=False and handle any partition operation outside the API.

jenniew: Do we remove repartition operation in read_parquet and let user do it explicitly?

jenniew:
Jenkins got error but not related this PR.

PR #2787, opened 2020-08-27
Update UTs and examples with init_orca_context

hkvision:
- Update all UTs and examples in Orca.
- Update README.
- Fix some import ray issues.


hkvision:
What about making default cluster_mode to "local"? As it would be more commonly used. @jason-dai

jason-dai:
If we change default to "local", maybe we should use "spark" (or "spark_submit") instead of "" when user needs to manually  start the spark job?

hkvision:
> If we change default to "local", maybe we should use "spark" (or "spark_submit") instead of "" when user needs to manually start the spark job?

Yeah, makes sense. I would change that.

hkvision:
Tested the examples manually on yarn.

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/4126/

PR #2824, opened 2020-09-04
Cluster Serving Timer

Litchilitchy:
This is timer which is not thread safe, thus only used in sequential pipeline. Noted multi-thread timer is not useful, because timing should be analyzed in sequential pipeline but not parallel operations.

Want to use dependencies as few as possible, thus did not use accumulators.


glorysdj:
https://github.com/intel-analytics/analytics-zoo/blob/master/zoo/src/main/scala/com/intel/analytics/zoo/serving/http/FrontEndApp.scala#L239

you can use com.codahale.metrics.MetricRegistry Timers
please also refer to https://github.com/intel-analytics/analytics-zoo/blob/master/zoo/src/main/scala/com/intel/analytics/zoo/serving/http/Supportive.scala

Issue #2830, opened 2020-09-05
Add over_write option at zoo.orca.learn.tf.estimator.save_keras_model()


jack1981:
def save_keras_model(self, path):
        self.model.save_model(path)

It is better to add "over_write" option at  the function as sometime makes sense to over write model

cyita:
Thanks for your suggestion. We will add the "over_write" option.

helenlly:
Hi Jack,

the fix is ready. pls let us know if any more questions or if we can close this issue. thanks.

PR #2835, opened 2020-09-07
Orca tf save_keras_model add overwrite option

cyita:
Related issue: #2830

### DIFF
--- a/pyzoo/zoo/orca/learn/tf/estimator.py
+++ b/pyzoo/zoo/orca/learn/tf/estimator.py
@@ -187,7 +187,7 @@ class Estimator(object):
         """
         Save tensorflow keras model in this estimator.
         :param path: keras model save path.
-        :param overwrite: Whether overwrite any existing model at the target location
+        :param overwrite: Whether to silently overwrite any existing file at the target location.
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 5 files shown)


pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -183,10 +183,11 @@ def save_tf_checkpoint(self, path):
         """
         raise NotImplementedError()
 
-    def save_keras_model(self, path):
+    def save_keras_model(self, path, overwrite):
```
jenniew, line 186: Why here you don't give default value for overwrite option?

pyzoo/zoo/tfpark/model.py
```diff
@@ -64,15 +64,17 @@ def load_func(file_path):
             self.model.load_weights(file_path, by_name)
         load_from_file(load_func, filepath)
 
-    def save_model(self, path):
+    def save_model(self, path, overwrite):
```
jenniew, line 67: We may give default value of overwrite.

jenniew:
Also update docs.

cyita:
Jenkins passed http://10.239.47.210:18888/view/ZOO-PR-PIPELINE/job/ZOO-PR-Validation/4158/

jason-dai:
Should overwrite be true or false by default?

cyita:
> Should overwrite be true or false by default?

The default value of overwrite in tf.keras.save is True.

jason-dai:
> > Should overwrite be true or false by default?
> 
> The default value of overwrite in tf.keras.save is True.

OK, LGTM

Issue #2840, opened 2020-09-08
AutoML: LSTMGridRandomRecipe is not considering all the features combination.


remisharoon:
LSTMGridRandomRecipe (in recipe.py) is not considering all the features combination as `np.random.randint` method's upper(high) limit is non-inclusive. This is causing the exclusion of all-features combination.

PR #2841, opened 2020-09-08
#2840 - AutoML: LSTMGridRandomRecipe is not considering all the features combination

remisharoon:
AutoML: LSTMGridRandomRecipe is not considering all the features combination. #2840

LSTMGridRandomRecipe (in recipe.py) is not considering all the features combination as np.random.randint method's upper(high) limit is non-inclusive. This is causing the exclusion of all-features combination.

np.random.randint method's upper(high) limit is non-inclusive, so adding one to high, to make it inclusive


yushan111 (approved):

yushan111:
Thank you for your contribution! I will merge the PR as soon as jenkins passed.

PR #2836, opened 2020-09-07
Fix RayOnSpark for k8s and add docs

hkvision:
- Fix issues for running RayOnSpark on k8s cluster (described below). (Tested on PyTorchTrainer and MXNet Estimator)
- Add inline docs for init_spark_on_k8s.
- Some code updates of Ray related. Move init_executor_gateway to ray_ctx.init().
- Wrap k8s into init_orca_context.

PR #2843, opened 2020-09-08
Move init_executor_gateway to ray_ctx.init

hkvision:
- Tested on yarn that running init_executor_gateway for multiple times is ok.
- So that creating a RayContext won't bring any side effects.


hkvision:
Would be combined in #2836

PR #2844, opened 2020-09-09
TorchModel add load inference only model from scala

qiuxin2012:
TorchModel add load inference only model from scala
And change model serialization from CloudPickleSerializer to `Torch.save` and `Torch.load`.


qiuxin2012:
jenkins passed

PR #2853, opened 2020-09-10
Add stop_orca_context to python exit handler

### LINKED ISSUE #2851: standalone cluster is not shut down if program exit unexpectedly
Maybe using context manager to call `stop_orca_context?`
```python
with OrcaContex() as ctx:
    # user program
```

yangw1234:
Fix #2851

### DIFF
--- a/pyzoo/zoo/util/spark.py
+++ b/pyzoo/zoo/util/spark.py
@@ -205,20 +205,16 @@ class SparkRunner:
     def stop_spark_standalone():
         import subprocess
         env = SparkRunner.standalone_env
-        if not env:
-            import pyspark
-            pyspark_home = os.path.abspath(pyspark.__file__ + "/../")
-            zoo_standalone_home = os.path.abspath(__file__ + "/../../share/bin/standalone")
-            pro = subprocess.Popen(["chmod", "-R", "+x", "{}/sbin".format(zoo_standalone_home)])
-            os.waitpid(pro.pid, 0)
-            env = {"SPARK_HOME": pyspark_home,
-                   "ZOO_STANDALONE_HOME": zoo_standalone_home}
-        stop_worker_pro = subprocess.Popen(
-            "{}/sbin/stop-worker.sh".format(env["ZOO_STANDALONE_HOME"]), shell=True, env=env)
-        os.waitpid(stop_worker_pro.pid, 0)
-        stop_master_pro = subprocess.Popen(
-            "{}/sbin/stop-master.sh".format(env["ZOO_STANDALONE_HOME"]), shell=True, env=env)
-        os.waitpid(stop_master_pro.pid, 0)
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/util/spark.py
```diff
@@ -205,20 +205,16 @@ def init_spark_standalone(self,
...
-        os.waitpid(stop_master_pro.pid, 0)
+        if env is not None:
+            stop_worker_pro = subprocess.Popen(
+                "{}/sbin/stop-worker.sh".format(env["ZOO_STANDALONE_HOME"]), shell=True, env=env)
+            os.waitpid(stop_worker_pro.pid, 0)
+            stop_master_pro = subprocess.Popen(
+                "{}/sbin/stop-master.sh".format(env["ZOO_STANDALONE_HOME"]), shell=True, env=env)
+            os.waitpid(stop_master_pro.pid, 0)
+        else:
+            # if env is None, then the standalone cluster is not started by analytics zoo
```
yangw1234, line 216: Is this correct? @hkvision

hkvision: As I test, seems if the cluster is manually started by the user, this script won't stop it, is it correct?
Previously my implementation is that even if the program exited accidentally, users can open a python shell to stop_orca_context manually.
If this is added to exit handler, then I think stop manually is not needed any longer.

pyzoo/zoo/util/spark.py
```diff
@@ -205,20 +205,16 @@ def init_spark_standalone(self,
...
-            os.waitpid(pro.pid, 0)
-            env = {"SPARK_HOME": pyspark_home,
-                   "ZOO_STANDALONE_HOME": zoo_standalone_home}
-        stop_worker_pro = subprocess.Popen(
-            "{}/sbin/stop-worker.sh".format(env["ZOO_STANDALONE_HOME"]), shell=True, env=env)
-        os.waitpid(stop_worker_pro.pid, 0)
-        stop_master_pro = subprocess.Popen(
-            "{}/sbin/stop-master.sh".format(env["ZOO_STANDALONE_HOME"]), shell=True, env=env)
-        os.waitpid(stop_master_pro.pid, 0)
+        if env is not None:
```
hkvision, line 208: just `if env`?

hkvision (approved):
LGTM

pyzoo/zoo/util/spark.py
yangw1234: Though it does not matter in this case, there are some subtle differences between `env` and `env is not None`.
E.g. 
```
env={}
bool(env) # False
env is not None # True
```

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/4197/ jenkin pass

PR #2860, opened 2020-09-11
remove horovod depency if not using horovod backend

yangw1234:


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/4201/ jenkins passed

PR #2862, opened 2020-09-11
Add quick option when building python whl

### LINKED ISSUE #2852: Python only option in make-dist.sh and pyzoo/dev/release.sh
scala/java code is changed much less frequently than python in development. However, we need to compile scala/java code every time we change something even if only in python. This is a relatively long time.

It would be nice to have a python only building options to shorten build time and only compile scala/java code when necessary.

hkvision:
Fix #2852 
quick = true won't build the jar.


hkvision:
Tested manually. @yangw1234 Take a look.

hkvision:
http://10.239.47.210:18888/job/ZOO-PR-Validation/4205/ jenkins passed.
Need to modify release job before the merge.

PR #2864, opened 2020-09-14
add ld_library_path to init spark on yarn

qiuxin2012:
Fix 
```
jep.JepExecption:ImportError: /lib64/libstdc++.so.6: version 'GLIBCXX_3.4.21' not found(required by libtorch_python.so)
```

Add executor env LD_LIBRARY_PATH to use libstdc++.so in conda.

### DIFF
--- a/pyzoo/zoo/util/spark.py
+++ b/pyzoo/zoo/util/spark.py
@@ -102,8 +102,8 @@ class SparkRunner:
                                          extra_executor_memory_for_ray)
             py_version = ".".join(platform.python_version().split(".")[0:2])
             preload_so = executor_python_env + "/lib/libpython" + py_version + "m.so"
-            ld_path = executor_python_env + "/lib:" + executor_python_env + "/lib/python" + \
-                      py_version + "/lib-dynload"
+            ld_path = executor_python_env + "/lib:" + executor_python_env + "/lib/python" +\
+                py_version + "/lib-dynload"
             if "spark.executorEnv.LD_LIBRARY_PATH" in conf:
                 ld_path = "{}:{}".format(ld_path, conf["spark.executorEnv.LD_LIBRARY_PATH"])
             conf.update({"spark.scheduler.minRegisteredResourcesRatio": "1.0",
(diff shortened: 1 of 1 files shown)


pyzoo/zoo/util/spark.py
```diff
@@ -102,8 +102,13 @@ def init_spark_on_yarn(self,
...
             py_version = ".".join(platform.python_version().split(".")[0:2])
             preload_so = executor_python_env + "/lib/libpython" + py_version + "m.so"
+            ld_path = executor_python_env + "/lib:" + executor_python_env + "/lib/python" + \
+                      py_version + "/lib-dynload"
+            if "spark.executorEnv.LD_LIBRARY_PATH" in conf:
+                ld_path = "{}:{}".format(ld_path, conf["spark.executorEnv.LD_LIBRARY_PATH"])
             conf.update({"spark.scheduler.minRegisteredResourcesRatio": "1.0",
                          "spark.executorEnv.PYTHONHOME": executor_python_env,
+                         "spark.executorEnv.LD_LIBRARY_PATH": ld_path,
                          "spark.executorEnv.LD_PRELOAD": preload_so})
```
hkvision, line 112: Add these two in init_spark_on_k8s as well?

yangw1234:
I think `spark.executorEnv.LD_LIBRARY_PATH` is dangerous because it will override spark's own version LD_LIBRARY_PATH, which may contain other user libraries like hbase, etc.

`spark.executor.extraLibraryPath` might be a better choice.

qiuxin2012:
> I think `spark.executorEnv.LD_LIBRARY_PATH` is dangerous because it will override spark's own version LD_LIBRARY_PATH, which may contain other user libraries like hbase, etc.
> 
> `spark.executor.extraLibraryPath` might be a better choice.

I'm using add not override, also matters?

pyzoo/zoo/util/spark.py
qiuxin2012: Sure, @GavinGu07 will add to k8s in another PR.

yangw1234:
> > I think `spark.executorEnv.LD_LIBRARY_PATH` is dangerous because it will override spark's own version LD_LIBRARY_PATH, which may contain other user libraries like hbase, etc.
> > `spark.executor.extraLibraryPath` might be a better choice.
> 
> I'm using add not override, also matters?

I mean in the users' worker machines, there might be an existing "LD_LIBRARY_PATH" environment variable. When you set `spark.executorEnv.LD_LIBRARY_PATH`, the value will just replace the existing "LD_LIBRARY_PATH", and this behavior is coded in spark library which you can not change, right?

On the other hand,  spark will prepend the `spark.executor.extraLibraryPath` value to existing "LD_LIBRARY_PATH", which is safe I think.

PR #2878, opened 2020-09-17
Fix stop_orca_context

hkvision:
Since stop_orca_context is added at exit time, if the user program finishes successfully and they manually call stop_orca_context, stop_orca_context would be called for one more time at exit time. So we shouldn't call `sc.getOrCreate()` all the time as this would create a new sc.
For yarn mode, it would through the following error:
```
py4j.protocol.Py4JJavaError: An error occurred while calling None.org.apache.spark.api.java.JavaSparkContext.
: java.io.FileNotFoundException: File file:/tmp/tmptoj600uh/python_env.tar.gz#python_env does not exist
```
As the tmp conda file is removed when sc is created for the first time.

### DIFF
--- a/pyzoo/zoo/orca/common.py
+++ b/pyzoo/zoo/orca/common.py
@@ -115,6 +115,7 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
 
     :return: An instance of SparkContext.
     """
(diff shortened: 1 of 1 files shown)


hkvision:
jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/4239/

PR #2880, opened 2020-09-17
Support modulepickle for Ray

hkvision:
I tested on yarn it works for two cases:
- https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/examples/orca/learn/horovod/pytorch_estimator.py I put the model definition in a separate python file not available on other nodes.
- Ali tianchi baseline code where the model is defined in a directory of python files not available on other nodes.

The loaded model have the same weights as the original model.
Since I'm still not familiar and confident with pickling, probably need someone to help to confirm this feature. Talk to you next week @jason-dai 

- [ ] Rename model_creator to model.

### DIFF
--- a/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
+++ b/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
@@ -149,7 +149,12 @@ class TorchRunner:
         """Runs the creator functions without any distributed coordination."""
 
         logger.debug("Creating model")
-        self.models = self.model_creator(self.config)
+        # Torch model is also callable, so can't use callable to check.
+        import types
+        if isinstance(self.model_creator, types.FunctionType):
+            self.models = self.model_creator(self.config)
+        else:
+            self.models = self.model_creator
         if not isinstance(self.models, Iterable):
             self.models = [self.models]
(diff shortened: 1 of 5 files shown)


qiuxin2012:
what's the difference between your modulepickle and cloudpickle?

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
```diff
@@ -110,7 +110,12 @@ def setup_components(self):
         """Runs the creator functions without any distributed coordination."""
 
         logger.debug("Creating model")
-        self.models = self.model_creator(self.config)
+        # Torch model is also callable, so can't use callable to check.
+        import types
+        if isinstance(self.model_creator, types.FunctionType):
```
qiuxin2012, line 115: Can we add a ClassType? Some torch model is definded as a class, we can pass the class name too.

hkvision: Why do you need ClassType? If so, what about the arguments to create such a class? Why can't users just create a model instance, which is reasonable?

hkvision:
> what's the difference between your modulepickle and cloudpickle?

It overrides save/load of cloudpickle and replaces the pickler.

PR #2884, opened 2020-09-21
expose tcmf num_workers to zouwu

yushan111:


pyzoo/test/zoo/zouwu/test_model_forecast.py
```diff
@@ -263,6 +263,42 @@ def get_pred(d):
...
+                               alt_iters=2)
+        horizon = np.random.randint(1, 50)
+        # construct data
+        id = np.arange(300)
+        data = np.random.rand(300, 480)
+        input = dict({'id': id, 'y': data})
+
+        from zoo.orca import init_orca_context, stop_orca_context
+
+        init_orca_context(cores=4, spark_log_level="INFO", init_ray_on_spark=True,
```
shane-huang, line 282: If user wants to use it, does he needs to create a spark context? Maybe we need to add a doc about how to use on a cluster (yarn) environment, i.e. where to specify yarn context, etc.

shane-huang:
We need to tell users how to set cluster related parameters in the api docs, e.g. how to set up yarn or spark master, do we still need to call init_nn_context now? or we config using init_orca_context is enough?

pyzoo/test/zoo/zouwu/test_model_forecast.py
yushan111: init_orca_context will create a spark context. User need to specify `cluster_mode="yarn-client"` in `init_orca_context`  to run on yarn according to [doc](https://analytics-zoo.github.io/master/#Orca/context/#initialization-and-termination). UT may not be able to include that much and an example may be better.

yushan111:
jenkins passed: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/4292/

PR #2892, opened 2020-09-22
upgrade xgboost to 1.2

dding3:


dding3:
jekins pass. http://10.239.47.210:18888/job/ZOO-PR-Validation/4275/

PR #2896, opened 2020-09-24
Cluster Serving Mock Baseline

Litchilitchy:
Baseline to run [3, 224, 224] image Tensor.

InferenceBaseline: contains only batch and inference
MockSingle: simulate Cluster Serving old design, single pipeline with Scala ParArray
MockParallel: simulate Cluster Serving new design, multiple pipeline

Tested on local and server, performance almost same.


Litchilitchy:
changed of some parameter pass, so not supporting of Spark is also in this PR

PR #2902, opened 2020-09-25
add pytorch ray_sgd trainer support to Orca Pytorch Estimator

jenniew:
Add pytorch ray_sgd trainer support to Orca Pytorch Estimator.

Because pytorch ray sgd trainer need data_creator in constructor, add data_creator=None to Estimator.from_torch(). No need to give data option to fit and evaluate when use ray sgd trainner.

Add some APIs to PytorchTrainer.

### DIFF
--- a/pyzoo/zoo/orca/learn/pytorch/estimator.py
+++ b/pyzoo/zoo/orca/learn/pytorch/estimator.py
@@ -50,9 +50,10 @@ class Estimator(object):
     def from_torch(*,
                    model,
                    optimizer,
+                   data_creator=None,
                    loss=None,
                    scheduler_creator=None,
-                   training_operator_cls=TrainingOperator,
+                   training_operator_cls=None,
                    initialization_hook=None,
                    config=None,
                    scheduler_step_freq="batch",
@@ -61,6 +62,8 @@ class Estimator(object):
                    model_dir=None,
                    backend="horovod"):
         if backend == "horovod":
+            if training_operator_cls is None:
+                training_operator_cls = TrainingOperator
             return PyTorchHorovodEstimatorWrapper(model_creator=model,
                                                   optimizer_creator=optimizer,
                                                   loss_creator=loss,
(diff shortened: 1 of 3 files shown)


pyzoo/zoo/orca/learn/pytorch/estimator.py
```diff
@@ -61,6 +62,8 @@ def from_torch(*,
                    model_dir=None,
                    backend="horovod"):
         if backend == "horovod":
+            if training_operator_cls is None:
+                training_operator_cls = TrainingOperator
```
hkvision, line 66: Why not ray sgd use the same TrainingOperator?

hkvision:
Shall we align the API for fit? It is awkward if different backends have different APIs.

pyzoo/zoo/orca/learn/pytorch/estimator.py
```diff
@@ -284,3 +301,85 @@ def set_l2_norm_gradient_clipping(self, clip_norm):
...
+                                              training_operator_cls=training_operator_cls,
+                                              initialization_hook=initialization_hook,
+                                              config=config,
+                                              num_workers=None,
+                                              use_fp16=False,
+                                              use_tqdm=use_tqdm,
+                                              scheduler_step_freq=scheduler_step_freq
+                                              )
+
+    def fit(self, epochs=1, profile=False, reduce_results=True, info=None):
```
hkvision, line 334: Is this method override? The arguments are not the same as its parent's.

jason-dai:
> Shall we align the API for fit? It is awkward if different backends have different APIs.

Yes, we should use the same API for different backends as much as possible

pyzoo/zoo/orca/learn/pytorch/estimator.py
```diff
@@ -77,6 +80,20 @@ def from_torch(*,
                                                 optimizer=optimizer,
                                                 model_dir=model_dir,
                                                 bigdl_type="float")
+        elif backend == "ray_sgd":
```
jason-dai, line 83: change backend to "torch_distributed"

hkvision:
> > Shall we align the API for fit? It is awkward if different backends have different APIs.
> 
> Yes, we should use the same API for different backends as much as possible

So will it be better to write our own TorchTrainer instead of importing from RaySGD and align the train method? As we already have TorchRunner, migrating TorchTrainer won't have much effort and we can fix some issues of RaySGD (abort creating a local worker on the driver node so that we don't need to use nested actors).

pyzoo/zoo/orca/learn/pytorch/estimator.py
jenniew: horovod backend use orca pytorch TrainingOperator, but our TrainingOperator constructor is not compatible with ray sgd torch TrainingOperator. So it cannot work in ray sgd torch trainer.

pyzoo/zoo/orca/learn/pytorch/estimator.py
jenniew: For ray sgd pytorch trainer, it needs pass train and val data when creating trainer. so for fit() and evaluate(), it does not need data parameter. To be compatible, we need to remove data parameter in super class.

pyzoo/zoo/orca/learn/pytorch/estimator.py
hkvision: I know. So my opinion is to migrate and modify the TorchTrainer instead of directly using it.

PR #2913, opened 2020-09-27
[WIP] Cluster Serving performance - inference block all others

Litchilitchy:
Example contains basic logic of blocking, `Model.queueing` is variable about how many threads are waiting for the model. 

Model inference should take the highest priority ideally. Because if one thread is waiting but no thread is doing inference, this thread is wasted. And inference should block other non-OMP threads in order to avoid conflict.

Please check if any potential bug or overhead if available.


Litchilitchy:
`sleep 200ms` is to block the thread and simulate inference, `sleep 50ms` is to simulate preprocessing.

Litchilitchy:
Synchronization bring overheads and overall it does not improve. closed.

PR #2920, opened 2020-09-28
Keras fit support x and y both be RDDs

### LINKED ISSUE #2918: Model.fit in keras style python API doesn't support both x and y being RDDs
<https://github.com/intel-analytics/analytics-zoo/blob/35fd8d19754296e6e8a9dcdd689c15e88a2e1dab/pyzoo/zoo/pipeline/api/keras/engine/topology.py#L217-L233>

Here, when calling fit, if both `x` and `y` are RDDs, an `TypeError("Unsupported training data type: %s" % type(x))` would be thrown.

Also, it would be nice if the documentation specifies the format of the single RDD to be passed, i.e., how the labels and features should be specified, in the case where `y=None`.

hkvision:
Fix intel-analytics/analytics-zoo-internal#615

### DIFF
--- a/pyzoo/zoo/pipeline/api/keras/engine/topology.py
+++ b/pyzoo/zoo/pipeline/api/keras/engine/topology.py
@@ -203,8 +203,12 @@ class KerasNet(ZooKerasLayer):
         Train a model for a fixed number of epochs on a DataSet.
 
         # Arguments
-        x: Input data. A Numpy array or RDD of Sample, ImageSet or TextSet.
-        y: Labels. A Numpy array. Default is None if x is already Sample RDD or ImageSet or TextSet.
+        x: Input data. We support multiple data types. x can be a Numpy array or RDD of Sample
+        or RDD of Numpy array or ImageSet or TextSet or FeatureSet.
+        y: Labels. If x is a Numpy array, y must be a Numpy array.
+        If x is RDD of Sample or ImageSet or TextSet or FeatureSet that already contains labels,
+        y should be None, which is the default value.
+        If x is RDD of Numpy array, y should also be RDD of Numpy array.
         batch_size: Number of samples per gradient update. Default is 32.
         nb_epoch: Number of epochs to train.
         validation_data: Tuple (x_val, y_val) where x_val and y_val are both Numpy arrays.
@@ -213,7 +217,6 @@ class KerasNet(ZooKerasLayer):
         distributed: Boolean. Whether to train the model in distributed mode or local mode.
                      Default is True. In local mode, x and y must both be Numpy arrays.
         """
-
         if distributed:
             if isinstance(x, np.ndarray) and isinstance(y, np.ndarray):
                 if validation_data:
@@ -226,8 +229,15 @@ class KerasNet(ZooKerasLayer):
                     x, y = x[:split_index], y[:split_index]
                     validation_data = to_sample_rdd(*validation_data)
                 training_data = to_sample_rdd(x, y)
-            elif (isinstance(x, RDD) or isinstance(x, ImageSet) or isinstance(x, TextSet)) \
-                    or isinstance(x, FeatureSet) and not y:
+            elif isinstance(x, RDD):
+                if not y:  # x is RDD of Sample, y should be None
+                    training_data = x
+                else:  # x and y are both RDD of Numpy arrays
+                    assert isinstance(y, RDD), "x and y must be both RDD of Numpy arrays"
+                    from zoo.common.utils import Sample
+                    training_data = x.zip(y).map(lambda item: Sample.from_ndarray(item[0], item[1]))
+            elif isinstance(x, ImageSet) or isinstance(x, TextSet) or isinstance(x, FeatureSet):
+                assert not y, "y should be None if x is an ImageSet, TextSet or FeatureSet"
                 training_data = x
             else:
                 raise TypeError("Unsupported training data type: %s" % type(x))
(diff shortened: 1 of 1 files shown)


fanzhuyifan:
Should Model.evaluate have the same behavior as Model.fit? i.e., both x and y are allowed to be RDDs.

hkvision:
> Should Model.evaluate have the same behavior as Model.fit? i.e., both x and y are allowed to be RDDs.

Yeah, I noticed and updated :)

hkvision:
> Should Model.evaluate have the same behavior as Model.fit? i.e., both x and y are allowed to be RDDs.

For predict, let's keep the RDD of Sample for now. Since in the view of Python, it is not able to distinguish from RDD of Sample and RDD of Numpy array...

yangw1234:
The rdds to be zipped must "have the **same number of partitions** and the **same number of elements in each partition** (e.g. one was made through a map on the other)".

I am kind of worried that supporting x and y to be both rdds might give users the impression that we support all kinds of rdds.

fanzhuyifan:
> 
> 
> The rdds to be zipped must "have the **same number of partitions** and the **same number of elements in each partition** (e.g. one was made through a map on the other)".
> 
> I am kind of worried that supporting x and y to be both rdds might give users the impression that we support all kinds of rdds.

In that case, do you think the api should automatically repartition while warning the user that performance is harmed, or should the API just not support this case and leave it up to the user?

fanzhuyifan:
Also, I believe that if some method does not use one of the inputs, e.g., if Model.fit does not use the y input, then the reasonable behavior would be checking that y is None and complaining and refusing to run while outputting a clear error message. It seems unreasonable that the user provides an important input that is not used in the method, but the method continues on as if nothing wrong happened.

hkvision:
> Also, I believe that if some method does not use one of the inputs, e.g., if Model.fit does not use the y input, then the reasonable behavior would be checking that y is None and complaining and refusing to run while outputting a clear error message. It seems unreasonable that the user provides an important input that is not used in the method, but the method continues on as if nothing wrong happened.

Yeah, I have made some update about arguments and error reporting in this PR.

hkvision:
> The rdds to be zipped must "have the **same number of partitions** and the **same number of elements in each partition** (e.g. one was made through a map on the other)".
> 
> I am kind of worried that supporting x and y to be both rdds might give users the impression that we support all kinds of rdds.

We primarily want users to use RDD of Sample. In the case where users want to input two separate RDDs, I think it is reasonable to make the assumption that they have the same number of partitions and records? What do you think? @yangw1234

yangw1234:
> > The rdds to be zipped must "have the **same number of partitions** and the **same number of elements in each partition** (e.g. one was made through a map on the other)".
> > I am kind of worried that supporting x and y to be both rdds might give users the impression that we support all kinds of rdds.
> 
> We primarily want users to use RDD of Sample. In the case where users want to input two separate RDDs, I think it is reasonable to make the assumption that they have the same number of partitions and records? What do you think? @yangw1234

Yeah, I think your argument makes sense too. Original I thought if we do not support two rdds, we can check them at the beginning and raise an error telling the user what he/she needs to do (if meets the requirements just zip, if not he/she needs restructure the data pipeline according to his/her data), otherwise the user might encounter errors further down which is hard to find the cause.

Anyway, both ways look fine to me.

pyzoo/zoo/pipeline/api/keras/engine/topology.py
```diff
@@ -203,8 +203,12 @@ def fit(self, x, y=None, batch_size=32, nb_epoch=10,
         Train a model for a fixed number of epochs on a DataSet.
 
         # Arguments
-        x: Input data. A Numpy array or RDD of Sample, ImageSet or TextSet.
-        y: Labels. A Numpy array. Default is None if x is already Sample RDD or ImageSet or TextSet.
+        x: Input data. We support multiple data types. x can be a Numpy array or RDD of
+        zoo.common.utils.Sample or RDD of Numpy array or ImageSet or TextSet or FeatureSet.
+        y: Labels. If x is a Numpy array, y must be a Numpy array.
+        If x is RDD of Sample or ImageSet or TextSet or FeatureSet which already contains labels,
```
qiuxin2012, line 209: FeatureSet is repartitoned, Please make sure the x and y can be zipped correctly.

hkvision:
> > > The rdds to be zipped must "have the **same number of partitions** and the **same number of elements in each partition** (e.g. one was made through a map on the other)".
> > > I am kind of worried that supporting x and y to be both rdds might give users the impression that we support all kinds of rdds.
> > 
> > 
> > We primarily want users to use RDD of Sample. In the case where users want to input two separate RDDs, I think it is reasonable to make the assumption that they have the same number of partitions and records? What do you think? @yangw1234
> 
> Yeah, I think your argument makes sense too. Original I thought if we do not support two rdds, we can check them at the beginning and raise an error telling the user what he/she needs to do (if meets the requirements just zip, if not he/she needs restructure the data pipeline according to his/her data), otherwise the user might encounter errors further down which is hard to find the cause.
> 
> Anyway, both ways look fine to me.

I checked the zip code, if the two rdds don't have the same number of partitions, an error would be directly raised in Python. If the number of elements are not the same, the error would be raised on JVM. So I think it is OK to just call zip by us and leave to Spark to handle the error for zip? We can write in the doc to remind users.
For different partitions, the error would be directly raised and it is clear; users should be aware of the error easily. For different number of records, I think this rarely happens if users handle their data correctly. @yangw1234 @qiuxin2012

PR #2974, opened 2020-10-21
Add big TF model support

jenniew:
Add cloneParameter.
Add TFModelBroadCast
clear extraParameter when broadcast model.
work around getExtraParameters collect.
Test train, evaluate and predict with big NCF model


jason-dai:
LGTM. You may open an issue to clean some duplication of util methods because they are private in bigdl.

jenniew:
I'll create another PR to fix duplicated util methods with reflection.

jason-dai:
> I'll create another PR to fix duplicated util methods with reflection.

Instead of reflection, make these method in BigDL public.

jenniew:
Jenkins Pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/4403/

jenniew:
> > I'll create another PR to fix duplicated util methods with reflection.
> 
> Instead of reflection, make these method in BigDL public.

Ok

PR #2975, opened 2020-10-21
fix pickle

jenniew:
Fix pickle using python pickle in parallel.


jenniew:
Tested on save/load big pickle files.

PR #2977, opened 2020-10-21
Add cluster node ip to no_proxy env for tf2 backend

yangw1234:
TF distributed will try to sent data through proxy, this may result an error.

Add cluster nodes to no_proxy env.


yangw1234:
jenkins passed http://172.168.2.101:8080/job/ZOO-PR-Python-Spark-2.4-py37-Horovod/264/

PR #3003, opened 2020-10-28
Orca PyTorch Estimator input torch loss instance

### LINKED ISSUE #2998: Orca Pytorch Estimator input loss instance instead of loss class?
https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/pytorch/torch_runner.py#L94
Currently when creating the Estimator, we can input loss to be a class of Torch Loss, and internally we would call `loss_creator()`. This won't work if users want to specify arguments when creating the loss, for example, size_average.
Better to input the loss instance instead of class directly?

hkvision:
Fix #2998 
@yangw1234 Take a look.
Also torch model and torch loss are also callable, can't use callable to identify the input is a function.


hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/4431/

hkvision:
Merge this first. Comment if there are further issues.

PR #3008, opened 2020-10-28
Disable predict on TFDataDataset to prevent returning incorrect result silently

yangw1234:


jason-dai:
Need to fix orca estimator.predict too

yangw1234:
jenkins passed http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/4446/

PR #3009, opened 2020-10-29
fix pytorch base model and rename RayTuneSearchEngine

yushan111:


yushan111:
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/4447/

PR #3012, opened 2020-10-29
Fix validate in Orca PyTorch Estimator

hkvision:
- If target is of size (batch, 1) and predicted of size (batch, ), the accuracy calculation `target == predicted` would result in a tensor of size (batch, batch), which is wrong.
- Also accuracy for binary classification is not supported previously.
- Handle cases where output or target has extra singleton dimensions.
- Add unit tests to cover this.


yangw1234 (approved):

hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/4473/

PR #3016, opened 2020-10-29
Add options for orca read_parquet

jenniew:
Add schema and options for Orca read_parquet


jenniew:
Jenkins pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/4450/

Issue #2946, opened 2020-10-12
ValueError: Last date time is bigger than current time!


MusaBilal4:
After using `TimeSequencePredictor` from `zoo.automl.regression.time_sequence_predictor` to train a model, if evaluating it on a test set which has timestamps greater than the current date, it throws an error `ValueError: Last date time is bigger than current time!`. The functions being used are `pipeline.evaluate` and `pipeline.predict`.
Is there a way to avoid this and use timestamps beyond current date in evaluation?

yushan111:
We assume that it may be impossible to get target data later than current time. 
If that could happen, sure, I will change the error into warning.

Thanks for the issue.

yushan111:
The main reason we didn't allow future datetime is that we use featuretools to generate time series feature. And it seems it failed to generate features for future time. We will look up the reason and work out a better solution.

yushan111:
@TheaperDeng

TheaperDeng:
The problem might be caused by a missing param `cutoff_time` when using `featuretools`. I will fix it soon.

helenlly:
@yushan111  since the PR is merged, any suggestion fir next step?

yushan111:
@MusaBilal4 Sorry for the late response. We have fixed it in analytics-zoo 0.9.0. You could have a try. Thank you for the issue!

PR #3021, opened 2020-10-31
automl allow future time in feature

TheaperDeng:
related to issue #2946

### DIFF
--- a/pyzoo/zoo/automl/feature/time_sequence.py
+++ b/pyzoo/zoo/automl/feature/time_sequence.py
@@ -393,10 +393,11 @@ class TimeSequenceFeatureTransformer(BaseFeatureTransformer):
 
         # check if the last datetime is large than current time.
         # In that case, feature tools generate NaN.
-        last_datetime = dt.iloc[-1]
-        current_time = np.datetime64('today', 's')
-        if last_datetime > current_time:
-            raise ValueError("Last date time is bigger than current time!")
+        # This test is ignored now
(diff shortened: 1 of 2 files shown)


pyzoo/zoo/automl/feature/time_sequence.py
```diff
@@ -541,6 +534,8 @@ def _generate_features(self, input_df):
         df = input_df.copy()
         df["id"] = df.index + 1
 
+        cutoff_time = np.datetime64(np.max(df[self.dt_col]))
```
yushan111, line 537: df[self.dt_col].max().to_numpy() or just use np.inf?

yushan111 (commented):
Others LGTM.

pyzoo/zoo/automl/feature/time_sequence.py
TheaperDeng: np.inf doesn't work. I will try the other one.

TheaperDeng:
Here is the related code in featuretools.
https://github.com/FeatureLabs/featuretools/blob/main/featuretools/computational_backends/calculate_feature_matrix.py#L172-L176
```python
        if cutoff_time is None:
            if entityset.time_type == NumericTimeIndex:
                cutoff_time = np.inf
            else:
                cutoff_time = datetime.now()
```

TheaperDeng:
PR validation pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/4463/

PR #3022, opened 2020-11-02
update Torch example

qiuxin2012:


pyzoo/zoo/examples/pytorch/train/mnist/main.py
```diff
@@ -89,7 +90,7 @@ def main():
         num_executors = 2
         num_cores_per_executor = 4
         hadoop_conf_dir = os.environ.get('HADOOP_CONF_DIR')
-        zoo_conda_name = os.environ.get('ZOO_CONDA_NAME')  # The name of the created conda-env
+        zoo_conda_name = detect_python_location().split("/")[-3]  # The name of the created conda-env
```
jason-dai, line 93: Why cannot the user specify the coda name manually? If so, we should provide a utility function to return the conda name directly; maybe we can also use it as in init_spark_on_yarn?

qiuxin2012: We can detect the conda env name automatically, then user don't need to set the env `ZOO_CONDA_NAME`.
I just add a new function `detect_conda_env_name`.

pyzoo/zoo/examples/pytorch/train/mnist/main.py
```diff
@@ -89,7 +90,7 @@ def main():
         num_executors = 2
         num_cores_per_executor = 4
         hadoop_conf_dir = os.environ.get('HADOOP_CONF_DIR')
-        zoo_conda_name = os.environ.get('ZOO_CONDA_NAME')  # The name of the created conda-env
+        zoo_conda_name = detect_conda_env_name()  # auto detect current conda env name
```
jason-dai, line 93: what if the user does not create a conda environment?

qiuxin2012: will throw an exception if conda is unavailable

pyzoo/zoo/orca/common.py
```diff
@@ -143,17 +143,16 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
                 "Directory path to hadoop conf not found for yarn-client mode. Please either " \
                 "specify argument hadoop_conf or set the environment variable HADOOP_CONF_DIR"
             hadoop_conf = kwargs["hadoop_conf"]
-        from zoo.util.utils import detect_python_location
-        python_location = detect_python_location()  # /path/to/conda/envs/conda_name/bin/python
-        assert "envs" in python_location, "You must use a conda environment for yarn-client mode"
+        from zoo.util.utils import detect_conda_env_name
+        conda_env_name = detect_conda_env_name()
```
qiuxin2012, line 147: @cyita I change this to detect_conda_env_name who use `conda info` to find the conda env name, seems better than current method.

qiuxin2012: @hkvision

qiuxin2012:
jenkins passed

PR #3028, opened 2020-11-04
Orca Pytorch Cifar10 Example

jing-xu:
cifar10.py is using backend bigdl and cifar10_pytorch.py is using backend pytorch


jing-xu:
Running with default settings: 10g memory, 4 cores with only 1 worker per core.

Training result:
`Train stats: [{'num_samples': 50000, 'epoch': 1, 'batch_count': 12500, 'train_loss': 1.6994677528655528, 'last_train_loss': 2.9595041275024414}, {'num_samples': 50000, 'epoch': 2, 'batch_count': 12500, 'train_loss': 1.3413889428651333, 'last_train_loss': 2.5499534606933594}]`

Compare to tutorial example: 
`Training loss: 1.329`

Validation result:
`validation stats: {'num_samples': 10000, 'batch_count': 2500, 'val_loss': 1.2570434251338243, 'last_val_loss': 1.322148084640503, 'val_accuracy': 0.5565, 'last_val_accuracy': 0.5}`

Compare to tutorial example: 
`Validation accuracy: 54%`

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,125 @@
...
+        init_orca_context(cores=1, memory="20g")
+    elif args.cluster_mode == "yarn":
+        init_orca_context(
+            cluster_mode="yarn-client", cores=4, num_nodes=2, memory="2g",
+            driver_memory="10g", driver_cores=1,
+            conf={"spark.rpc.message.maxSize": "1024",
+                  "spark.task.maxFailures": "1",
+                  "spark.driver.extraJavaOptions": "-Dbigdl.failure.retryTimes=1"})
+    '''
+    init_orca_context(cluster_mode=args.cluster_mode, cores=args.cores, num_nodes=args.num_nodes, memory=args.memory,  env={"http_proxy": "http://10.239.4.100:913", "https_proxy": "https://10.239.4.100:913"} )
```
yangw1234, line 111: the proxy settings should not be included here

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,125 @@
...
+
+    #model.train()
```
hkvision, line 113: remove the comments

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,125 @@
...
+    criterion = nn.CrossEntropyLoss()
+    zoo_estimator = Estimator.from_torch(model=model_creator, optimizer=optimizer_creator, loss=criterion, config={
+        "batch_size":args.total_batch}, backend="pytorch")
```
hkvision, line 116: no total_batch in args

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
jing-xu: solved

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
jing-xu: solved

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+
+import torch.nn as nn
+import torch.nn.functional as F
+
+import torch.optim as optim
+
+from zoo.orca import init_orca_context, stop_orca_context
+from zoo.orca.learn.pytorch import Estimator
+from zoo.orca.learn.metrics import Accuracy
+from zoo.orca.learn.trigger import EveryEpoch
```
hkvision, line 19: These are not needed for pytorch backend

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+os.environ['KMP_DUPLICATE_LIB_OK']='True'
```
hkvision, line 22: Set this to env in init_orca_context?

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+    parser.add_argument("--workers_per_node", type=int, default=1,
+                        help="The number of workers to run on each node")
+    parser.add_argument('--epochs', type=int, default=2, metavar='N',
+                        help='number of epochs to train (default: 2)')
+    parser.add_argument('--worker_batch', type=int, default=16, metavar='N',
+                        help='input worker batch for training per executor(default: 16)')
+    args = parser.parse_args()
+
+
+    classes = ('plane', 'car', 'bird', 'cat',
```
hkvision, line 97: Is it used? If not, add to comments?

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+    return optimizer
+
+def train_data_creator(config):
+    transform = transforms.Compose(
+        [transforms.ToTensor(),
+         transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))])
+
+    trainset = torchvision.datasets.CIFAR10(root='./data', train=True,
+                                            download=True, transform=transform)
+    trainloader = torch.utils.data.DataLoader(trainset, batch_size=config.get("batch_size",32),
```
hkvision, line 58: space before , : config.get("batch_size", 32)

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+        "batch_size":args.worker_batch}, backend="pytorch")
+    stats = zoo_estimator.fit(train_data_creator, epochs=args.epochs)
+    print("Train stats: {}".format(stats))
+    val_stats = zoo_estimator.evaluate(validation_data_creator)
+    print("validation stats: {}".format(val_stats))
+    stop_orca_context()
+
+
+if __name__ == '__main__':
+    main()
```
hkvision, line 113: no new line at the end of the file would cause style check to fail

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+    classes = ('plane', 'car', 'bird', 'cat',
+               'deer', 'dog', 'frog', 'horse', 'ship', 'truck')
+
+    init_orca_context(cluster_mode=args.cluster_mode, cores=args.cores, num_nodes=args.num_nodes, memory=args.memory)
+
+    criterion = nn.CrossEntropyLoss()
+    zoo_estimator = Estimator.from_torch(model=model_creator, optimizer=optimizer_creator, loss=criterion, config={
+        "batch_size":args.worker_batch}, backend="pytorch")
```
hkvision, line 104: space after :

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+    stats = zoo_estimator.fit(train_data_creator, epochs=args.epochs)
```
hkvision, line 105: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/pytorch/estimator.py#L113
I have changed the implementation and set batch_size here in fit instead of in config :)

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
...
+    print("Train stats: {}".format(stats))
+    val_stats = zoo_estimator.evaluate(validation_data_creator)
+    print("validation stats: {}".format(val_stats))
```
hkvision, line 108: Add `estimator.shutdown()` after finishes.

hkvision:
What's the result of directly running the tutorial without using Orca?

hkvision:
Add a simple readme, similar to here: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/examples/orca/learn/mxnet/README.md

pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10_pytorch.py
```diff
@@ -0,0 +1,113 @@
+from __future__ import print_function
```
hkvision, line 1: Add a license header: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/examples/orca/learn/mxnet/lenet_mnist.py#L1-L15

jing-xu:
> What's the result of directly running the tutorial without using Orca?
> Is the difference in accuracy and loss due to using larger batch_size?

The results running the tutorial codes directly with batch size 16 is
`Training stats: {'epoch': 1, 'training_accuracy': 2.007647456893921'}, {'epoch': 2, 'training_accuracy': 1.5634780088043212}
Validation stats: {'val_accuracy': 0.46}`

Those results are similar compared to what we have with Orca

jing-xu:
Compared to Cifar10 on Pytorch tutorial, they share the same model and same input data, and get similar training and validation accuracy. The only thing I changed is to add an estimator.

jason-dai:
Several comments:
1)  Need to give credit to the original code; see https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/examples/ray/parameter_server/model.py for an example
2) Why do we need two versions? I think we just need one version, and use a argument to control the backend if needed.
3) The code does not have the same logic as the original code at https://github.com/pytorch/tutorials/blob/master/beginner_source/blitz/cifar10_tutorial.py; in particular, it does not have the logic for testing the Network on test data.

jing-xu:
> Several comments:
> 
> 1. Need to give credit to the original code; see https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/examples/ray/parameter_server/model.py for an example
> 2. Why do we need two versions? I think we just need one version, and use a argument to control the backend if needed.
> 3. The code does not have the same logic as the original code at https://github.com/pytorch/tutorials/blob/master/beginner_source/blitz/cifar10_tutorial.py; in particular, it does not have the logic for testing the Network on test data.

@jason-dai Would you mind giving more explanation on the third point? Currently the only difference compared to the original code is the calculation of accuracy per class, which is not supported by current Orca Pytorch estimator. We need to modify the implementation of the estimator in order to calculate that. Are there any other differences I should pay attention to?

jason-dai:
> > Several comments:
> > 
> > 1. Need to give credit to the original code; see https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/examples/ray/parameter_server/model.py for an example
> > 2. Why do we need two versions? I think we just need one version, and use a argument to control the backend if needed.
> > 3. The code does not have the same logic as the original code at https://github.com/pytorch/tutorials/blob/master/beginner_source/blitz/cifar10_tutorial.py; in particular, it does not have the logic for testing the Network on test data.
> 
> @jason-dai Would you mind giving more explanation on the third point? Currently the only difference compared to the original code is the calculation of accuracy per class, which is not supported by current Orca Pytorch estimator. We need to modify the implementation of the estimator in order to calculate that. Are there any other differences I should pay attention to?

In the original code, no validation/test data are used in the training process; instead, after the training, the trained model is used to compute the accuracy of the test data.

If there is any missing feature, you may just submit an issue in github.

PR #3032, opened 2020-11-04
XShards save/load numpy

jenniew:
Support Xshards save/load  to/from numpy distributedly.


pyzoo/zoo/orca/data/shard.py
```diff
@@ -441,6 +476,28 @@ def save_pickle(self, path, batchSize=10):
...
+                makedirs(path)
+
+            def save(path):
+                def save_func(index, iterator):
+                    data = list(iterator)
+                    assert len(data) <= 1
+
+                    if len(data) == 1:
+                        print("data 0 is :", data[0])
+                        save_numpy(os.path.join(path, str(index) + ".npy"), data[0])
```
hkvision, line 489: Better not use duplicate names for save_numpy to avoid possible overrides.

jenniew: updated

PR #3104, opened 2020-11-19
Orca Pytorch Estimator: Tensorboard in pytorch_ray_estimator

jing-xu:
Combined tensorboard into pytorch_ray_estimator and auto-remove previous histories.

### DIFF
--- a/pyzoo/zoo/orca/learn/pytorch/pytorch_ray_estimator.py
+++ b/pyzoo/zoo/orca/learn/pytorch/pytorch_ray_estimator.py
@@ -15,6 +15,8 @@
 #
 
 import io
+import os
+import shutil
 import types
 import logging
 import numbers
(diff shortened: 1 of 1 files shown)


pyzoo/zoo/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -214,6 +218,11 @@ def transform_func(worker, shards_ref):
                                                        info=info)
 
         epoch_stats = list(map(list, zip(*worker_stats)))
+        if os.path.exists("./runs"):
+            shutil.rmtree("./runs")
+        writer = SummaryWriter('runs/Training_loss')
+        for stat in epoch_stats:
+            writer.add_scalar("training_loss", stat[0]['train_loss'], stat[0]['epoch'])
```
jing-xu, line 225: "runs" is a folder that record logs for tensorboard to draw. Check whether there exists such a folder first. If so, delete it when runing new task. Then I add a writer that could add train_loss vs epoch info to log. In this way, when running "tensorboard --logdir=runs", it will plot the graph on localhost:6006

Issue #3081, opened 2020-11-14
Training with orca.tf.estimator converges at a higher loss when compared to using TensorFlow/PyTorch with the same settings and data.


ottermegazord:
Using the same architecture, data and hyperparameters, I am able to replicate the performance of my TensorFlow model in PyTorch within a very small degree of variance. However the same implementation on Analytics Zoo yields a significantly higher loss function, even if the models are trained with the same inputs as the models built using TensorFlow and PyTorch. Is this expected?

Model Architecture
```
def model_zoo():
    """Autoencoder"""
    input_1 = tf.keras.Input(shape=(2016,), name='input_1')
    encoder_1 = tf.keras.layers.Dense(1024, name='encoder_1', activation='relu')(input_1)
    encoder_2 = tf.keras.layers.Dense(256, name='encoder_2', activation='relu')(encoder_1)
    encoder_3 = tf.keras.layers.Dense(128, name='encoder_3', activation='relu')(encoder_2)

    decoder_1 = tf.keras.layers.Dense(128, name='decoder_1', activation='relu')(encoder_3)
    decoder_2 = tf.keras.layers.Dense(256, name='decoder_2', activation='relu')(decoder_1)
    decoder_3 = tf.keras.layers.Dense(1024, name='decoder_3', activation='relu')(decoder_2)
    decoder_4 = tf.keras.layers.Dense(2016, name='decoder_4', activation='relu')(decoder_3)

    """DNN 3"""
    dense_3_1 = tf.keras.layers.Dense(128, name='dense_3_1', activation='relu')(encoder_3)
    dense_3_2 = tf.keras.layers.Dense(64, name='dense_3_2', activation='relu')(dense_3_1)
    dense_3_3 = tf.keras.layers.Dense(32, name='dense_3_3', activation='relu')(dense_3_2)
    dense_3_3 = tf.keras.layers.BatchNormalization()(dense_3_3)
    dense_3_3 = tf.keras.layers.Dropout(0.2)(dense_3_3)

    output = tf.keras.layers.Dense(1, name='output', activation='sigmoid')(dense_3_3)

    model = tf.keras.Model(inputs=[input_1],
                           outputs=[decoder_4, output])

    num_steps = 35

    learning_rate_fn = tf.keras.optimizers.schedules.PiecewiseConstantDecay(
        [50 * num_steps, 100 * num_steps, 30 * num_steps, 20 * num_steps, 50 * num_steps],
[...]

PR #3107, opened 2020-11-19
Fix converge issue in Orca TF Estimator

jenniew:
Fix converge issue in Orca TF Estimator.
Related issue: https://github.com/intel-analytics/analytics-zoo/issues/3081


pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -628,6 +631,8 @@ def fit(self, data,
                              hard_code_batch_size=hard_code_batch_size,
                              sequential_order=False, shuffle=True,
                              auto_shard_files=auto_shard_files)
+        if isinstance(dataset, TFNdarrayDataset):
+            dataset = _standarize_feature_label_dataset(dataset, self.model.model)
```
jason-dai, line 635: It seems that we are doing this for `KerasModel`, but not `TFOptimizer`? Why don't we handle it in `TFNdarrayDataset`? @yangw1234

yangw1234: This is a special treatment for Keras models because Keras requires a target to have a shape of [batch, 1] not [batch,]. This is not generally the case for other models.

jason-dai: > This is a special treatment for Keras models because Keras requires a target to have a shape of [batch, 1] not [batch,]. This is not generally the case for other models.

It seems that the processing is not very consistent.
1) It's not handled properly in `KerasModel.evaluate` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L209) or `KerasModel.predict` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L275)
2) It's not handled in `TFOptimizer.from_keras` at all

I think we need to handle it properly in TFPark (both `KerasModel` and `TFOptimizer`), if not in `TFNdarrayDataset`

yangw1234: Yeah, we should handle it properly in both `KerasModel` and `TFOptimizer`!

yangw1234: @jason-dai While we are on this subject, `TFOptimizer.from_keras` is essentially trying to reimplement tf.keras.fit in analytics zoo. 

I think it is very hard, if not impossible, to match the exact behaviors of keras in analytics zoo. There are more than 8k lines of code in tf.keras just dealing with the training logic and more than 5k lines of code of tests covering it. In contrast, tf_optimizer.py only has 761 lines of code. I think there must be some logic we are missing.

So I am thinking, should we support all keras models (both tf 1.15 and 2.x) in ray-based estimators? And use bigdl based estimator for graph API only. It is much easier to just call keras instead of reimplementing it.  And other features like callbacks are supported out of the box.

jason-dai: > @jason-dai While we are on this subject, `TFOptimizer.from_keras` is essentially trying to reimplement tf.keras.fit in analytics zoo.
> 
> I think it is very hard, if not impossible, to match the exact behaviors of keras in analytics zoo. There are more than 8k lines of code in tf.keras just dealing with the training logic and more than 5k lines of code of tests covering it. In contrast, tf_optimizer.py only has 761 lines of code. I think there must be some logic we are missing.
> 
> So I am thinking, should we support all keras models (both tf 1.15 and 2.x) in ray-based estimators? And use bigdl based estimator for graph API only. It is much easier to just call keras instead of reimplementing it. And other features like callbacks are supported out of the box.

That is orthogonal to this PR and we can handle it in a separate issue.

Looks like we are using `KerasModel` in both `TFKerasWrapper.predict` and `TFKerasWrapper.evaluate`, I wonder why we are not using `KerasModel` in `TFKerasWrapper.fit`? if so, we can just make sure `KerasModel` has the correct behavior.

yangw1234:
I still cannot figure out why this fix works. 
Is it possible to write a unit test for it?

jason-dai:
> I still cannot figure out why this fix works.
> Is it possible to write a unit test for it?

That's exactly what you did in `KerasModel.fit` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L142), yes? BTW, why didn't we use `KerasModel.fit` in `TFKerasWrapper.fit`?

yangw1234:
> > I still cannot figure out why this fix works.
> > Is it possible to write a unit test for it?
> 
> That's exactly what you did in `KerasModel.fit` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L142), yes? BTW, why didn't we use `KerasModel.fit` in `TFKerasWrapper.fit`?

I mean the standardize_... func just add one dimension to label‘s shape, it should not affect convergence. 

I am not sure why Model.fit is not used here. Maybe to support checkpoint I guess?

jenniew:
> > > I still cannot figure out why this fix works.
> > > Is it possible to write a unit test for it?
> > 
> > 
> > That's exactly what you did in `KerasModel.fit` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L142), yes? BTW, why didn't we use `KerasModel.fit` in `TFKerasWrapper.fit`?
> 
> I mean the standardize_... func just add one dimension to label‘s shape, it should not affect convergence.
> 
> I am not sure why Model.fit is not used here. Maybe to support checkpoint I guess?

Yes, KerasModel does not support load bigdl checkpoint up to now. So I use low-level TFOptimizer to support that. If we add support to load bigdl checkpoint, then we can use KerasModel.fit directly.

jason-dai:
> > > > I still cannot figure out why this fix works.
> > > > Is it possible to write a unit test for it?
> > > 
> > > 
> > > That's exactly what you did in `KerasModel.fit` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L142), yes? BTW, why didn't we use `KerasModel.fit` in `TFKerasWrapper.fit`?
> > 
> > 
> > I mean the standardize_... func just add one dimension to label‘s shape, it should not affect convergence.
> > I am not sure why Model.fit is not used here. Maybe to support checkpoint I guess?
> 
> Yes, KerasModel does not support load bigdl checkpoint up to now. So I use low-level TFOptimizer to support that. If we add support to load bigdl checkpoint in KerasModel, then we can use KerasModel.fit directly.

OK - I think we can merge this PR for now (maybe also fixing [`KerasModel.evaluate`](https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L209) an [`KerasModel.predict`](https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L275)); we can add checkpoint support in `KerasModel` later.

jenniew:
jenkins pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/4573/

jason-dai:
@jenniew you may also want to fix `KerasModel.evaluate` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L209) and `KerasModel.predict` (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/tfpark/model.py#L275)

PR #3112, opened 2020-11-20
cleanup jar-with-dependency.jar 

qiuxin2012:
When set `-P spark_2.4+`, our build will include some useless jars.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -195,9 +195,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
-                <executions>
-
-                </executions>
+                <version>2.4</version>
             </plugin>
 	    <plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -195,9 +195,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
-                <executions>
-
-                </executions>
+                <version>2.4</version>
```
qiuxin2012, line 198: The latest `maven-assembly-plugin seems not compatible with maven-shade-plugin, the all-in-one package `jar-with-dependencies.jar` doesn't pack zoo-core, change to the 2.4 version will fix this error.

Litchilitchy:
tested on Cluster Serving, works well

qiuxin2012:
jenkins passed

PR #3125, opened 2020-11-24
add pytorch app tests

GavinGu07:


apps/run-app-tests-pytorch-pip.sh
```diff
@@ -0,0 +1,65 @@
...
+    pip uninstall -y analytics-zoo
+    pip uninstall -y bigdl
+    pip uninstall -y pyspark
+}
+
+chmod +x ${ANALYTICS_ZOO_HOME}/apps/ipynb2py.sh
+
+set -e
+
+RUN_PART1=0
```
qiuxin2012, line 16: we don't need so many parts here.

GavinGu07: Get. I will update it.

apps/run-app-tests-pytorch-pip.sh
```diff
@@ -0,0 +1,32 @@
...
+    pip uninstall -y analytics-zoo
+    pip uninstall -y bigdl
+    pip uninstall -y pyspark
+}
+
+chmod +x ${ANALYTICS_ZOO_HOME}/apps/ipynb2py.sh
+
+set -e
+
+echo "#15 start app test for pytorch face-generation"
```
qiuxin2012, line 16: `#1`

GavinGu07: I have deleted the number since there is only one test.

PR #3135, opened 2020-11-27
orca.data.image initial design

yangw1234:


jason-dai:
Maybe the user code can simply be
```
orca.data.image.write_parquet(format=..., ...)

orca.data.image.read_parquet(...)
```

pyzoo/zoo/orca/data/image/parquet_dataset.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class ParquetDataset:
+
+    @staticmethod
+    def write(path, generator, schema, block_size=1000, write_mode="overwrite", **kwargs):
+        """
+        Take each record in the generator and write it to a parquet file.
+        
+        **generator**
```
jason-dai, line 20: is `generator` a method or a data object (e.g., a dictionary)?

pyzoo/zoo/orca/data/image/parquet_dataset.py
```diff
@@ -0,0 +1,111 @@
...
+        :param kwargs: other args
+        """
+
+        sc = SparkContext.getOrCreate()
+        spark = SparkSession(sc)
+        node_num, core_num = get_node_and_core_number()
+        for i, chunk in enumerate(chunks(generator, block_size)):
+            chunk_path = os.path.join(path, f"chunk={i}")
+            rows_rdd = sc.parallelize(chunk, core_num * node_num).map(lambda x: dict_to_row(schema, x))
+            spark.createDataFrame(rows_rdd).write.mode(write_mode).parquet(chunk_path)
```
jason-dai, line 69: why are we writing multiple parquet files?

pyzoo/zoo/orca/data/image/parquet_dataset.py
yangw1234: a python generator object (produced by yield, like an iterator), each element in the generator is a dictionary.
e.g
```
def make_generator(num):
    for i in range(num):
         yield {"id": i}

generator = make_generator(num)

```

pyzoo/zoo/orca/data/image/parquet_dataset.py
yangw1234: sc.parallelize will walk all the elements in the generator and it may not fit the memory.

jason-dai: Is `block_size=1000` reasonable for a single parquet file size?

yangw1234: I just put a number here first and I think I need to do some investigations to determine the final default number.

pyzoo/zoo/orca/data/image/parquet_dataset.py
```diff
@@ -0,0 +1,122 @@
...
+            f.write(json.dumps(schema, cls=EnumEncoder))
+
+
+    @staticmethod
+    def _read_as_dict_rdd(path):
+        sc = SparkContext.getOrCreate()
+        spark = SparkSession(sc)
+        df = spark.read.parquet(path)
+        schema_path = os.path.join(path, "_orca_metadata")
+        with open(schema_path, "r") as fp:
```
jason-dai, line 82: this only works for the local file, no?

pyzoo/zoo/orca/data/image/utils.py
```diff
@@ -0,0 +1,120 @@
...
+class DType(Enum):
+    INT32 = 1
+    FLOAT32 = 2
+    STRING = 3
+
+
+class FeatureType(Enum):
+    IMAGE = 1
+    NDARRAY = 2
+    SCALAR = 3
```
jason-dai, line 20: Maybe we can just combine DType and FeatureType?

pyzoo/zoo/orca/data/image/parquet_dataset.py
```diff
@@ -126,22 +119,51 @@ def read_as_torch(path):
...
-        pass
-
-    @staticmethod
-    def write_mnist(images, labels, output_path):
-        pass
-
-    @staticmethod
-    def write_fashion_mnist(images, labels, output_path):
-        pass
+def write_from_image_directory(directory, label_map, output_path,
```
jason-dai, line 122: It is not clear what this method is; I think it makes more sense to use `write_imagenet`, `write_coco`, etc., explicitly.

pyzoo/zoo/orca/data/image/parquet_dataset.py
```diff
@@ -0,0 +1,169 @@
...
+        3. Image, which takes a string representing a image file in local file system and save the raw file content bytes.
+           The corresponding parquet type is BYTE_ARRAY.
+        
+        :param path: the output path, e.g. file:///output/path, hdfs:///output/path
+        :param generator: generate a dict, whose key is a string and value is one of (a scalar value, ndarray, image file path)
+        :param schema: a dict, whose key is a string, value is one of (schema_field.Scalar, schema_field.NDarray, schema_field.Image)
+        :param kwargs: other args
+        """
+
+        sc = SparkContext.getOrCreate()
```
jason-dai, line 57: init_nn_context

pyzoo/zoo/orca/data/image/parquet_dataset.py
yangw1234: It is very common to have a dataset of the following format and this method is to support it.

Keras (https://keras.io/api/preprocessing/image/#imagedatasetfromdirectory-function), tfds (https://www.tensorflow.org/datasets/api_docs/python/tfds/folder_dataset/ImageFolder), pytorch (https://pytorch.org/docs/stable/torchvision/datasets.html#datasetfolder) all have a similar function.

```
path/to/image_dir/
    label1/  # Ex: 'airplane' or '0015'
      xxx.png
      xxy.png
      xxz.png
    label2/
      xxx.png
      xxy.png
      xxz.png
```

jason-dai: As the initial implementation, I think we should first focus on an end-to-end minimum example (maybe mnist or fashion-mnist).

yangw1234: ok

pyzoo/zoo/orca/data/image/parquet_dataset.py
```diff
@@ -0,0 +1,139 @@
...
+                              shape=labels[1:])
+    }
+
+    def make_generator():
+        for i in range(images.shape[0]):
+            yield {"image": images[i], "labels": labels[i]}
+
+    ParquetDataset.write(output_path, make_generator(), schema, **kwargs)
+
+def write_mnist(images, labels, output_path, **kwargs):
```
jason-dai, line 138: I think the input should be a folder or files?

yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/4742/ jenkins passed

yangw1234:
I'll first merge it to let the inspur folks start to work. Further comments will be addressed in follow up prs.

PR #3140, opened 2020-11-27
TFNet load saved model support boolean

cyita:


yangw1234:
LGTM,

It would be nice to add a unit test. And also maybe we can add other unsupported data types if it is not too much trouble.

cyita:
Jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/4611/

PR #3158, opened 2020-12-02
fix jep stuck in scala 2.12

qiuxin2012:
fix jep stuck in scala 2.12, and add a 100s timeout in jep.


qiuxin2012:
jenkins passed

PR #3159, opened 2020-12-02
Python Estimator eval return value bugfix

cyita:

PR #3160, opened 2020-12-02
change retrun type of estimator's evaluate

### LINKED ISSUE #3156: Pytorch Spark estimator evaluate returns JavaObject
The result of evaluate from Scala is JavaObject, can't be accessed or processed in Python, can only print the result. The return result in PythonEstimator is wrong.

qiuxin2012:
fixed #3156


qiuxin2012:
duplicated #3159

Issue #3184, opened 2020-12-04
PyTorch Ray Estimator seems not syncing weights for different workers


hkvision:
If I intentionally add `time.sleep` in `data_creator` to let different workers randomly sleep for different amount of time, the workers don't start the training together (i.e. the worker sleeps for less time starts first without waiting for other workers to sync weights).
Seems each worker is training separately and they are not syncing weights.
Also if I intentionally make each worker have different data size, the worker with larger data will continue training until the training finishes.
This behavior occurs for both torch and horovod backends.

hkvision:
Horovod result:
```
(pid=17328) Sleeping for 5 seconds
(pid=17329) Sleeping for 1 seconds
(pid=17329) Number of files on this worker:  8
(pid=17329) Number of files to be trained:  6
  0%|          | 0/1172 [00:00<?, ?batch/s]
(pid=17329) Data size on worker:  75000
  0%|          | 1/1172 [00:00<00:47, 24.62batch/s, loss=0.735]
(pid=17329) Worker 0 trained batch 0
  0%|          | 4/1172 [00:00<00:31, 37.48batch/s, loss=0.621]
(pid=17329) Worker 0 trained batch 1
(pid=17329) Worker 0 trained batch 2
(pid=17329) Worker 0 trained batch 3
(pid=17329) Worker 0 trained batch 4
(pid=17329) Worker 0 trained batch 5
  1%|          | 6/1172 [00:00<00:30, 38.33batch/s, loss=0.506]
  1%|          | 8/1172 [00:00<00:29, 39.31batch/s, loss=0.507]
(pid=17329) Worker 0 trained batch 6
(pid=17329) Worker 0 trained batch 7
  1%|          | 10/1172 [00:00<00:29, 40.01batch/s, loss=0.44]
(pid=17329) Worker 0 trained batch 8
(pid=17329) Worker 0 trained batch 9
(pid=17329) Worker 0 trained batch 10
(pid=17329) Worker 0 trained batch 11
  1%|          | 12/1172 [00:00<00:28, 40.83batch/s, loss=0.67] 
  1%|▏         | 15/1172 [00:00<00:27, 41.44batch/s, loss=0.537]
 Worker 0 trained batch 12
(pid=17329) Worker 0 trained batch 13
(pid=17329) Worker 0 trained batch 14
```
At the very beginning only worker with pid 17329 is training.
```
(pid=17328) Worker 0 trained batch 17
[...]

PR #3185, opened 2020-12-04
Fix missing DDP in torch distributed

hkvision:
#3184 
Fix for torch_distributed backend.

### DIFF
--- a/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
+++ b/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
@@ -195,10 +195,15 @@ class TorchRunner:
 
     def setup_operator(self):
         """Create the training operator."""
+        from torch.nn.parallel import DistributedDataParallel
+        training_models = [
+            DistributedDataParallel(model)
+            for model in self.models
+        ]
         self.training_operator =\
             self.training_operator_cls(
                 self.config,
-                models=self.models,
+                models=training_models,
                 optimizers=self.optimizers,
                 criterion=self.criterion,
                 world_rank=self.rank,
--- a/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
+++ b/pyzoo/zoo/orca/learn/pytorch/torch_runner.py
@@ -125,7 +125,7 @@ class TorchRunner:
         self.rank = hvd.rank()
         self.size = hvd.size()
         self.setup_components_horovod()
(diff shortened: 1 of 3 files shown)


pyzoo/zoo/orca/learn/pytorch/training_operator.py
```diff
@@ -193,27 +193,29 @@ def train_epoch(self, ...):
...
-                    postfix.update(loss=metrics["train_loss"])
-                _progress_bar.set_postfix(postfix)
-
-            if self.scheduler and batch_info.get(
-                    SCHEDULER_STEP) == SCHEDULER_STEP_BATCH:
-                self.scheduler.step()
-
-            metric_meters.update(metrics, n=metrics.pop(NUM_SAMPLES, 1))
-            self.global_step += 1
+        with self.model.join():
```
hkvision, line 196: This can handle uneven data for different workers, but only available at torch 1.7.0: https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html

pyzoo/zoo/orca/learn/pytorch/torch_runner.py
```diff
@@ -195,10 +195,15 @@ def setup_components_horovod(self):
 
     def setup_operator(self):
         """Create the training operator."""
+        from torch.nn.parallel import DistributedDataParallel
```
yangw1234, line 198: It seems horovod does not need this.

Maybe put it in `setup_torch_distribute`?

hkvision: okay

yangw1234:
Is it possible to write a unit test?

pyzoo/zoo/orca/learn/pytorch/training_operator.py
hkvision: I will remove this patch first since our test environment doesn't have torch 1.7.0.
Can add later if we target to support torch 1.7.0

hkvision:
> Is it possible to write a unit test?
> 
> E.g. checking if different workers has the same weights after training.

Added. Take a look.

hkvision:
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/4715/ One example test gets crashed. All other tests passed. Merge it first.

PR #3186, opened 2020-12-04
fix es_read typo

dding3:


dding3:
jekins failed. However the error is about "failed to mkdir", not related to this PR, merge it first.

http://10.239.47.210:18888/job/ZOO-PR-Validation/4692/

PR #3187, opened 2020-12-05
fix horovod pytorch estimator

### LINKED ISSUE #3182: Single node run two horovod runners get wrong rank and size
If I run Pytorch horovod backend for two workers on a single local node, `hvd.size()` is 1 for both workers, and the two workers both have `hvd.rank() == 0`, is it the correct behavior?
If so, in this case, when tqdm is enabled, the progress bar of two workers are mixed because they both have world_rank=0.

Not tested on multiple nodes, not sure if this problem only exists on single node.

### LINKED ISSUE #3184: PyTorch Ray Estimator seems not syncing weights for different workers

yangw1234:
Fix #3182
Fix #3184

### DIFF
--- a/pyzoo/test/zoo/orca/learn/ray/pytorch/test_estimator_horovod_backend.py
+++ b/pyzoo/test/zoo/orca/learn/ray/pytorch/test_estimator_horovod_backend.py
@@ -51,7 +51,7 @@ class TestPyTorchEstimator(TestCase):
                                                       validation_loss1)
         estimator.shutdown()
 
-    def test_horovod_initialized_correctly(TestCase):
+    def test_horovod_initialized_correctly(self):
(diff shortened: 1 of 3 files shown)


hkvision (approved):

yangw1234:
ut passed, example tests failures is irrelvant http://172.168.2.101:8080/job/ZOO-PR-Validation/4718/

PR #3192, opened 2020-12-07
add pytorch horovod example test

yangw1234:


yangw1234:
example test passed http://172.168.2.101:8080/job/ZOO-PR-Python-ExampleTests-py37-ray-horovod-Spark3/6/

PR #3193, opened 2020-12-07
run ray stop before ray tests started (fix ray memory test fail issue)

yangw1234:


yangw1234:
increase success rate to 4 out of 5

PR #3214, opened 2020-12-09
Change epoch and batchsize for Cifar10 in jep test

jing-xu:


hkvision (approved):

pyzoo/zoo/examples/run-example-tests-jep.sh
```diff
@@ -48,7 +48,12 @@ else
     wget -nv $FTP_URI/analytics-zoo-data/cifar10.zip -P ${ANALYTICS_ZOO_ROOT}/pyzoo/zoo/examples/orca/learn/pytorch/cifar10
     unzip ${ANALYTICS_ZOO_ROOT}/pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10.zip
 fi
-python ${ANALYTICS_ZOO_ROOT}/pyzoo/zoo/examples/orca/learn/pytorch/cifar10/cifar10.py
+
+sed "s/epochs=2/epochs=1/g;s/batch_size=4/batch_size=16/g" \
```
qiuxin2012, line 52: set batchsize 256？

jason-dai: If batchsize 256 is preferred, shall we use 256 in the example code?

hkvision: The original code use 4. 256 is only for faster training speed?

jing-xu:
http://10.239.47.210:18888/job/ZOO-PR-Pip-JEP-UT-Spark2.4/76/
with test batchsize 256, it could be finished in 223s.

pyzoo/zoo/examples/run-example-tests-jep.sh
jing-xu: The accuracy will drop a lot using the batchsize256. Keep batchsize 4 in the example and only use batchsize256 for test.

qiuxin2012: @jing-xu  To get the same accuracy, you should also increase learning rate when you increase batchsize. You can have a try.

Issue #3233, opened 2020-12-15
Upgrade torch 1.1.0 to >=1.5.0 on Jenkins


hkvision:
@pinggao18 @pinggao187

hkvision:
http://10.239.47.210:18888/job/ZOO-PR-Python-ExampleTests-py36-ray/984//consoleFull
http://10.239.47.210:18888/job/ZOO-PR-Python-ExampleTests-py36-ray/984/consoleFull
These two errors reported by @jing-xu should be related to the low version.

hkvision:
Upgraded. Closing this issue.

PR #3265, opened 2020-12-21
Add back example test for fashion-mnist

hkvision:
The example should be ok to run after fixing #3233


hkvision:
http://10.239.47.210:18888/job/ZOO-PR-Python-ExampleTests-py37-ray/1097/

PR #3268, opened 2020-12-22
Add validation during Pytorch training

### LINKED ISSUE #2997: Orca pytorch Estimator support validation when training
Currently we have to call validate manually for training. It would be better if we can pass train data and val data during fit.

hkvision:
Fix intel-analytics/BigDL#4468 

This patch is a draft for dlrm workload. Need to refactor this before merge.

### DIFF
--- a/pyzoo/zoo/orca/learn/pytorch/training_operator.py
+++ b/pyzoo/zoo/orca/learn/pytorch/training_operator.py
@@ -206,7 +206,10 @@ class TrainingOperator:
 
             if self.scheduler and batch_info.get(
                     SCHEDULER_STEP) == SCHEDULER_STEP_BATCH:
+                scheduler_start = time.time()
                 self.scheduler.step()
+                scheduler_end = time.time()
+                print("Scheduler time: ", (scheduler_end - scheduler_start) * 1000)
 
             t2 = time.time()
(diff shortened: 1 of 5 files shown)


pyzoo/zoo/orca/learn/pytorch/training_operator.py
```diff
@@ -316,13 +321,24 @@ def validate(self, val_iterator, info):
...
                 metrics = self.validate_batch(batch, batch_info)
+                scores.append(metrics.pop("output", None))
+                targets.append(metrics.pop("target", None))
                 metric_meters.update(metrics, n=metrics.pop(NUM_SAMPLES, 1))
-
+        import numpy as np
+        import sklearn.metrics
+        scores = np.concatenate(scores, axis=0)
+        targets = np.concatenate(targets, axis=0)
+        roc_auc = sklearn.metrics.roc_auc_score(targets, scores)
```
yangw1234, line 338: Do we need to collect `scores` and `targets` to the driver and calculate AUC in the driver?

hkvision: Collecting all the predictions to drivers is expensive. Wondering if we can directly reduce auc for all workers on the driver.

PR #3260, opened 2020-12-18
Unified standardize dataset for KerasModel and Orca Estimator

jenniew:
Move _standarize_feature_label_dataset to TFOptimizer.from_keras.
Move _standarize_feature_dataset to TFPredictor.from_keras

PR #3276, opened 2020-12-23
standardize data label according to how Keras standardize input_data

leonardozcm:
This issue occured to me when I found loss was not decreasing, because keras will expand a dimension to ys at the position just behind batchsize, which is 1, but orca skiped this standardizing.
And when it comes to orca, batchsize will be added to the first dimension later, so labels should be standarized as `y = tf.expand_dims(y, axis=0)`.

Reference to tensorflow keras api r1.15 and pr#3260:
- [traning_utils.py](https://github.com/tensorflow/tensorflow/blob/0396c28d6b32177a6f2df3ee459860cf292c395d/tensorflow/python/keras/engine/training_utils.py#L430)
- [training.py](https://github.com/tensorflow/tensorflow/blob/0396c28d6b32177a6f2df3ee459860cf292c395d/tensorflow/python/keras/engine/training.py#L2511)
- [#3260](https://github.com/intel-analytics/analytics-zoo/pull/3260)


jason-dai:
Is this similar to https://github.com/intel-analytics/analytics-zoo/pull/3260 @yangw1234 @jenniew ?

yangw1234:
> Is this similar to #3260 @yangw1234 @jenniew ?

Yes, this is similar.

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -619,12 +619,22 @@ def fit(self, data,
         if checkpoint_trigger is not None:
             checkpoint_trigger = Trigger.convert_trigger(checkpoint_trigger)
 
+        if is_tf_data_dataset(data):
+            def standardize_target_data(x, y):
+                if y is not None and len(y.shape) == 0:
```
yangw1234, line 624: You may need to checkout if `y` is a single tensor. There are cases where `y` has multiple tensors.

leonardozcm: ok, I will check this out.

jason-dai: If so, which PR should we merge?

yangw1234: #3260 only deals with TFNdarrayDataset (backed by a rdd of ndarray). We now find that tf.data.Dataset also needs to add this.

jenniew: yes, this PR is only for tf.data.Dataset. We may need both.

leonardozcm:
tested on jenkins, and ends with different task failed @yangw1234 
http://10.239.47.210:18888/job/ZOO-PR-Validation/4849/
http://10.239.47.210:18888/job/ZOO-PR-Validation/4848/

yangw1234:
LGTM

PR #3280, opened 2020-12-23
Add memory type support for Orca tf estimator

jenniew:
Add mem type support for Orca tf estimator fit(). 
Current support xshards and dataframe datasource.
Add ZooTrigger in python side to support multiple slice dataset.
Fix validation issue in multiple slice feature set training.
Add test cases.


pyzoo/test/zoo/orca/learn/spark/test_estimator_keras_for_spark.py
```diff
@@ -557,6 +557,105 @@ def test_submodel_in_keras_squential(self):
...
+            return result
+
+        data_shard = data_shard.transform_shard(transform)
+
+        est = Estimator.from_keras(keras_model=model)
+        est.fit(data=data_shard,
+                batch_size=4,
+                epochs=10,
+                validation_data=data_shard,
+                memory_type='2')
```
yangw1234, line 584: What does `memory_type='2'` mean? Maybe using enum type?

jenniew: memory_type=2 means creating DiskFeatureSet with 2 slices.

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
```diff
@@ -1436,6 +1436,7 @@ private[zoo] class InternalDistriOptimizer[T: ClassTag] (
       if (checkPointTrigger.isDefined) {
         if (checkPointTrigger.get.isInstanceOf[ZooTrigger]) {
           checkPointTrigger.get.asInstanceOf[ZooTrigger].setZooState(state)
+          validationTrigger.get.asInstanceOf[ZooTrigger].setZooState(state)
```
jason-dai, line 1439: `if (validationTrigger.isDefined)`?

pyzoo/zoo/tfpark/tf_optimizer.py
```diff
@@ -751,6 +754,14 @@ def optimize(self, end_trigger=None, checkpoint_trigger=None):
         if checkpoint_trigger is None:
             checkpoint_trigger = EveryEpoch()
 
+        if isinstance(self.train_data, FeatureSet):
+            if self.train_data.value.getNumOfSlice() != 1:
+                if isinstance(checkpoint_trigger, EveryEpoch):
+                    checkpoint_trigger = ZEveryEpoch()
+                elif not isinstance(checkpoint_trigger, ZooTrigger):
+                    raise Exception("Excepted zoo.util.triggers.ZooTrigger." +
+                                    " Please change your trigger to an instance of ZooTrigger.")
```
jason-dai, line 763: Please use a trigger defined in zoo.util.triggers

jason-dai:
1) Instead of changing `Estimator.fit` API, we can add a config in `OrcaContext` (such as `OrcaContext.train_data_store`) and keep `Estimator.fit` unchanged

2) We would like to have Orca users to use `orca.lean.trigger`, and need to have the Orca triggers to extend `zoo.util.triggers` (maybe in a follow-up PR)

3) For the next release, we should migrate the changes in `ZooTrigger` to BigDL triggers

pyzoo/test/zoo/orca/learn/spark/test_estimator_keras_for_spark.py
jason-dai: Instead of using '2', we should use 'DISK_2', and maybe also 'DISK' (which has a default slice value)

jenniew: In scala side, we have same definition, do we change both python and scala side to "DISK_2"?

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
jenniew: if checkPointTrigger.isDefined, internal optimizer would clone checkPointTrigger to validationTrigger. So if checkPointTrigger.isDefined,  validationTrigger should be defined.

pyzoo/test/zoo/orca/learn/spark/test_estimator_keras_for_spark.py
jason-dai: Yes, we should change all the implementations (including FeatureSet) to something like "DISK_2"

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
jason-dai: Only if `validationMethod != null && validationSet != null`

pyzoo/zoo/tfpark/tf_optimizer.py
jenniew: updated

zoo/src/main/scala/com/intel/analytics/zoo/pipeline/api/keras/models/Topology.scala
jenniew: updated

pyzoo/test/zoo/orca/learn/spark/test_estimator_keras_for_spark.py
jenniew: updated

jenniew:
Change disk memory type from "n" to "DISK_n".
Use OrcaContext.train_data_store to set  train memory type without change Orca tf estimator fit().

jason-dai:
LGTM

jason-dai:
You also need to update https://github.com/intel-analytics/analytics-zoo/blob/master/docs/readthedocs/Orca/Overview/orca-context.md; see https://github.com/intel-analytics/analytics-zoo/pull/3287 for an example.

docs/readthedocs/Orca/Overview/orca-context.md
```diff
@@ -36,6 +36,8 @@ Under the hood, `OrcaContext` will automatically provision Apache Spark and/or R
  - log_output: shall we always set it to true?
  - pandas_read_backend: we may describe it in xShards?
  - serialize_data_creation: shall we always set it to true?
+ * `OrcaContext.train_data_store`: Setting memory type for train data storage. Either `DRAM`, `PMEM`, or `DISK_n`. The default value is `DRAM`, you can change it to `PMEM` if have AEP hardware. If you give `DISK_n`, in which `n` is an integer, we will cache the data into disk, and hold only `1/n` of the data in memory. After going through the `1/n`,
```
jason-dai, line 39: Default to "DRAM"; setting it to `DISK_n` (e.g., `DISK_2`) if the training data cannot fit in memory (this will store the data on disk, and cache only `1/n` of the data in memory; after going through the `1/n`, we will release the current cache, and load another `1/n` into memory).

jason-dai:
You also need to make `orca.learn.trigger` subclasses of `zoo.util.triggers`, probably in a follow-up PR.

jenniew:
> You also need to make `orca.learn.trigger` subclasses of `zoo.util.triggers`, probably in a follow-up PR.

OK

hkvision:
> You also need to update https://github.com/intel-analytics/analytics-zoo/blob/master/docs/readthedocs/Orca/Overview/orca-context.md; see #3287 for an example.

@jenniew You can update this after I merge that PR.

jenniew:
Jenkins: http://10.239.47.210:18888/job/ZOO-PR-Validation/4895/, failed with unrelated case

Issue #2623, opened 2020-07-27
update sklearn version of automl


yushan111:
We should add docs and update setup that we only support sklearn < 0.23 or we should add sklearn 0.23 support.

yushan111:
related issue #2621

yushan111:
Sklearn >= 0.23 will throw error：RuntimeError: The reset parameter is False but there is no n_features_in_ attribute. Is this estimator fitted?

The error message is not clear enough. We should capture the error and prompt to downgrade sklearn.

PR #3304, opened 2020-12-30
automl feature transform sklearn >=0.23 support

TheaperDeng:
**Goal**:

Fix issue #2623.

**Change log**:

I add the conditional attribute `n_features_in_` setting in `_scale` function.

**Reason**:

This might be faster and better for future support.

**Reference**:

In sklearn version 0.23.X, it set a strict check on the attribute `n_features_in_`, while we only save `mean_` and `scale_` for the `StandardScaler`. This is the main reason for issue #2623.   
In sklearn version 0.24.X, this strict check is eliminated for stateless estimator(e.g. `StandardScaler`), still it might be added back later. see discussions sklearn pr 18011 and [here](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep010/proposal.html)

### DIFF
--- a/pyzoo/zoo/automl/feature/time_sequence.py
+++ b/pyzoo/zoo/automl/feature/time_sequence.py
@@ -17,6 +17,7 @@
 from zoo.automl.common.util import save_config
 from zoo.automl.feature.abstract import BaseFeatureTransformer
 
+import sklearn
 from sklearn.preprocessing import MinMaxScaler, StandardScaler
 import pandas as pd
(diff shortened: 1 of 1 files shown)


TheaperDeng:
PR validation passed(combination):
http://10.239.47.210:18888/job/ZOO-PR-Validation/4878/
http://10.239.47.210:18888/job/ZOO-PR-Python-Spark-2.3-py36k216/3994/

PR #3334, opened 2021-01-07
[WIP]Transfer spark rdd to ray objects

yushan111:


pyzoo/zoo/orca/data/utils.py
```diff
@@ -309,3 +312,105 @@ def get_size(x):
...
+    port = ip2port[ip]
+    address = ("localhost", port)
+    pa_schema = to_arrow_schema(schema)
+
+    import time
+    counter = 0
+    serialize_time = 0.0
+    sending_time = 0.0
+    total_bytes = 0
+    with Client(address) as conn:
```
yangw1234, line 382: Wrap this Client as a class such as `PartitionWriter`? 

Then we can have subclasses like `MultiProcessingConnectionWriter`, `PlasmaWriter`.

```python
class PartitionWriter:
    def __init__(self, partition_id)
    
    def write_batch(self, batch)

    def close(self)
```

pyzoo/zoo/orca/data/partition_holder.py
```diff
@@ -0,0 +1,214 @@
...
+    label = [v for k, v in result_dict if k in label_cols]
+
+    return data, label
+
+
+def numpy_to_torch_data_loader(data, label, params):
+    from torch.utils.data import Dataset, DataLoader
+    from zoo.orca.data.utils import index_data, get_size
+
+    class NDArrayDataset(Dataset):
```
yangw1234, line 148: `torch.utils.data.TensorDataset` does almost exactly this.

pyzoo/zoo/orca/data/partition_holder.py
```diff
@@ -0,0 +1,214 @@
...
+from zoo.orca.data.utils import deserialize_using_pa_stream, write_to_ray_python_client
+from zoo.orca.learn.pytorch.utils import find_free_port
+
+
+class PartitionHolder:
+    # for ray actor
+    def __init__(self, max_connections=8):
+        import queue
+        from concurrent.futures import ThreadPoolExecutor
+        self.data_queue = queue.Queue()
```
yangw1234, line 14: wrap the reading logic into a `DataReceiver`? Then we can have `MultiProcessConnReceiver` and `PlasmaReceiver`.

```python
class DataReceiver:
    def start(self):
     
    def close(self):
     
    def get_data_dict(self)
```

Issue #3329, opened 2021-01-06
In python 3.7.3 orca.estimitor.predict throw exception "ValueError: Can not deserialize PairRDD with different number of items in batches"


jack1981:
When we run zoo.orca.learn.tf.estimator.predict , throw this exception , only happened at python 3.7.3 , 3.6 was fine
Sample code : 

from zoo.orca.learn.tf.estimator import Estimator
    #Orca the predict function supports native spark data frame ! Just need to tell batch_size and feature_cols
    prediction_df = est.predict(inferenceDF, batch_size=batch_size, feature_cols=['features'])

Exception information : 
        at java.lang.Thread.run(Thread.java:748)
Caused by: org.apache.spark.SparkException: Job aborted due to stage failure: Task 202 in stage 98677.0 failed 1 times, most recent failure: Lost task 202.0 in stage 98677.0 (TID 17564, localhost, executor driver): org.apache.spark.api.python.PythonException: Traceback (most recent call last):
  File "/opt/work/spark-2.4.3/python/lib/pyspark.zip/pyspark/worker.py", line 377, in main
    process()
  File "/opt/work/spark-2.4.3/python/lib/pyspark.zip/pyspark/worker.py", line 372, in process
    serializer.dump_stream(func(split_index, iterator), outfile)
  File "/opt/work/spark-2.4.3/python/lib/pyspark.zip/pyspark/serializers.py", line 393, in dump_stream
    vs = list(itertools.islice(iterator, batch))
  File "/opt/work/spark-2.4.3/python/lib/pyspark.zip/pyspark/serializers.py", line 460, in _load_stream_without_unbatching
    " in batches: (%d, %d)" % (len(key_batch), len(val_batch)))
ValueError: Can not deserialize PairRDD with different number of items in batches: (23, 26)

        at org.apache.spark.api.python.BasePythonRunner$ReaderIterator.handlePythonException(PythonRunner.scala:452)
        at org.apache.spark.api.python.PythonRunner$$anon$1.read(PythonRunner.scala:588)
        at org.apache.spark.api.python.PythonRunner$$anon$1.read(PythonRunner.scala:571)
        at org.apache.spark.api.python.BasePythonRunner$ReaderIterator.hasNext(PythonRunner.scala:406)
        at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
        at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:440)
[...]

PR #3341, opened 2021-01-08
Fix Orca predict on Dataframe on python 3.7.3

jenniew:
Select column operation on Spark Dataframe on python 3.7.3 would change partition for big data. Remove select operation when change dataframe to tfdataset.


jenniew:
Related issue：https://github.com/intel-analytics/analytics-zoo/issues/3329

jenniew:
Jenkins: http://10.239.47.210:18888/job/ZOO-PR-Validation/4909/

PR #3344, opened 2021-01-08
Fix mem type test

jenniew:
Add test data amount to avoid rdd sample unbalanced issue.


jenniew:
Jenkins: http://10.239.47.210:18888/job/ZOO-PR-Validation/4907/

PR #3361, opened 2021-01-11
update es doc format

dding3:


pyzoo/zoo/orca/data/elastic_search.py
```diff
@@ -94,13 +96,14 @@ def write_df(esConfig, esResource, df):
...
                elastic search(eg. ip, port, es query etc).
-        :param esResource Optional. resource file in elastic search.
+        :param esResource: Optional. resource file in elastic search.
                It also can be set in esConfig
-        :param filter Optional. Request only those fields from Elasticsearch
-        :param esQuery Optional. es query
-        :return Spark DataFrame. Each row represents a document in ES.
+        :param filter: Optional. Request only those fields from Elasticsearch
+        :param esQuery: Optional. es query
+        :return: Spark DataFrame. Each row represents a document in ES.
```
jason-dai, line 106: Does it return Dataframe or RDD?

dding3: It will return RDD. Updated the doc

jason-dai:
LGTM

PR #3380, opened 2021-01-14
clean up orca estimators predict

yangw1234:
- [x] if input is a xshard, then output will another xshard with an additional key "prediction"
- [x] if input is a dataframe, then output will be another dataframe with an additional column "prediction"
- [x] unit tests


yangw1234:
http://172.168.2.101:8080/view/ZOO-PR/job/ZOO-PR-Validation/4973/ jenkins passed

PR #3405, opened 2021-01-20
Update index.rst

helenlly:
release.md is referenced by other pages and created, so remove its link from index.rst.  
 after remove,  release download page will not show in home page and left menu.

tets demo: https://testhelendoc.readthedocs.io/en/latest/


jason-dai:
LGTM

Issue #3401, opened 2021-01-19
Using Orca Estimator get_model method cannot return keras model object.


GitEasonXu:
Using Orca Estimator get_model method cannot return keras model object. Maybe it is related to `BN` layer.
```python
est.save("/tmp/yolov1.ckpt")
est.restore("/tmp/yolov1.ckpt")
model = est.get_model()
model.save("/tmp/yolov1.h5")
```

GitEasonXu:
```python
    def get_model(self):
        """Returns the learned model."""
        state = ray.get(self.remote_workers[0].get_state.remote())
        #state_refs = [w.get_state.remote() for w in self.remote_workers]
        #state = ray.get(state_refs[0])
        
        return self._get_model_from_state(state)
```
Should I change `state = ray.get(self.remote_workers[0].get_state.remote())` to `state_refs = [w.get_state.remote() for w in self.remote_workers]  state = ray.get(state_refs[0])`

yangw1234:
> ```python
>     def get_model(self):
>         """Returns the learned model."""
>         state = ray.get(self.remote_workers[0].get_state.remote())
>         #state_refs = [w.get_state.remote() for w in self.remote_workers]
>         #state = ray.get(state_refs[0])
>         
>         return self._get_model_from_state(state)
> ```
> 
> Should I change `state = ray.get(self.remote_workers[0].get_state.remote())` to `state_refs = [w.get_state.remote() for w in self.remote_workers] state = ray.get(state_refs[0])`
Yeah, I think that's the correct fix. Could you try if that works?

GitEasonXu:
Need change the two methods, and I have tested it without problems.
```python     
def get_model(self):
        """Returns the learned model."""
        #state = ray.get(self.remote_workers[0].get_state.remote())

        state_refs = [w.get_state.remote() for w in self.remote_workers]
        state = ray.get(state_refs[0])

        return self._get_model_from_state(state)
```
```python
    def _get_model_from_state(self, state):
        """Creates model and load weights from state"""

        model = self.model_creator(self.config)
        model.set_weights(state["weights"])

        # This part is due to ray.get() changing scalar np.int64 object to int
        if state["optimizer_weights"]:
            state["optimizer_weights"] = np.array(
                state["optimizer_weights"], dtype=np.int64)

            if model.optimizer.weights == []:
                model.make_train_function()
            model.optimizer.set_weights(state["optimizer_weights"])
        
        return model
```

[...]

PR #3418, opened 2021-01-22
Avoid workers' waiting and fix _get_model_from_state

leonardozcm:
Reference to #3401


leonardozcm:
PR in testing

pyzoo/zoo/orca/learn/tf2/estimator.py
```diff
@@ -362,11 +364,15 @@ def _get_model_from_state(self, state):
...
-            state["optimizer_weights"][0], dtype=np.int64)
-
-        if model.optimizer.weights == []:
-            model._make_train_function()
-        model.optimizer.set_weights(state["optimizer_weights"])
+        if state["optimizer_weights"]:
+            if model.optimizer.weights == []:
+                model.make_train_function()
+
+            if not len(model.optimizer.get_weights()) == 0 and \
```
yangw1234, line 371: make some comments on why we should do this?

leonardozcm: > make some comments on why we should do this?
It is a incorrect implemention, as for `model.optimizer.weight` is still empty after `model.make_train_function()` called, and this implement just skip `set_weights` step to avoid this probelm.

yangw1234:
http://10.239.47.210:18888/job/ZOO-PR-Validation/4992/

PR #3419, opened 2021-01-22
add transfer learning tf example

KimiLiuQh:
add the 'transfer_learning.ipynb'  for tf example with save model


Litchilitchy:
This is moved to [here](https://github.com/intel-analytics/analytics-zoo/pull/3425)

PR #3425, opened 2021-01-25
Cluster Serving transformation from keras guide

Litchilitchy:


jason-dai:
In general it looks good; some comments:

1) Break the notebook into several sections, such as:
    - Original Keras application
    - Export TensorFlow Saved Model
    - Deploy Cluster Serving
    - Predict using Cluster Serving
2) Show more details for Cluster Serving, including example config file, model directory, launch command, prediction result, etc.
3) Reduce the output (e.g., just predict one sample in the Keras application, no need to show the model topology, etc.)

Litchilitchy:
@jason-dai refine PR is at #3452

PR #3432, opened 2021-01-26
Fix batch_size issue

xiejinglei:
Previously if `config['batch_size']` is specified but `batch_size` is not specified in the `fit` function (TF2 Estimator), `config['batch_size']` will be over-written by 32. This PR aims to fix this issue.


hkvision:
The change looks good.
Shall we add the documentation in fit for the default value to be 32 since user may not know what None means?

jason-dai:
To be consistent with other estimator APIs, we should simply use the `batch_size` specified in `fit`, reporting error or warning if 'batch_size' is also specified in `config`

hkvision:
> To be consistent with other estimator APIs, we should simply use the `batch_size` specified in `fit`, reporting error or warning if 'batch_size' is also specified in `config`

In the data_creator, users would call config["batch_size"] to retrieve the value. If we directly report an error, it may be confusing to users why they can call config["batch_size"]. I suggest make this a warning instead of an error.

xiejinglei:
Thanks for the comments. I have updated the changes.

jason-dai:
> > To be consistent with other estimator APIs, we should simply use the `batch_size` specified in `fit`, reporting error or warning if 'batch_size' is also specified in `config`
> 
> In the data_creator, users would call config["batch_size"] to retrieve the value. If we directly report an error, it may be confusing to users why they can call config["batch_size"]. I suggest make this a warning instead of an error.

It's better to 

> > To be consistent with other estimator APIs, we should simply use the `batch_size` specified in `fit`, reporting error or warning if 'batch_size' is also specified in `config`
> 
> In the data_creator, users would call config["batch_size"] to retrieve the value. If we directly report an error, it may be confusing to users why they can call config["batch_size"]. I suggest make this a warning instead of an error.

Then shall we make `batch_size` part of the `data_creater` arguments @yangw1234 ?

readthedocs/source/doc/Orca/QuickStart/orca-tf2keras-quickstart.md
```diff
@@ -95,14 +95,15 @@ First, create an Estimator.
 ```python
 from zoo.orca.learn.tf2 import Estimator
 
-batch_size = 320
-est = Estimator.from_keras(model_creator=model_creator, config={"batch_size": batch_size}, workers_per_node=2)
+est = Estimator.from_keras(model_creator=model_creator, config={}, workers_per_node=2)
```
hkvision, line 98: if config is empty, then just remove this argument? (it is default to be None)

xiejinglei: Ok, I have removed `config={}`.

yangw1234:
> > > To be consistent with other estimator APIs, we should simply use the `batch_size` specified in `fit`, reporting error or warning if 'batch_size' is also specified in `config`
> > 
> > 
> > In the data_creator, users would call config["batch_size"] to retrieve the value. If we directly report an error, it may be confusing to users why they can call config["batch_size"]. I suggest make this a warning instead of an error.
> 
> It's better to
> 
> > > To be consistent with other estimator APIs, we should simply use the `batch_size` specified in `fit`, reporting error or warning if 'batch_size' is also specified in `config`
> > 
> > 
> > In the data_creator, users would call config["batch_size"] to retrieve the value. If we directly report an error, it may be confusing to users why they can call config["batch_size"]. I suggest make this a warning instead of an error.
> 
> Then shall we make `batch_size` part of the `data_creater` arguments @yangw1234 ?

Yeah, making it an argument to `data_creator` sounds good to me. If then, maybe we should add it to all "creators" argument. For example, some models (tf object detection) might need this `batch_size` to construct the graph. @jason-dai

hkvision:
But for SparkXShards input, we still need batch_size in fit?

yangw1234:
> But for SparkXShards input, we still need batch_size in fit? In this case, putting batch_size in creator would be confusing and make the API inconsistent...

Users still pass `batch_size` in fit, they just use the local batch_size in `xxx_creator`, which passed by us, right?

hkvision:
> > But for SparkXShards input, we still need batch_size in fit? In this case, putting batch_size in creator would be confusing and make the API inconsistent...
> 
> Users still pass `batch_size` in fit, they just use the local batch_size in `xxx_creator`, which passed by us, right?

OK. This makes sense

xiejinglei:
I have changed the default `batch_size=32` in the `fit` function to be `None` and added `batch_size` to the arguments of `data_creator` and `validation_data_creator`. If the changes are fine, I will also change the document accordingly.

pyzoo/zoo/orca/learn/tf2/tf_runner.py
```diff
@@ -320,18 +322,33 @@ def setup_distributed(self, urls, world_rank, world_size):
...
              callbacks=None, validation_data_creator=None, class_weight=None,
              steps_per_epoch=None, validation_steps=None, validation_freq=1,
              data_config=None):
         """Runs a training epoch and updates the model parameters."""
         config = self.config.copy()
         if data_config is not None:
             config.update(data_config)
-        config['batch_size'] = batch_size
+
+        if 'batch_size' not in config:
```
hkvision, line 334: After the change, I think we should expect users to specify batch_size in config. So need to modify this.

xiejinglei: I have updated the code so that the default value of `batch_size` is again 32, and if a user specify `batch_size` in `config`, an exception will be raised, telling him/her to input `batch_size` in the `fit` function.

readthedocs/source/doc/Orca/QuickStart/orca-tf2keras-quickstart.md
```diff
@@ -95,14 +95,15 @@ First, create an Estimator.
 ```python
 from zoo.orca.learn.tf2 import Estimator
 
-batch_size = 320
-est = Estimator.from_keras(model_creator=model_creator, config={"batch_size": batch_size}, workers_per_node=2)
+est = Estimator.from_keras(model_creator=model_creator, workers_per_node=2)
 ```
 
-Next, fit and evaluate using the Estimator.
+Next, fit and evaluate using the Estimator. Note that if `batch_size` is not specified in the `fit` function, it will be set to 32 by default.
```
hkvision, line 101: Is this still needed? I think we can remove this one.

hkvision:
I think we need to modify the unit tests for the change: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/test/zoo/orca/learn/ray/tf/test_tf_ray_estimator.py#L50

readthedocs/source/doc/Orca/QuickStart/orca-tf2keras-quickstart.md
xiejinglei: Removed now

xiejinglei:
> I think we need to modify the unit tests for the change: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/test/zoo/orca/learn/ray/tf/test_tf_ray_estimator.py#L50

I have just modified several files including this one to adapt to the changes.

hkvision (approved):

docs/docs/colab-notebook/orca/quickstart/tf2_keras_lenet_mnist.ipynb
```diff
@@ -253,27 +253,27 @@
...
+        "    dataset = dataset.batch(batch_size)\n",
+        "    return dataset\n",
+        "\n",
+        "def val_data_creator(config, batch_size):\n",
+        "    _, (val_feature, val_label) = tf.keras.datasets.mnist.load_data()\n",
+        "\n",
+        "    dataset = tf.data.Dataset.from_tensor_slices((val_feature, val_label))\n",
+        "    dataset = dataset.repeat()\n",
+        "    dataset = dataset.map(preprocess)\n",
+        "    dataset = dataset.batch(batch_size)\n",
```
jason-dai, line 276: remove `config={}` in step 4 below.

pyzoo/zoo/orca/learn/tf2/tf_runner.py
```diff
@@ -328,10 +329,15 @@ def step(self, data_creator, epochs=1, batch_size=32, verbose=1,
         config = self.config.copy()
         if data_config is not None:
             config.update(data_config)
+
+        if 'batch_size' in config:
```
jason-dai, line 333: change to "batch_size"

pyzoo/zoo/orca/learn/tf2/tf_runner.py
```diff
@@ -328,10 +329,15 @@ def step(self, data_creator, epochs=1, batch_size=32, verbose=1,
...
+            raise Exception("Please do not specify batch_size in config. Input batch_size in the fit function of the "
+                            "estimator instead.")
         config['batch_size'] = batch_size
```
jason-dai, line 336: change to "batch_size"

pyzoo/zoo/orca/learn/tf2/tf_runner.py
```diff
@@ -328,10 +329,15 @@ def step(self, data_creator, epochs=1, batch_size=32, verbose=1,
         config = self.config.copy()
         if data_config is not None:
             config.update(data_config)
+
+        if 'batch_size' in config:
+            raise Exception("Please do not specify batch_size in config. Input batch_size in the fit function of the "
+                            "estimator instead.")
```
jason-dai, line 335: Check for the same in `validate`?

hkvision: let's put this in the constructor to avoid duplicates

xiejinglei: I think moving it to the construction method could be a better idea. I will try this.

docs/docs/colab-notebook/orca/quickstart/tf2_keras_lenet_mnist.ipynb
xiejinglei: Fixed. Thanks

pyzoo/zoo/orca/learn/tf2/tf_runner.py
xiejinglei: Fixed now.

xiejinglei:
PR validation has been completed. Link: http://10.239.47.210:18888/job/ZOO-PR-Validation/5038/
For the two cases that do not pass in the above link, they are passed here: http://10.239.47.210:18888/job/ZOO-PR-Python-ExampleTests/2238/
(Failures may happen due to random system/connection errors)

PR #3436, opened 2021-01-27
support multi-inputs model in bigdl estimator when data is xshards

yangw1234:


cyita:
Maybe add a test case of a multi-inputs model?

yangw1234:
> Maybe add a test case of a multi-inputs model?

Good catch! Added.

PR #3451, opened 2021-01-29
Orca bigdl Estimator init tensorboard related factors

leonardozcm:
Interface to initiialize `app_name` and `logdir` for `tensorboard` or `TrainSummary`.


pyzoo/zoo/orca/learn/bigdl/estimator.py
```diff
@@ -275,6 +275,12 @@ def set_l2_norm_gradient_clipping(self, clip_norm):
         self.nn_estimator.setGradientClippingByL2Norm(clip_norm)
         self.estimator.set_l2_norm_gradient_clipping(clip_norm)
 
+    def set_tensorboard(self, log_dir, app_name):
```
yangw1234, line 278: already defined in super class? https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/spark_estimator.py#L97

PR #3474, opened 2021-02-07
fix python style check

Le-Zheng:


Le-Zheng:
http://10.239.47.210:18888/job/ZOO-PR-Python-Spark-3.0-py37/462/console

PR #3476, opened 2021-02-07
fix default arguments (driver port)

Le-Zheng:
fix http://10.239.47.210:18888/view/ZOO-NB/job/ZOO-NB-Python-integration-test-horovod/15/

### DIFF
--- a/pyzoo/zoo/examples/orca/learn/horovod/pytorch_estimator.py
+++ b/pyzoo/zoo/examples/orca/learn/horovod/pytorch_estimator.py
@@ -124,7 +124,7 @@ if __name__ == "__main__":
                              "You can change it with your k8s image.")
     parser.add_argument('--k8s_driver_host', type=str, default="",
                         help="The k8s driver localhost.")
-    parser.add_argument('--k8s_driver_port', type=int, default="",
+    parser.add_argument('--k8s_driver_port', type=str, default="",
                         help="The k8s driver port.")
 
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


Le-Zheng:
jenkins http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Python-integration-test-horovod/3/console passed

pyzoo/zoo/examples/orca/learn/horovod/pytorch_estimator.py
```diff
@@ -124,7 +124,7 @@ def train_example(workers_per_node):
                              "You can change it with your k8s image.")
     parser.add_argument('--k8s_driver_host', type=str, default="",
                         help="The k8s driver localhost.")
-    parser.add_argument('--k8s_driver_port', type=int, default="",
+    parser.add_argument('--k8s_driver_port', type=str, default="",
```
hkvision, line 127: You shouldn't change this back to str. Instead you should change the default value to None or some default int.

Le-Zheng: Why should use `port` as int instead of string? What benefit we can get?

hkvision: To avoid wrong user input. If it is string, what if user input something like "xxxxx"?

Le-Zheng: However, if it is int, what if users input something like 0 or -1000. 
Using "empty-string" as the default port looks more resonable https://stackoverflow.com/questions/47992477/why-is-port-a-string-and-not-an-integer
We may add a check (string to int) to check if it is a valid port input.

PR #3479, opened 2021-02-08
Fix install java doc on centos

hkvision:
- Fix JAVA_HOME on centos
- Fix typo for OpenJDK


hkvision:
@jason-dai Seems on centos, java path is different from ubuntu (tested on clx node vs almaren node), update the doc. Take a look.

hkvision:
https://hkvision-jdk.readthedocs.io/en/latest/doc/UserGuide/python.html

PR #3480, opened 2021-02-08
Cluster Serving auto set flink parallelism

Litchilitchy:
@jason-dai user can just set `model_num`, model number in serving pipeline  and `core_num`, core number each model would use.

To make this more readable, in code, `coreNum` is renamed to `thrdPerModel`

The whole pipeline parallelism will be controlled by `model_num`, this would be consistent with inference model design

### DIFF
--- a/scripts/cluster-serving/config.yaml
+++ b/scripts/cluster-serving/config.yaml
@@ -11,9 +11,9 @@ data:
   # default, None
   filter:
 params:
-  # default, 4
+  # default, number of cores per model in serving
   core_number:
-  # default: model number will auto-adapt to core_number, do not set it if not sure about the behavior
+  # default: number of models in serving
   model_number:
   # default: OFF
   performance_mode:
--- a/zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
+++ b/zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
@@ -42,7 +42,8 @@ object ClusterServing {
       .text("Config Path of Cluster Serving")
       .action((x, params) => params.copy(configPath = x))
     opt[Int]('t', "testMode")
-      .text("Text Mode of Parallelism 1")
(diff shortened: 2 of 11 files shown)


scripts/cluster-serving/config.yaml
```diff
@@ -11,9 +11,9 @@ data:
   # default, None
   filter:
 params:
-  # default, 4
+  # default, number of cores per model in serving
```
jason-dai, line 14: number of cores (or threads) used per model in serving

scripts/cluster-serving/config.yaml
```diff
@@ -11,9 +11,9 @@ data:
...
   core_number:
-  # default: model number will auto-adapt to core_number, do not set it if not sure about the behavior
+  # default: number of models in serving
```
jason-dai, line 16: number of models used in serving

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
```diff
@@ -42,7 +42,8 @@ object ClusterServing {
       .text("Config Path of Cluster Serving")
       .action((x, params) => params.copy(configPath = x))
     opt[Int]('t', "testMode")
-      .text("Text Mode of Parallelism 1")
+      .text("Text Mode controlling Flink parallelism, this should not be controlled by user" +
```
jason-dai, line 45: Test Mode?

Litchilitchy:
@jason-dai a new pr #3674 is added for modification of this.

zoo/src/main/scala/com/intel/analytics/zoo/serving/ClusterServing.scala
Litchilitchy: this is just for IDE debug quick config, will not used by user. Is this not proper? I would remove it if so.

PR #3515, opened 2021-02-23
add ppml docker for ppml/trusted-big-data-ml

glorysdj:
ppml/trusted-big-data-ml is tested and ready


qiuxin2012 (approved):

jason-dai:
1) In the current implementation, it is shown as a Docker to run a specific example in Analytics Zoo; instead, we should make it a general solution in which the user can develop and run any Spark application (Dataframes, MLlib, BigDL, etc.)

2) How can we automatically test it?

glorysdj:
> 1. In the current implementation, it is shown as a Docker to run a specific example in Analytics Zoo; instead, we should make it a general solution in which the user can develop and run any Spark application (Dataframes, MLlib, BigDL, etc.)

Yes, we will change the script and the doc to support to run any spark app.

for 2, we may need a jenkins slave on sgx server to run automation tests. once we have a free sgx server, we will try to enable a jenkins job.

PR #3529, opened 2021-03-02
Give the plug-in customized model a better way to save/load

TheaperDeng:
Major changes:

1. When RaySearcher find out that a customized model is used, then it will call the save directly.
2. Add a restore method for modelbuilder, so that user can restore easily from the checkpoint even if they stop the search midway.
3. Change the plugin model demo and show the usage of this new easy save/load

Some minor changes:

1. Change base_pytorch_model's evaluation method's param from metric to metrics
2. Add a comparation between searched model and manually fixed hyperparam model, which demos that searched model has a better result.

### DIFF
--- a/pyzoo/test/zoo/automl/model/test_tcn.py
+++ b/pyzoo/test/zoo/automl/model/test_tcn.py
@@ -51,7 +51,7 @@ class TestTcn(TestCase):
         self.model.fit_eval(self.train_data[0], self.train_data[1], self.val_data, **config)
         mse, smape = self.model.evaluate(self.val_data[0],
                                          self.val_data[1],
-                                         metric=["mse", "smape"])
+                                         metrics=["mse", "smape"])
         assert len(mse) == self.val_data[1].shape[-1] * self.val_data[1].shape[-2]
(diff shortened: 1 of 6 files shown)


yushan111:
Current usage: 
```python
    # rebuild this best config model and evaluate
    best_model = modelBuilder.restore(best_trials[0].model_path)
    searched_best_model = best_model.evaluate(data["val_x"], data["val_y"], metrics=["rmse"])
```

It is a little wired to restore from `modelBuilder` and get a model.
I recommend to change `restore` to name like `build_from_ckpt`

yushan111:
Besides, it is a little messy to demonstrate all usages like restore from ckpt, build from pytorch, build from keras in one demo. We could split into two examples, one for pytorch and another for keras. Or we can also add some UTs for these different usages.

TheaperDeng:
PR combine pass:
http://10.239.47.210:18888/job/ZOO-PR-Validation/5158/
http://10.239.47.210:18888/job/ZOO-PR-Python-ExampleTests-py37-Spark3/589/

PR #3535, opened 2021-03-02
hot fix stylecheck 

Le-Zheng:


Le-Zheng:
http://10.239.47.210:18888/job/ZOO-PR-Python-Spark-2.4-py37/1950/console

PR #3560, opened 2021-03-08
Enable Cluster Serving message encrytpion

qiyuangong:


zoo/src/main/scala/com/intel/analytics/zoo/serving/preprocessing/PreProcessing.scala
```diff
@@ -26,6 +26,7 @@ import org.apache.log4j.Logger
 
 import scala.collection.mutable.ArrayBuffer
 import com.intel.analytics.bigdl.utils.{T, Table}
+import com.intel.analytics.zoo.serving.http.FrontEndApp.decryptWithAES256
```
glorysdj, line 29: EncryptSupportive is a trait, PreProcessing can extends it, and call encryptWithAES256/decryptWithAES256 directly

zoo/src/main/scala/com/intel/analytics/zoo/serving/utils/ClusterServingHelper.scala
```diff
@@ -80,6 +80,7 @@ class ClusterServingHelper(_configPath: String = "config.yaml", _modelDir: Strin
   var redisSecureTrustStoreToken: String = null
 
   var modelEncrypted: Boolean = false
+  var recordEncrypted: Boolean = false
```
glorysdj, line 83: we really should define a class of the config yaml, and parse the object directly, not add a field then add a method...

zoo/src/main/scala/com/intel/analytics/zoo/serving/preprocessing/PreProcessing.scala
qiyuangong: Already addressed. Thanks

glorysdj (approved):
LGTM

qiyuangong:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/5196, excepted a TF related fail, which is not related to this PR,

PR #3566, opened 2021-03-10
fix tf2 notebook [ray upgrade followup]

yangw1234:


yangw1234:
http://172.168.2.101:8080/job/ZOO-PR-Quickstart-notebook/110/ notebook tests passed

PR #3579, opened 2021-03-11
add shutdown tf.sess

Le-Zheng:


pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -704,6 +704,9 @@ def set_l2_norm_gradient_clipping(self, clip_norm):
         """
         raise NotImplementedError
 
+    def shutdown(self):
```
jason-dai, line 707: Add Python doc

jason-dai:
We should also move `shutdown` from ray_estimator (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/ray_estimator.py#L86) to base_estimator (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/base_estimator.py#L20)

pyzoo/zoo/orca/learn/tf/estimator.py
```diff
@@ -704,6 +704,12 @@ def set_l2_norm_gradient_clipping(self, clip_norm):
...
+        """
+        Close session and release resources.
```
jason-dai, line 709: Close TensorFlow session

pyzoo/zoo/orca/learn/base_estimator.py
```diff
@@ -79,3 +79,12 @@ def load(self, checkpoint):
...
         raise NotImplementedError
+
+    @abstractmethod
+    def shutdown(self):
+        """
+        Releases resources.
+
+        :return:
+        """
+        raise NotImplementedError
```
jason-dai, line 90: I think you can just use `pass` here, so that there is no need to implement it in every subclass

Le-Zheng: `pass` still throws instantiate abstract class error.  spark_estimator/ray_estimator needs to implement all abstract methods of BaseEstimator. Subclass of spark_estimator needs too.

jason-dai: > `pass` still throws instantiate abstract class error. spark_estimator/ray_estimator needs to implement all abstract methods of BaseEstimator. Subclass of spark_estimator needs too.

First, I don't think you should throw `NotImplementedError` in all the Spark-based estimators; instead, just `pass` if there is nothing to shutdown.

Secondly, can we make `shutdown` in `spark_estimator` a non-static method that `pass`, so that subclass of spark_estimator does not need to implement it if there is nothing to shutdown?

Le-Zheng: > > `pass` still throws instantiate abstract class error. spark_estimator/ray_estimator needs to implement all abstract methods of BaseEstimator. Subclass of spark_estimator needs too.
> 
> First, I don't think you should throw `NotImplementedError` in all the Spark-based estimators; instead, just `pass` if there is nothing to shutdown.
> 
> Secondly, can we make `shutdown` in `spark_estimator` a non-static method that `pass`, so that subclass of spark_estimator does not need to implement it if there is nothing to shutdown?

Okay. `shutdown` in BaseEstimator and spark_estimator use `pass`.  And using non-static `shutdown`  does not need to implement in the subclass.

jason-dai:
LGTM

Le-Zheng:
Jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/5173/

Issue #3345, opened 2021-01-08
Unify orca estimators' save and load


yangw1234:
## Current Issue

For spark based estimators, the file `saved` cannot be `loaded`, because `save` saves the framework format (tf checkpoint, keras h5, ...), `load` looks for orca checkpoint, which is never explicitly saved by calling `save`, but saved automatically by bigdl DistriOptimizer.

## Proposal
1. all `save` use the underlining framework's format, and should be able to loaded using `load` of the same estimator.
2. for spark estimators, add additionally `load_orca_checkpoint` to load the checkpoint produced by bigdl DistriOptimizer.

## Tasks

- [x] implement spark pytorch `save` using pytorch format
- [x] change spark pytorch `load` to load pytorch format, add `load_orca_checkpoint` API using the original `load` logic
- [x] change spark tf/keras load to use tf/keras format, add `load_orca_checkoupt` API using the original `load` logic

xiejinglei:
#3581 and #3665 fixes PyTorch Estimator and #3628 fixes TF Estimator.

PR #3581, opened 2021-03-12
save and load methods for PytorchSparkEstimator

xiejinglei:
This PR relates to issue #3345. 

1. Since `PytorchSparkEstimator` accepts both pytorch optimizer and orca optimizer, I use the `save` mothod of `bigdl.optim.optimizer.OptimMethod` to save the optimizer state. For the model, I save the `state_dict` directly. Currently I don't find  a way to save both of them together, so I save them in 2 separate but related files.

2. `load_orca_checkpoint` already exists in the class, so I just change the original `load` function to `load_orca_model_or_checkpoint`. 

Please have a look if the modification logic is ok. Thanks!


pyzoo/zoo/orca/learn/pytorch/estimator.py
```diff
@@ -264,11 +265,11 @@ def __init__(self, model, loss, optimizer, metrics=None, model_dir=None,
             self.loss = TorchLoss.from_pytorch(loss)
         if optimizer is None:
             from zoo.orca.learn.optimizers.schedule import Default
-            optimizer = SGD(learningrate_schedule=Default())
-        if isinstance(optimizer, TorchOptimizer):
-            optimizer = TorchOptim.from_pytorch(optimizer)
+            self.optimizer = SGD(learningrate_schedule=Default())
+        elif isinstance(optimizer, TorchOptimizer):
```
cyita, line 269: `elif -> if` or `self.optimizer = SGD(learningrate_schedule=Default())` to `self.optimizer = SGD(learningrate_schedule=Default()).get_optimizer()`

xiejinglei: Thanks! The mistake is fixed.

jason-dai:
I don't think we need `load_orca_model_or_checkpoint`; instead, we have two options:
1) Move `load_orca_model_or_checkpoint` logic to `load_latest_orca_checkpoint`
2) Or we can combine `load_orca_checkpoint` and `load_latest_orca_checkpoint` into a single method `load_orca_checkpoint`, with `version` default to None (i.e., loading latest checkpoint if it is None)

@yangw1234 which one makes more sense?

jason-dai:
Also need a unit test for the change

pyzoo/zoo/orca/learn/pytorch/estimator.py
```diff
@@ -451,16 +452,47 @@ def get_model(self):
...
+        :param model_path: path to the saved model.
         :return:
         """
-        raise NotImplementedError
 
-    def load(self, checkpoint, loss=None):
+        try:
+            pytorch_model = self.get_model().load_state_dict(torch.load(model_path))
+            self.model = TorchModel.from_pytorch(pytorch_model)
+            optim_path = self._get_optimizer_path(model_path)
```
jason-dai, line 490: We should support the case where there is no optimizer in the path. The user may have an existing PyTorch model (e.g., trained by PyTorch), and the we can simply load the model and set the optimizer to None.

xiejinglei:
Thanks for the comments. I will continue working on this on Monday

yangw1234:
> I don't think we need `load_orca_model_or_checkpoint`; instead, we have two options:
> 
> 1. Move `load_orca_model_or_checkpoint` logic to `load_latest_orca_checkpoint`
> 2. Or we can combine `load_orca_checkpoint` and `load_latest_orca_checkpoint` into a single method `load_orca_checkpoint`, with `version` default to None (i.e., loading latest checkpoint if it is None)
> 
> @yangw1234 which one makes more sense?

Hi @jason-dai , personally, I prefer option 2. I think a `load_orca_checkpoint` is enough and easy to understand.

jason-dai:
> > I don't think we need `load_orca_model_or_checkpoint`; instead, we have two options:
> > 
> > 1. Move `load_orca_model_or_checkpoint` logic to `load_latest_orca_checkpoint`
> > 2. Or we can combine `load_orca_checkpoint` and `load_latest_orca_checkpoint` into a single method `load_orca_checkpoint`, with `version` default to None (i.e., loading latest checkpoint if it is None)
> > 
> > @yangw1234 which one makes more sense?
> 
> Hi @jason-dai , personally, I prefer option 2. I think a `load_orca_checkpoint` is enough and easy to understand.

I'm OK with it

pyzoo/zoo/orca/learn/pytorch/estimator.py
xiejinglei: Since an optimizer is set for the estimator when it is initialized (if user does not specify optimizer, it will use SGD by default), is it possibly better to keep the optimizer unchanged rather than setting it to None in this case?

jason-dai: You may just pass `None` to the constructor, which will properly handle the case, yes?

pyzoo/zoo/orca/learn/pytorch/estimator.py
```diff
@@ -451,47 +452,90 @@ def get_model(self):
...
         :return:
         """
         import os
         from bigdl.nn.layer import Model
         from bigdl.optim.optimizer import OptimMethod
-        assert prefix is not None, "You should provide optimMethod prefix, " \
-                                   "for example 'optimMethod-TorchModelf53bddcc'"
+        from zoo.orca.learn.utils import find_latest_checkpoint
+        
+        if loss is not None:
```
yangw1234, line 527: talked with @cyita , I think we do not need the `loss` here. Just use the one when estimator created.

pyzoo/zoo/orca/learn/pytorch/estimator.py
```diff
@@ -500,14 +544,6 @@ def load_orca_checkpoint(self, path, version, prefix=None):
                              "and checkpoint type.")
         self.estimator = SparkEstimator(self.model, optimizer, self.model_dir)
 
-    def load_latest_orca_checkpoint(self, path):
```
yangw1234, line 503: let's remove this method from the super class. Also need to update `zoo.orca.learn.tf.estimator`, `zoo.orca.learn.bigd.estimator`, `zoo.orca.learn.openvino.estimator`

xiejinglei: OK I will have a look how to change these modules

pyzoo/zoo/orca/learn/pytorch/estimator.py
xiejinglei: OK I get your point. Thanks! I have changed the code accordingly

pyzoo/zoo/orca/learn/spark_estimator.py
```diff
@@ -85,11 +85,11 @@ def save(self, model_path):
         raise NotImplementedError
 
     @abstractmethod
-    def load(self, checkpoint):
+    def load(self, model_path):
```
jason-dai, line 88: also need to update `BaseEstimator.load` and `ray_estimator.Estimator.load`

xiejinglei: Have discussed with Yang, and we agree it's better to update them. The code is now fixed. Thanks!

xiejinglei:
PR validation passed: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/5215/
For `zoo.orca.learn.tf.estimator` and other related estimators, I will open a new PR to fix them.

yangw1234:
LGTM

Issue #3507, opened 2021-02-22
Missing DataFrame evaluating implment in Orca bigdl estimator 


leonardozcm:
See https://github.com/intel-analytics/analytics-zoo/blob/09a31de0c054428554a845a5bda9257ee49c5774/pyzoo/zoo/orca/learn/bigdl/estimator.py#L193

PR #3586, opened 2021-03-15
Implement nn_estimator evaluating method in Orca Bigdl estimator

leonardozcm:
Related issue: #3507
When call `estimator.evaluate` to deal with a dataFrame, it will call scala method `internalEval` to process data.

Usage:
```python
 result = est.evaluate(data=validationDF, batch_size=16, feature_cols="image")
```
As a result it returns a Jlist of EvaluatedResult, which will be converted to a python dict like:
```
2021-03-16 00:04:27 INFO  DistriOptimizer$:1759 - Top1Accuracy is Accuracy(correct: 8, count: 26, accuracy: 0.3076923076923077)
{'Top1Accuracy': 0.3076923191547394}
```
Working Progress:
- [x] Scala internalEval method
- [x] Python nn_estimator eval method
- [x] UT

And if there is any spark official evaluating trait or base class you know, pls leave your comments, thanks!

### DIFF
--- a/pyzoo/zoo/pipeline/nnframes/nn_classifier.py
+++ b/pyzoo/zoo/pipeline/nnframes/nn_classifier.py
@@ -25,7 +25,6 @@ from bigdl.util.common import *
 from zoo.feature.common import *
 from zoo import init_nncontext
 
-
 if sys.version >= '3':
     long = int
     unicode = str
@@ -384,6 +383,24 @@ class NNEstimator(JavaEstimator, HasFeaturesCol, HasLabelCol, HasPredictionCol, 
         """
         return self.validation_config
 
+    def setNNBatchSize(self, batch_size):
+        pythonBigDL_method_name = "setNNBatchSize"
+        callZooFunc(self.bigdl_type, pythonBigDL_method_name, self.value,
+                           batch_size)
+        return self
+
+    def setNNFeaturesCol(self, feature_cols):
+        pythonBigDL_method_name = "setNNFeaturesCol"
+        callZooFunc(self.bigdl_type, pythonBigDL_method_name, self.value,
+                           feature_cols)
+        return self
+
+    def setNNLabelCol(self, label_cols):
+        pythonBigDL_method_name = "setNNLabelCol"
+        callZooFunc(self.bigdl_type, pythonBigDL_method_name, self.value,
+                           label_cols)
+        return self
+
     def clearGradientClipping(self):
         """
         Clear clipping params, in this case, clipping will not be applied.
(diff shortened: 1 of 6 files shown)


zoo/src/main/scala/com/intel/analytics/zoo/pipeline/nnframes/NNEstimator.scala
```diff
@@ -477,6 +489,30 @@ class NNEstimator[T: ClassTag] private[zoo] (
     wrapBigDLModel(model)
   }
 
+  def internalEval(dataFrame: DataFrame): JList[EvaluatedResult] = {
+    val validationFeatureset = getDataSet(dataFrame, validationBatchSize)
```
yangw1234, line 493: I think you can directly call `model.evaluate`.

leonardozcm: synced offline

leonardozcm:
http://10.239.47.210:18888/job/ZOO-PR-Validation/5200/

pyzoo/zoo/pipeline/nnframes/nn_classifier.py
```diff
@@ -473,6 +490,12 @@ def setLabelCol(self, value):
         """
         return self._set(labelCol=value)
 
+    def eval(self, val_data):
```
yangw1234, line 493: `NNEstimator` is public API, I think we can fisrt move this method (and the other three set methods) to somewhere else and not expose them to the user of `NNEstimator`.

leonardozcm: have change to `_eval`

leonardozcm:
http://10.239.47.210:18888/job/ZOO-PR-Python-ExampleTests/2434/

pyzoo/zoo/pipeline/nnframes/nn_classifier.py
```diff
@@ -384,6 +383,24 @@ def getValidation(self):
         """
         return self.validation_config
 
+    def setNNBatchSize(self, batch_size):
```
yangw1234, line 386: also make these methods private

yangw1234: and add some comments documenting why you need these methods

leonardozcm: Sure

yangw1234:
let's first merge it after jenkins passed. We can look into whether we can use pyspark.ml.evaluator later.

leonardozcm:
http://10.239.47.210:18888/job/ZOO-PR-Validation/5225/

Issue #3554, opened 2021-03-08
When I use the cluster-serving to inference face-detection model by streaming, the result is wrong.


gtfaiwxm:
When I use the cluster-serving to inference face-detection model by streaming, the result is wrong. When I write one data , then time.sleep(3) , the result is right.

glorysdj:
Hi @gtfaiwxm, as we discussed, please paste more info about the model and the input/output. thanks.

gtfaiwxm:
Hi, @glorysdj , the face-detection-0100 model is downloaded from openvino-model zoo(2020.2 version) . The net outputs a blob with shape: [1, 1, N, 7], where N is the number of detected bounding boxes. For each detection, the description has the format: [image_id, label, conf, x_min, y_min, x_max, y_max], where:

- `image_id` - ID of the image in the batch
- `label` - predicted class ID
- `conf` - confidence for the predicted class
- (`x_min`, `y_min`) - coordinates of the top left bounding box corner
- (`x_max`, `y_max`) - coordinates of the bottom right bounding box corner.  When I keep writing an image data, the result :
{'image-4-test_1.jpg': array([[[14.        ,  1.        ,  0.02250869, ...,  0.        ,
          0.        ,  0.        ],
        [14.        ,  1.        ,  0.02248706, ...,  0.        ,
          0.        ,  0.        ],
        [14.        ,  1.        ,  0.022426  , ...,  0.        ,
          0.        ,  0.        ],
        ...,
        [15.        ,  1.        ,  0.02205388, ...,  0.        ,
          0.        ,  0.        ],
        [15.        ,  1.        ,  0.02199526, ...,  0.        ,
          0.        ,  0.        ],
        [15.        ,  1.        ,  0.02189598, ...,  0.        ,
          0.        ,  0.        ]]], dtype=float32), 'image-1-test_1.jpg': array([[[0.0000000e+00, 1.0000000e+00, 9.9770606e-01, ...,
         5.1740110e-03, 7.5382727e-01, 8.2223165e-01],
        [1.0000000e+00, 1.0000000e+00, 9.9770606e-01, ...,
         0.0000000e+00, 0.0000000e+00, 0.0000000e+00],
        [1.0000000e+00, 1.0000000e+00, 9.9604505e-01, ...,
         0.0000000e+00, 0.0000000e+00, 0.0000000e+00],
        ...,
        [1.0000000e+01, 1.0000000e+00, 9.5114928e-01, ...,
[...]

PR #3589, opened 2021-03-16
fix batch inference issue

KimiLiuQh:
fix the issue #3554 by adding an "batch_inference" entry in config.yaml, and set the thread of model to 1 if false and 4 if true when loading config in ClusterServingHelper.


gtfaiwxm:
Hi, Will this affect the throughput of the model? What's more, I think this is just a temporary solution, because I have run some models, and this one model has this problem.

Litchilitchy (changes requested):
This PR is better not to be merged for user-friendly concern.

Litchilitchy:
The related issue is fixed by #3690 , closed.

PR #3592, opened 2021-03-16
change path readthedocs to docs/readthedocs

helenlly:
1. change path readthedocs to docs/readthedocs
2. update config file with new path: .readthedocs.yml 

test demo :https://testhelendoc.readthedocs.io/en/latest/


jason-dai:
LGTM

PR #3601, opened 2021-03-17
add distributed cluster serving all in one scripts

glorysdj:
this pr is to add start and stop all ppml cluster serving components in one scripts with a config file to config environments.


qiyuangong (approved):
LTGM

PR #3609, opened 2021-03-18
use local images in quick start pages

helenlly:
before use remote image links in quick start pages, which may cause connection errors if no VPN.
now use local image instead( readthedocs/image)

demo: https://testhelendoc.readthedocs.io/en/latest/doc/Orca/QuickStart/orca-tf-quickstart.html
and 2 images shows correct in the page if without VPN


Le-Zheng (approved):

jason-dai:
LGTM

PR #3612, opened 2021-03-19
Cluster Serving contribution guide

Litchilitchy:


jason-dai:
How did we delete requests from input queue?

Litchilitchy:
> How did we delete requests from input queue?

serving scala code would check memory and delete records of input queue

`dequeue` and `query_and_delete()` or `query(delete=False)` could delete records of output queue

For customer feature, this could be controlled by contributor? And we could review later.

Litchilitchy:
> > How did we delete requests from input queue?
> 
> serving scala code would check memory and delete records of input queue
> 
> `dequeue` and `query_and_delete()` or `query(delete=False)` could delete records of output queue
> 
> For customer feature, this could be controlled by contributor in scala code? And we could review later.

PR #3619, opened 2021-03-22
Remove orca bigdl example resnet_finetune temporarily

leonardozcm:
For it is confused to import pytorch model in bigdl example, we may need to remove `example/orca/learn/bigdl/resnet_finetune` for now and follow up later.


leonardozcm:
http://10.239.47.210:18888/job/ZOO-PR-Validation/5217/

yangw1234:
LGTM

PR #3620, opened 2021-03-23
test bigdl 0.12.2-snapshot

Le-Zheng:


qiuxin2012:
https://github.com/intel-analytics/analytics-zoo/blob/864a6aaae29bb9eb665f5bc7557c8b653c538037/zoo/src/main/scala/com/intel/analytics/zoo/common/MKLBlas.scala#L22 change this vserf to bigdl core's

PR #3630, opened 2021-03-23
Load torch model from java value

### LINKED ISSUE #3606: Loss of functionality after `load_orca_checkpoint` in Orca PyTorch Estimator (bigdl backend)
In the function `load_orca_checkpoint` of `PyTorchSparkEstimator` ([link](https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/pytorch/estimator.py#L499)), there is a line
```py
self.model = Model.load(os.path.join(path, "model.{}".format(version)))
```
which loads the model as a `Layer` type (in module `bigdl.nn.layer`). However, the original `self.model` should be of `TorchModel` type ([link](https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/pipeline/api/torch/torch_model.py#L36)), a subclass of `Layer`. 

So after calling `load_orca_checkpoint`, the model will be replaced by a `Layer` type model, loosing the functionalities that `TorchModel` has. For example, the function `to_pytorch`, which converts the model to a pytorch model. As a result, the method `get_model` of `PyTorchSparkEstimator` cannot be used any longer.

In PR #3581 , I also add new save and load functionalities for `PyTorchSparkEstimator`. If this issue remains, the new `save` will not work after using `load_orca_checkpoint` as well.

qiuxin2012:
Load torch model from java value

fixed #3606

### DIFF
--- a/pyzoo/test/zoo/pipeline/api/torch/test_torch.py
+++ b/pyzoo/test/zoo/pipeline/api/torch/test_torch.py
@@ -225,7 +225,7 @@ class TestPytorch(TestCase):
             loaded_torchModel = TorchModel.from_value(loaded_model.value)
             dummy_input = torch.ones(16, 2)
             loaded_torchModel.forward(dummy_input.numpy())
-            loaded_torchModel.toPytorch()
+            loaded_torchModel.to_pytorch()
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


cyita (approved):

qiuxin2012:
jenkins passed

PR #3583, opened 2021-03-15
Add PyTorch distributed (Ray backend) Quickstart

GZHoffie:
Add notebook and markdown for the tutorial of `PytorchRayEstimator`.

PR #3632, opened 2021-03-23
Refine Orca Pytorch Quickstart

GZHoffie:
Adjusted detail in Orca Pytorch Quickstart markdown and notebook according to comments in #3583.


GZHoffie:
Jenkins passed here: http://10.239.47.210:18888/job/ZOO-PR-Quickstart-notebook/209/, readthedocs build was also successful.

hkvision (approved):
Thanks for the update.

jason-dai:
Tyr to run the PyTorch quickstart, and ran into several issues:

1) For some reason MNIST website gives HTTP 503 errors when running torchvision.datasets.MNIST; see https://stackoverflow.com/questions/66646604/http-error-503-service-unavailable-whan-trying-to-download-mnist-data for a solution

2) Run training and got an JEP error `jep.JepException: <class 'SystemError'>: unknown opcode`

3) Don't combine two versions of code into a single notebook for BigDL backend. We should keep the quickstart simple. Instead, keep the original BigDL backend quickstart unchanged, and add a new notebook for using Data Creator Function with BigDL backend.

@yangw1234 @GZHoffie

yangw1234:
> Tyr to run the PyTorch quickstart, and ran into several issues:
> 
> 1. For some reason MNIST website gives HTTP 503 errors when running torchvision.datasets.MNIST; see https://stackoverflow.com/questions/66646604/http-error-503-service-unavailable-whan-trying-to-download-mnist-data for a solution
> 2. Run training and got an JEP error `jep.JepException: <class 'SystemError'>: unknown opcode`
> 3. Don't combine two versions of code into a single notebook for BigDL backend. We should keep the quickstart simple. Instead, keep the original BigDL backend quickstart unchanged, and add a new notebook for using Data Creator Function with BigDL backend.
> 
> @yangw1234 @GZHoffie

The second error `jep.JepException: <class 'SystemError'>: unknown opcode` is probably becase colab upgrade the default python version to 3.7 and we install our packages in a conda env with python3.6.

yangw1234:
@GZHoffie @Le-Zheng maybe we can detect the hosted python version and installed the same version in conda?

Le-Zheng:
> @GZHoffie @Le-Zheng maybe we can detect the hosted python version and installed the same version in conda?

Ok. we need to get colab python version first. For example, we add `python_version = "{}.{}".format(sys.version_info.major, sys.version_info.minor)` , then make conda install the same python_version with host.

GZHoffie:
> > @GZHoffie @Le-Zheng maybe we can detect the hosted python version and installed the same version in conda?
> 
> Ok. we need to get colab python version first. For example, we add `python_version = "{}.{}".format(sys.version_info.major, sys.version_info.minor)` , then make conda install the same python_version with host.

Got it, thanks. I will open a separate PR for the change.

PR #3636, opened 2021-03-24
Create install-graphene-driver.sh

ManfeiBai:


ppml/install-graphene-driver.sh
```diff
@@ -0,0 +1,11 @@
+sudo apt install dkms
```
glorysdj, line 1: should start with #!/bin/bash
and please put 3 scripts here to a folder /ppml/scripts
and please test 3 scripts here

ManfeiBai: tks,done~

PR #3657, opened 2021-03-25
Add --upgrade option to --pre

hkvision:
Tested that `--upgrade --pre` would always install the latest nightly build whl.


hkvision:
jenkins passed: http://10.239.47.210:18888/job/ZOO-PR-Validation/5248/

jason-dai:
Can we use `-U` instead? @hkvision

hkvision:
> Can we use `-U` instead? @hkvision

Yes, -U and --upgrade are equivalent. I use --upgrade in companion with --pre (make it looks better at first glance)

PR #3674, opened 2021-03-29
Cluster Serving model parallelism control modification

Litchilitchy:


Litchilitchy:
some changes from #3480

PR #3679, opened 2021-03-29
Fix overwrite issue in save model

jenniew:
The over_write option was not applied when save to remote file

### DIFF
--- a/pyzoo/zoo/common/utils.py
+++ b/pyzoo/zoo/common/utils.py
@@ -83,7 +83,8 @@ def save_file(save_func, path, **kwargs):
 
         try:
             save_func(temp_path, **kwargs)
-            put_local_file_to_remote(temp_path, path)
+            if "over_write" in kwargs:
+                put_local_file_to_remote(temp_path, path, over_write=kwargs['over_write'])
         finally:
             os.remove(temp_path)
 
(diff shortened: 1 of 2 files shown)


jenniew:
Jenkins pass: http://10.239.47.210:18888/job/ZOO-PR-Validation/5277/

pyzoo/zoo/common/utils.py
```diff
@@ -83,7 +83,8 @@ def save_file(save_func, path, **kwargs):
 
         try:
             save_func(temp_path, **kwargs)
-            put_local_file_to_remote(temp_path, path)
+            if "over_write" in kwargs:
```
jason-dai, line 86: 1) Add an argument `overwrite=false` for the method?
2) need to check `overwrite` even for local path?
3) what if `overwrite` is false, but file does not exist (e.g., for remote)?

jenniew: Updated:
```
            if "overwrite" in kwargs:
                put_local_file_to_remote(temp_path, path, over_write=kwargs['overwrite'])
            else:
                put_local_file_to_remote(temp_path, path)
```
For local path,  just call save_func(path, **kwargs), no need to check overwrite.
If overwrite is false,  it will create file on remote

jenniew:
Jenkins: http://10.239.47.210:18888/job/ZOO-PR-Validation/5283/
Test on local and remote

PR #3681, opened 2021-03-30
Added check status script for local cluster serving

ColossusChang:
- Added a script `check-status.sh` for troubleshooting local cluster serving.
- Added a part in the Trusted Cluster Serving README file on how to use the new script.
- Added commands in `start-all.sh` to check the status of each component after starting them.
- The edits on `generate-keys.sh` actually result from an older commit which did not get merged. The reason for editing this is that creating a new file using redirection may encounter a "permission denied" error (at least when I last tried).


ppml/trusted-cluster-serving/scala/docker-graphene/README.md
```diff
@@ -39,6 +39,17 @@ Then run the example with docker: <br>
...
 ```
+##### Troubleshooting
+You can run the script `check-status.sh` in the docker container to check whether the components have been correctly started. 
+To test a specific component, pass one or more argument to it among the following:
+"redis", "flinkjm", "flinktm", "frontend", and "cluster". To test all components, you can either pass no argument or pass the "all" argument.
+
+It is suggested to run this script once after starting local cluster serving to verify that all components are up and running.
+```bash
+docker exec -it flink-local bash check-status.sh
+```
```
glorysdj, line 50: please show the example result of the check all
and maybe also give a example of check flinkjm here.

glorysdj: please also mention this check status in the distributed mode

ppml/trusted-cluster-serving/scala/docker-graphene/README.md
```diff
@@ -39,6 +39,39 @@ Then run the example with docker: <br>
 ```bash
 ./start-local-cluster-serving.sh
 ```
+##### Troubleshooting
+You can run the script `check-status.sh` in the docker container to check whether the components have been correctly started. 
+Note that this only works for local cluster serving (for now).
+
+To test a specific component, pass one or more argument to it among the following:
+"redis", "flinkjm", "flinktm", "frontend", and "cluster". For example, run the following command to check the status of the Flink job master.
```
glorysdj, line 47: better to change cluster to serving

glorysdj (approved):
LGTM

PR #3684, opened 2021-03-30
Fix link in RayOnSpark user guide

hkvision:
The directory structure changed and makes the link invalid.


hkvision:
https://hkvision-ray-link.readthedocs.io/en/latest/doc/Ray/Overview/ray.html

PR #3690, opened 2021-03-30
remove add dim of input for openvino for batch infer

KimiLiuQh:
remove the adding dim operation before predict to support batch inference of openvino model
add openvino face detection 0100 model to openvino model specific test
add batch inference for correctness unit test


KimiLiuQh:
jenkis pr validation passed with ```Build #5291 (Mar 30, 2021 4:50:41 PM)```

Litchilitchy (approved):

PR #3541, opened 2021-03-03
modify to add http service guide

KimiLiuQh:
add the python version of http service example in cluster serving, and way to parse requested json data.

PR #3695, opened 2021-03-31
Cluster Serving http return add quote for py json parse

Litchilitchy:


Litchilitchy:
Http response could directly parse to ndarray now. 

Http json response is also validated by json validator online.

Also added example notebook.

Litchilitchy:
modification of #3541 has finished in this PR

PR #3710, opened 2021-04-01
Added distributed-check-status for distributed cluster serving

ColossusChang:
- Added a distributed-check-status.sh to check the status of the components on distributed cluster serving.
- Fixed the paths to check-status.sh in the doc and in start-all.sh
- Added loop breakout condition for start-distributed-cluster-serving.sh


glorysdj (approved):
LGTM

PR #3725, opened 2021-04-05
dien preprocessing and feature table

songhappy:


pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,82 @@
...
+        .dropna(subset=['asin', 'categories']) \
+        .selectExpr("*", "get_category(categories) as category") \
+        .withColumnRenamed("asin", "item").drop("categories").persist(storageLevel=StorageLevel.DISK_ONLY)
+
+    full_df = review_df.join(meta_df, on="item", how="left").fillna("default", ["item"])
+    item_size = full_df.select("item").distinct().count()
+
+    full_df.persist(StorageLevel.DISK_ONLY)
+    full_tbl = FeatureTable.from_spark_df(full_df)
+    user_index = full_tbl.gen_string2ind('user')
```
jason-dai, line 58: you should use `FeatureTable.gen_string_idx`

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -115,9 +140,147 @@ def gen_string_idx(self, columns, freq_limit):
...
+        df = df.withColumn("history", explode(col("history"))) \
+                .drop("his_collect")
+        for c in cols:
+            df = df.withColumn(c, col("history." + c)) \
+                    .withColumn(c + "_history", col("history." + c + '_history'))
+        df = df.drop("history")
+        return FeatureTable(df)
+
+    def gen_neg_hist_seq(self, item_size, item2cat, item_history_col='item_history', neg_num=5):
+        sc = SparkContext.getOrCreate()
```
cyita, line 211: Use OrcaContext.get_spark_context() instead?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -36,6 +39,22 @@ def _read_parquet(paths):
         df = spark.read.parquet(*paths)
         return df
 
+    @staticmethod
+    def _read_json(sc, paths, cols):
+        if sc:
```
cyita, line 44: Do we need sc here?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -115,9 +140,147 @@ def gen_string_idx(self, columns, freq_limit):
...
+                if isinstance(seq[0], list):
+                    return seq + [[0] * len(seq[0]) for i in range(seq_len - length)]
+                else:
+                    return seq + [0] * (seq_len - length)
+            else:
+                return seq[:seq_len]
+
+        df = self.df
+        for c in padding_cols:
+            col_type = df.schema[c].dataType
```
cyita, line 262: Do we need to check the col_type if it isn't supported?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -115,9 +140,147 @@ def gen_string_idx(self, columns, freq_limit):
...
+            pad_udf = udf(pad, col_type)
+            df = df.withColumn(c, pad_udf(col(c)))
+        return FeatureTable(df)
+
+    def mask(self, mask_cols=None, seq_len=100):
+        def add_mask(seq):
```
cyita, line 268: Check the type of `seq`?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -115,9 +140,45 @@ def gen_string_idx(self, columns, freq_limit):
...
 
+    def gen_ind2ind(self, cols, indices):
+        df = self.encode_string(cols, indices).df.select(*cols).distinct()
+        return FeatureTable(df)
+
     def _clone(self, df):
         return FeatureTable(df)
 
+    def gen_negtive_samples(self, item_size, item_col="item", label_col="label", neg_num=1):
+        df = callZooFunc("float", "genNegSamples", self.df, item_size, item_col, label_col, neg_num)
```
jenniew, line 151: typo for negtive? change to 'gen_negative_samples'?

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,99 @@
...
+    get_label = udf(lambda x: [float(x), 1 - float(x)], ArrayType(FloatType()))
+    reset_cat = udf(lambda item: item2cat_map[item], returnType=IntegerType())
+
+    full_tbl = full_tbl\
+        .encode_string(['user', 'item', 'category'], [indices[0], indices[1], indices[2]]) \
+        .gen_hist_seq(user_col="user", cols=['item', 'category'],
+                      sort_col='time', min_len=1, max_len=100)\
+        .gen_length("item_history")\
+        .gen_negative_samples(item_size, item_col='item', neg_num=1) \
+        .transform_python_udf("item", "category", reset_cat) \
```
jason-dai, line 87: why do we need this step? if it is absolutely needed, you should apply it on `meta_df` using spark dataframe, something like:

```python
 meta_df = spark.read.json(options.meta_file).select(['asin', 'categories'])\
        ...\
        .withColumn("category", reset_cat(col("category")))
        .persist(storageLevel=StorageLevel.DISK_ONLY)
```

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,99 @@
...
+        return cat.strip().lower()
+    spark.udf.register("get_category", get_category, StringType())
+    meta_df = spark.read.json(options.meta_file).select(['asin', 'categories'])\
+        .dropna(subset=['asin', 'categories']) \
+        .selectExpr("*", "get_category(categories) as category") \
+        .withColumnRenamed("asin", "item").drop("categories").distinct()\
+        .persist(storageLevel=StorageLevel.DISK_ONLY)
+    print("meta_data, ", meta_df.count())
+
+    full_df = review_df.join(meta_df, on="item", how="left").fillna("default", ["category"]) \
```
jason-dai, line 64: We should support `join` operation on Tables, something like:

```python
item_tbl = FeatureTable.from_spark_df(item_df)
meta_tbl = FeatureTable.from_spark_df(meta_df)
full_tbl = item_tbl .join(meta_tbl, ...)
```

songhappy: added join in FeatureTable

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
songhappy: reset_cat intends to  reset category for negative items. We could do it right in two ways.
1. encode string of users items in review table,  encode category in meta table, add negative items in review table, join with meta table, then generate history sequence and other columns
2. join first, then encode users items categories, add negative sample in full table, then correct categories for negative samples. 
The down side of first approach is it shuffles 2 times review data when join, which could cause problems.
The down side of second approach is it needs to fix category for negative samples.
From performance perspective, I chose the second approach, if it is confusing, I can switch to the first approach

pyzoo/zoo/friesian/feature/table.py
songhappy: done

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,80 @@
...
+    item2cat.write_parquet(options.output+"item2cat")
+
+    get_length = udf(lambda x: len(x), IntegerType())
+    get_label = udf(lambda x: [float(x), 1 - float(x)], ArrayType(FloatType()))
+
+    full_tbl = full_tbl\
+        .encode_string(['user', 'item', 'category'], [indices[0], indices[1], indices[2]]) \
+        .gen_hist_seq(user_col="user", cols=['item', 'category'], sort_col='time', min_len=1, max_len=100)\
+        .gen_length("item_history")\
+        .gen_negtive_samples(item_size, item_col='item', neg_num=1) \
```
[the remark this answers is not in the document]
songhappy: done

pyzoo/zoo/friesian/feature/table.py
songhappy: no, rewrote

jason-dai:
I think we would like something like:

```python
txn_tbl = FeatureTable.from_spark_df(...) #(user, item, time)
feat_tbl = FeatureTable.from_spark_df(...) #(item, category)

txn_tbl.add_hist_seq(col="item", time_col="time") \
      .add_neg_samples(col="item", label_col="label")\
      .add_neg_hist_seq(col="item_hist_seq")\
      .add_feature(cols=["item", "item_hist_seq", "neg_item_his_seq"], feat_tbl, feat_id="item" )
```

songhappy:
All tests passed except one about AUC, which has nothing to do with this PR.  
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/5464/

jenniew:
LGTM.

songhappy:
jekins about this PR all passed, failed at some other isseue
http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/5493/

jason-dai:
It is unclear to me what the right semantics should be if an `item` has no associated  `category`
- All the items come from `item_tbl`, which should all have associated categories
- Unless there is an item in `transaction_tbl` not in `item_tbl`? I'm not sure how you want to handle this case though.

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+        .distinct()
+
+    full_tbl = transaction_tbl\
+        .encode_string(['user', 'item'], [user_index[0], item_category_indices[0]])\
+        .add_hist_seq(user_col="user", cols=['item'],
+                      sort_col='time', min_len=1, max_len=100)\
+        .add_neg_hist_seq(item_size, 'item_hist_seq', neg_num=5) \
+        .add_negative_samples(item_size, item_col='item', neg_num=1)\
+        .join(item2cat, "item")\
+        .add_feature(["item_hist_seq", "neg_item_hist_seq"], item_tbl, default_cat)\
```
jason-dai, line 88: is `item_tbl` correct? should it be `item2cat`?

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+        .mask_pad(mask_cols=['item_hist_seq'],
+                  padding_cols=['item_hist_seq', 'category_hist_seq',
+                                'neg_item_hist_seq', 'neg_category_hist_seq'],
+                  seq_len=100) \
+        .add_length("item_hist_seq") \
```
jason-dai, line 93: `item_hist_seq` has been pad to 100, no?

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+        .persist(storageLevel=StorageLevel.DISK_ONLY)
+    item_tbl = FeatureTable(item_df)
+
+    print("item_tbl, ", item_tbl.count())
+
+    item_category_indices = item_tbl.gen_string_idx(["item", "category"], 1)
+    cat_default = item_category_indices[1].df.filter("category == 'default'").collect()
+    default_cat = cat_default[0][1] if cat_default else item_category_indices[1].count()
+    new_row = spark.createDataFrame([("default", int(default_cat))], ["category", "id"])
+    category_index = StringIndex(item_category_indices[1].df.union(new_row).distinct()
```
jason-dai, line 71: do we really need this?

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+        .encode_string(["item", "category"], [item_category_indices[0], category_index])\
+        .distinct()
+
+    full_tbl = transaction_tbl\
+        .encode_string(['user', 'item'], [user_index[0], item_category_indices[0]])\
+        .add_hist_seq(user_col="user", cols=['item'],
+                      sort_col='time', min_len=1, max_len=100)\
+        .add_neg_hist_seq(item_size, 'item_hist_seq', neg_num=5) \
+        .add_negative_samples(item_size, item_col='item', neg_num=1)\
+        .join(item2cat, "item")\
```
jason-dai, line 87: we can simply combine this with `add_feature` below: 
```
add_feature(["item", "item_hist_seq", "neg_item_hist_seq"], item2cat, default_cat)
```

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -204,7 +244,8 @@ def encode_string(self, columns, indices):
             col_name = columns[i]
             index_tbl.broadcast()
             data_df = data_df.join(index_tbl.df, col_name, how="left") \
-                .drop(col_name).withColumnRenamed("id", col_name)
+                .drop(col_name).withColumnRenamed("id", col_name)\
+                .dropna(subset=[col_name])
```
jason-dai, line 248: don't `dropna` here; the user can choose to `fillna` or `dropna` in the application code after encoding strings

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
                                    zip(df_id_list, columns)))
         return string_idx_list
 
+    def gen_ind2ind(self, cols, indices):
```
jason-dai, line 278: Is this needed?

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+      } else {
+        s"post_pad_matrix($c) as $c"
+      }
+    }).mkString(",")
+    val leftCols = df.columns.filter(x => !colNames.contains(x)).mkString(",")
+
+    df.sqlContext.sql(s"select $leftCols, $selectStatement from tmp")
+  }
+
+  def addLength(df: DataFrame, colName: String): DataFrame = {
```
jason-dai, line 349: there is no need to do this in scala; just implement it using python

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+        .withColumnRenamed('unixReviewTime', 'time')\
+        .dropna("any").persist(storageLevel=StorageLevel.DISK_ONLY)
+    transaction_tbl = FeatureTable(transaction_df)
+    print("review_tbl, ", transaction_tbl.count())
+
+    # read meta data
+    def get_category(x):
+        cat = x[0][-1] if x[0][-1] is not None else "default"
+        return cat.strip().lower()
+    spark.udf.register("get_category", get_category, StringType())
```
jason-dai, line 57: `get_cat_udf = udf(lambda x: get_category(x), StringType())`

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+
+    full_tbl = transaction_tbl\
+        .encode_string(['user', 'item'], [user_index[0], item_category_indices[0]])\
+        .add_hist_seq(user_col="user", cols=['item'],
+                      sort_col='time', min_len=1, max_len=100)\
+        .add_neg_hist_seq(item_size, 'item_hist_seq', neg_num=5) \
+        .add_negative_samples(item_size, item_col='item', neg_num=1)\
+        .join(item2cat, "item")\
+        .add_feature(["item_hist_seq", "neg_item_hist_seq"], item_tbl, default_cat)\
+        .mask_pad(mask_cols=['item_hist_seq'],
```
jason-dai, line 89: I think `mask` should be a part of the `pad` operation: it is meaningless if you do `mask` but not `pad`, yes?

Just change it to `pad(cols=[...], length=..., mask_cols=[...])`

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
jason-dai, line 88: why do you need `default_cat` here?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+
+         :param padding_cols: list of string, columns need to be padded with 0s.
+         :param seq_len:  int, length of padded column
+
+         :return: FeatureTable
+         """
+        df = callZooFunc("float", "postPad", self.df, padding_cols, seq_len)
+        return FeatureTable(df)
+
+    def mask(self, mask_cols, seq_len=100):
```
jason-dai, line 349: I don't think we need a standalone `mask` operation; just combine it with `pad`

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+         :param item_history_col:  string, this column should be a list of visits in history
+         :param neg_num:  int, for each positive record, add neg_num of negative samples
+
+         :return: FeatureTable
+         """
+
+        df = callZooFunc("float", "addNegHisSeq", self.df, item_size, item_history_col, neg_num)
+        return FeatureTable(df)
+
+    def pad(self, padding_cols, seq_len=100):
```
jason-dai, line 337: change to `pad(self, cols, length, mask_cols=None)`

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+         Generagte length of a colum
+
+         :param col_name: string.
+
+         :return: FeatureTable
+         """
+        df = callZooFunc("float", "addLength", self.df, col_name)
+        return FeatureTable(df)
+
+    def mask_pad(self, padding_cols, mask_cols, seq_len=100):
```
jason-dai, line 372: don't need this

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
     }
     resultDF
   }
+
+  def addHistSeq(df: DataFrame,
+                 userCol: String,
+                 colNamesin: JList[String],
+                 sortCol: String,
+                 minLength: Int,
+                 maxLength: Int): DataFrame = {
```
jason-dai, line 192: 1) Change the definition to `addHistSeq(df, cols, userCol, timeCol, minLenth, maxLength)`

2) We shouldn't assume that `cols` (or `colNamesin`) are all Integer in the implementation

3) We shouldn't only keep `userCol`, `cols` (or `colNamesin`) and their `hist_seq` in the final output (throwing away `timeCol` and other existing columns).

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+    df.sparkSession.conf.set("spark.sql.legacy.allowUntypedScalaUDF", "true")
+    val colNames: Array[String] = colNamesin.asScala.toArray
+    val schema = ArrayType(StructType(colNames.flatMap(c =>
+      Seq(StructField(c, IntegerType), StructField(c + "_history", ArrayType(IntegerType))))))
+
+    val genHisUDF = udf(f = (his_collect: Seq[Row]) => {
+      val full_rows = his_collect.sortBy(x => x.getAs[Long](sortCol)).toArray
+      val n = full_rows.length
+      val result: Seq[Row] = (0 to n - 1).map(i => {
```
jason-dai, line 201: instead of `(0 to n - 1).map`, use `(minLength-1 to n-1).map`, and then there is no need to filter by `minLength` later

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+          col("history." + c + "_history").as(c + "_hist_seq")))
+    val collectColumns = colNames.map(c => col(c)) ++ Seq(col(sortCol))
+    val filterCondition = colNames.map(c =>
+      "size(" + c +  s"_hist_seq) >= $minLength").mkString(" and ")
+
+    df.groupBy(userCol)
+      .agg(collect_list(struct(collectColumns: _*)).as("his_collect"))
+      .filter(size(col("his_collect")) >= 1)
+      .withColumn("history", genHisUDF(col("his_collect")))
+      .withColumn("history", explode(col("history")))
```
jason-dai, line 226: what if the input dataframe contains column `his_collect` or  `history`?

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+      .drop("his_collect")
+      .select(selectColumns: _*)
+      .filter(filterCondition)
+  }
```
jason-dai, line 230: The implementation is a bit too complex and not easy to understand. I think it can be simplified as follows:

```scala
    // Assume the schema for df is userCol, timeCol, col1, ..., coln, col1', ... coln'
    // We need to generate hist_seq for col1, ..., coln, but not col1', ..., coln'
    // genHistwill return a new field "friesian_history" whose schema is:
    // Array[Struct<userCol, timeCol, col1, ..., coln, col1', ..., coln', col1_hist_seq, ..., coln_hist_seq>]
    val genHist = udf(f = (his_collect: Seq[Row]) => {
      val full_rows = his_collect.sortBy(x => x.getAs[Long](sortCol)).toArray
      val n = full_rows.length
      val result: Seq[Row] = (minLength - 1 to n - 1).map(i => {
        val lowerBound = ...
        val row = full_rows(i)
        val seq = colIndex.map(c =>         //colIndex is the position index for colNames
          full_rows.slice(lowerBound, i).map(r => r.get(c)))
        Row.fromSeq(row.toSeq ++ seq)
      })
      result
    }, schema)

df.groupBy(userCol)
    .agg(collect_list(struct(cols: _*)).as("friesien_hist_collect"))
      .withColumn("friesian_history", explode(genHist(col("friesien_hist_collect"))))
      .select(userCol, "friesian_history.*")
```

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+    }
+
+    df.createOrReplaceTempView("tmp")
+    df.sqlContext.udf.register("mask", maskUdf)
+    val selectStatement = colNames.toArray().map(c => s"mask($c) as $c" + "_mask").mkString(",")
+
+    df.sqlContext.sql(s"select *, $selectStatement from tmp")
+  }
+
+  def addNegHisSeq(df: DataFrame, itemSize: Int,
```
jason-dai, line 249: addNegHistSeq

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+        }
+      }
+      val result = Row.fromSeq(row.toSeq ++ Array[Any](negItemSeq))
+      result
+    })
+
+    val newSchema = StructType(df.schema.fields ++ Array(
+      StructField("neg_item_hist_seq", ArrayType(ArrayType(IntegerType)))))
+
+   sqlContext.createDataFrame(combinedRDD, newSchema)
```
jason-dai, line 274: I think you may use an UDF instead of `rdd.map`:

df.withColumn("negative_"+historyCol, negSample(col(historyCol)))

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+        var neg = 0
+        do {
+          neg = r.nextInt(itemSize)
+        } while (neg == itemindex)
+        neg
+      }).map(x => (x, 0)) ++ Seq((itemindex, 1)))
+
+    val columns = df.columns.filter(x => x != itemCol).mkString(",")
+
+    val negativedf = df.withColumn("negative", negativeUdf(col(itemCol)))
```
jason-dai, line 297: df.withColumn("item_label", negativeUdf(col(itemCol)))

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+      .withColumn("negative", explode(col("negative")))
```
jason-dai, line 298: Make `negativeUdf` return a `Struct<item, label>`, and then:
```scala
df.withColumn("item_label", explode(negativeUdf(col(itemCol))))
   .select(columns, "item_label.*")
```

zoo/src/main/scala/com/intel/analytics/zoo/friesian/python/PythonFriesian.scala
```diff
@@ -174,10 +177,177 @@ class PythonFriesian[T: ClassTag](implicit ev: TensorNumeric[T]) extends PythonZ
...
+
+    negativedf.createOrReplaceTempView("tmp")
+
+    df.sqlContext
+      .sql(s"select $columns , negative._1 as item, negative._2 as $labelCol from tmp")
+  }
+
+  def postPad(df: DataFrame, colNames: JList[String], maxLength: Int = 100): DataFrame = {
```
jason-dai, line 306: combine pad and mask: `postPad(df, cols, maxLength, maskCols=None)`

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+         :param on: string, join on this column
+         :param how: string
+
+         :return: FeatureTable
+         """
+        assert isinstance(table, Table), "the joined table should be a Table"
+        joined_df = self.df.join(table.df, on=on, how=how)
+        return FeatureTable(joined_df)
+
+    def add_feature(self, item_cols, feature_tbl, default_value):
```
jason-dai, line 412: We should not assume that `feature_tbl` only has two Int columns; in general, we should assume that `feature_tbl` is a map form item to its feature list.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+        """
+         Get the category or other field from another map like FeatureTable
```
jason-dai, line 414: there is no such as concept of `category` in the library

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+
+         :param item_cols: list[string]
+         :param feature_tbl: FeatureTable with two columns [category, item]
+         :param defalut_cat_index: default value for category if key does not exist
+
+         :return: FeatureTable
+         """
+        item2cat_map = dict(feature_tbl.df.distinct().rdd.map(lambda row: (row[0], row[1]))
```
jason-dai, line 422: `item2cat_map` is incorrect in concept; it would be `item_feature_map`, and we should use broadcast here.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+        assert isinstance(table, Table), "the joined table should be a Table"
+        joined_df = self.df.join(table.df, on=on, how=how)
+        return FeatureTable(joined_df)
+
+    def add_feature(self, item_cols, feature_tbl, default_value):
+        """
+         Get the category or other field from another map like FeatureTable
+
+         :param item_cols: list[string]
+         :param feature_tbl: FeatureTable with two columns [category, item]
```
jason-dai, line 417: this comment is incorrect

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+         :param defalut_cat_index: default value for category if key does not exist
+
+         :return: FeatureTable
+         """
+        item2cat_map = dict(feature_tbl.df.distinct().rdd.map(lambda row: (row[0], row[1]))
+                            .collect())
+
+        def gen_cat(items):
+            getcat = lambda item: item2cat_map.get(item, default_value)
```
jason-dai, line 426: I don't think we need `default_value` here

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+                    cats.append(line_cats)
+            else:
+                raise ValueError('only int, list[int], and list[list[int]] are supported.')
+            return cats
+
+        df = self.df
+        for c in item_cols:
+            col_type = df.schema[c].dataType
+            cat_udf = udf(gen_cat, col_type)
+            df = df.withColumn(c.replace("item", "category"), cat_udf(col(c)))
```
jason-dai, line 445: this is wrong; you cannot assume there is "item" in the column name, and shouldn't have a concept of "category" in the library.

jason-dai:
See the comments above; and please also add Python doc and Scala doc @songhappy

pyzoo/zoo/examples/friesian/feature/dien_preprocessing.py
songhappy: sometimes, no need to have mask when padding; sometimes, we need add mask when padding. removed mask method

jason-dai: > sometimes, no need to have mask when padding; sometimes, we need add mask when padding. removed mask method

You can make `mask_cols` default to None

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -234,9 +275,176 @@ def gen_string_idx(self, columns, freq_limit):
...
+         :param padding_cols: list of string, columns need to be padded with 0s.
+         :param mask_cols: list of string, columns need to be masked with 1s and 0s.
+         :param seq_len:  int, length of masked column
+
+         :return: FeatureTable
+         """
+        table = self.mask(mask_cols, seq_len)
+        return table.pad(padding_cols, seq_len)
+
+    def transform_python_udf(self, in_col, out_col, udf_func):
```
jason-dai, line 385: change to `apply(in_col, out_col=None, func)`, where `func` is just a simple python function (not a Spark UDF function); you can construct `udf_func` based on `func`

PR #3731, opened 2021-04-07
Add initial version of auto estimator

yushan111:
This PR implemented an initial version of `orca.automl.AutoEstimator`.

It includes the new APIs: `AutoEstimator.from_torch`,  `AutoEstimator.from_keras`, `auto_estimator.fit`, `auto_estimator.get_best_model`  

And UTs for `auto_pytorch_estimator` and `auto_keras_estimator` as examples.


pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -0,0 +1,70 @@
...
+
+class AutoEstimator:
+    def __init__(self, estimator):
+        self.estimator = estimator
+
+    @staticmethod
+    def from_torch(*,
+                   model_creator,
+                   optimizer_creator,
+                   loss_creator,
```
jason-dai, line 25: can we also support `optimizer_creator` and `loss_creator` with string values?

yushan111: I have changed names to `optimizer` and `loss`. And they have supported string values.

pyzoo/zoo/automl/model/base_pytorch_model.py
```diff
@@ -40,15 +40,28 @@ def __init__(self, model_creator, optimizer_creator, loss_creator,
         self.onnx_model = None
         self.onnx_model_built = False
 
+    def _create_loss(self):
+        if isinstance(self.loss_creator, torch.nn.modules.loss._Loss):
```
jason-dai, line 44: is it an instance or class?

pyzoo/zoo/automl/model/base_pytorch_model.py
```diff
@@ -40,15 +40,28 @@ def __init__(self, model_creator, optimizer_creator, loss_creator,
...
+            self.criterion = self.loss_creator
+        else:
+            self.criterion = self.loss_creator(self.config)
+
+    def _create_optimizer(self):
+        if issubclass(self.optimizer_creator, torch.optim.Optimizer):
```
jason-dai, line 50: is it an instance or class?

pyzoo/zoo/automl/model/base_pytorch_model.py
```diff
@@ -40,15 +40,28 @@ def __init__(self, model_creator, optimizer_creator, loss_creator,
...
+            # use torch default parameter values if user pass optimizer name or optimizer class.
+            self.optimizer = self.optimizer_creator(self.model.parameters())
```
jason-dai, line 52: the user may want to tune learning rate even if he or she specifies, say, "Adam"?

pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -0,0 +1,105 @@
...
+            data,
+            recipe=None,
+            metric=None,
+            search_alg=None,
+            search_alg_params=None,
+            scheduler=None,
+            scheduler_params=None,
+            ):
+
+        self.searcher.compile(data=data,
```
jason-dai, line 93: can `fit` be invoked multiple times? maybe we should create the `searcher` in the `fit` method?

pyzoo/zoo/automl/model/base_pytorch_model.py
yushan111: It could be an instance, since we may expect user to pass `loss=nn.BCELoss()` as orca estimator does. ([orca UT](https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/test/zoo/orca/learn/ray/pytorch/test_estimator_pytorch_backend.py#L147))

pyzoo/zoo/automl/model/base_pytorch_model.py
yushan111: It may need to be a class, since a torch optimizer is instantiated as below (the parameters need to be passed during instantiation while the parameters for loss is passed after instantiation).
`optimizer = optim.SGD(model.parameters(), lr=0.01, momentum=0.9)`

pyzoo/zoo/orca/automl/auto_estimator.py
yushan111: I think it might be ok if user invoke `fit` multiple times, since the model is created in `self.searcher.compile`

pyzoo/zoo/automl/model/base_pytorch_model.py
yushan111: Sure, I did so because the learning rate name that user should pass in recipe may be a little hard to unify. 
Shall we use a constant such as `LR_NAME` to define the learning rate name user should use in his/her recipe or we specify the learning rate name like `lr` in document

yushan111: lr has been added in [`_create_optimizer`](https://github.com/intel-analytics/analytics-zoo/pull/3731/files#diff-267f3b46c5536ed684ade4f853e6d88f990389ad549fb7517a2371b4fbcce2e6R58-R59) 
And the corresponding [document](https://github.com/intel-analytics/analytics-zoo/pull/3731/files#diff-5fc158591048ac9fcb1536eb5aaabd4b5176f872c537953816c919e8367fb0e2R35-R39)

pyzoo/zoo/orca/automl/auto_estimator.py
yushan111: I am not sure about the expected behavior to invoke `auto_estimator.fit` multiple times

pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -94,7 +95,8 @@ def fit(self,
             scheduler=None,
             scheduler_params=None,
             ):
-
+        if self._fitted:
+            raise RuntimeError("auto_estimator.fit is not expected to be called multiple times.")
```
jason-dai, line 99: This AutoEstimator has already been fitted and cannot fit again.

yushan111:
Jenkins passed: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/5371/

pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -0,0 +1,115 @@
...
+        :param name: Name of the auto estimator.
+        :return: an AutoEstimator object.
+        """
+        from zoo.orca.automl.pytorch_utils import validate_pytorch_loss, \
+            validate_pytorch_optim
+        from zoo.automl.model import ModelBuilder
+        from zoo.automl.search import SearchEngineFactory
+        loss = validate_pytorch_loss(loss)
+        optimizer = validate_pytorch_optim(optimizer)
+        model_builder = ModelBuilder.from_pytorch(model_creator=model_creator,
```
shane-huang, line 55: Shall we test the behavior if user passes a model_creator that creates a non-pytorch model by mistake? It seems ModelBuilder or PyTorchBaseModel does not check whether it is pytorch model or not.

pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -0,0 +1,115 @@
...
+                                                  optimizer_creator=optimizer,
+                                                  loss_creator=loss)
+        searcher = SearchEngineFactory.create_engine(backend="ray",
+                                                     logs_dir=logs_dir,
+                                                     resources_per_trial=resources_per_trial,
+                                                     name=name)
+        return AutoEstimator(model_builder=model_builder, searcher=searcher)
+
+    @staticmethod
+    def from_keras(*,
```
shane-huang, line 65: shall we use the same name "from_tfkeras" as in ModelBuilder class?

pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -0,0 +1,115 @@
...
+
+        :param model_creator: Tensorflow keras model creator function.
+        :param logs_dir: Local directory to save logs and results.
+        :param resources_per_trial: Dict. resources for each trial. e.g. {"cpu": 2}.
+        :param name: Name of the auto estimator.
+        :return: an AutoEstimator object.
+        """
+        from zoo.automl.model import ModelBuilder
+        from zoo.automl.search import SearchEngineFactory
+        model_builder = ModelBuilder.from_tfkeras(model_creator=model_creator)
```
shane-huang, line 82: * Here we assume ```model_creator``` creates a model and call ```model.compile```. What if user does not call compile? Maybe we need to check that.

pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -0,0 +1,115 @@
...
+        searcher = SearchEngineFactory.create_engine(backend="ray",
+                                                     logs_dir=logs_dir,
+                                                     resources_per_trial=resources_per_trial,
+                                                     name=name)
+        return AutoEstimator(model_builder=model_builder, searcher=searcher)
+
+    def fit(self,
+            data,
+            recipe=None,
+            metric=None,
```
shane-huang, line 92: What if this ```metric``` paramter does not align with the metric that is passed in model.compile when user defines the model_creator?

pyzoo/zoo/orca/automl/auto_estimator.py
yushan111: This is to align the usage in orca tf2 estimator https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/tf2/estimator.py#L36

And I did some refactor in ModelBuilder and removed the factory method (`from_tfkeras` and `from torch`) in PR #3785

pyzoo/zoo/orca/automl/auto_estimator.py
yushan111: Sure, I will add check in `PytorchBaseModel` in a new PR

pyzoo/zoo/orca/automl/auto_estimator.py
yushan111: Sure, I will add check in new PR

pyzoo/zoo/orca/automl/auto_estimator.py
yushan111: I will add check for the metric value.
Currently the `metric` in `auto_estimator` is not used if it creates a keras model.
The expected behavior might be we use the metric in `auto_estimator.fit` to evaluate trials and raise warning if the `metric` in `auto_estimator.fit` is not the same as the metric used to compile keras model.

Issue #3768, opened 2021-04-13
Friesian Table missing doc


hkvision:
At least add some inline doc to describe the functionality, argument data type and meanings.

Issue #3767, opened 2021-04-13
Friesian freq_limit is not user friendly to be a string


hkvision:
https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/friesian/feature/table.py#L112
In the scala implementation, freq_limit can only be a string. This is not user-friendly.
At least, in Python, we should at least support freq_limit to be an interger (or a dict?)

PR #3779, opened 2021-04-13
Add friesian docs, make freq_limit more user friendly

cyita:
related issue: #3768 , #3767

### DIFF
--- a/pyzoo/zoo/friesian/feature/table.py
+++ b/pyzoo/zoo/friesian/feature/table.py
@@ -244,7 +244,7 @@ class StringIndex(Table):
     @classmethod
     def read_parquet(cls, paths, col_name=None):
         """
-        Loads Parquet files, returning the result as a `FeatureTable`.
+        Loads Parquet files, returning the result as a `StringIndex`.
 
         :param paths: str or a list of str. The path/paths to Parquet file(s).
         :param col_name: str. The column name of the corresponding categorical column. If
--- a/pyzoo/zoo/friesian/feature/table.py
+++ b/pyzoo/zoo/friesian/feature/table.py
@@ -221,7 +221,7 @@ class FeatureTable(Table):
         """
         if not isinstance(columns, list):
             columns = [columns]
-        if freq_limit is not None:
+        if freq_limit:
             if isinstance(freq_limit, int):
                 freq_limit = str(freq_limit)
             elif isinstance(freq_limit, dict):
--- a/pyzoo/zoo/friesian/feature/table.py
+++ b/pyzoo/zoo/friesian/feature/table.py
@@ -40,44 +40,114 @@ class Table:
         return Table(df)
 
     def compute(self):
+        """
+        Trigger computation of Table
+
+        :return:
+        """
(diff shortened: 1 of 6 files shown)


pyzoo/zoo/friesian/feature/table.py
```diff
@@ -110,6 +205,22 @@ def encode_string(self, columns, indices):
         return FeatureTable(data_df)
 
     def gen_string_idx(self, columns, freq_limit):
+        """
+        Generate unique index value of categorical features
+
+        :param columns: str or a list of str, target columns to generate StringIndex.
+        :param freq_limit: int, str, dict or None.
+        Categories with a count/frequency below freq_limit will be ommited from the encoding.
+        Can be represented as both an integer, str, dict or None. For instance, 15, "15",
```
jason-dai, line 214: why do we support string anyway?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -40,44 +40,114 @@ def _clone(self, df):
         return Table(df)
 
     def compute(self):
+        """
```
jason-dai, line 43: is the python doc format supported by readthedocs @helenlly ?

pyzoo/zoo/friesian/feature/table.py
hkvision: The original dlrm example treats it as a string, this is strange to me... I think we can just support int or dict?

pyzoo/zoo/friesian/feature/table.py
helenlly: yes, but some formats need be modified: @cyita 
1) need add a blank line before " : return:"
2) need add a blank line before " :param " 
otherwise it can't show wrap in the page, 
3) to the words followed ":param ", no need to add return by yourself, otherwise it will show a blank line in the page

I have modified some codes for your reference here:
***********************************
def drop(self, *cols):
        """
        Returns a new Table that drops the specified column.
        This is a no-op if schema doesn't contain the given column name(s).
        
        :param cols: a string name of the column to drop, or a list of string name of the columns to drop.
        
        :return: A new Table that drops the specified column.
        """
        return self._clone(self.df.drop(*cols))
******************************************
you can see the effect in 
https://testhelendoc.readthedocs.io/en/latest/doc/PythonAPI/Orca/orca.html#module-zoo.friesian.feature.table.

pyzoo/zoo/friesian/feature/table.py
cyita: Updated and only support int, dict and None.

pyzoo/zoo/friesian/feature/table.py
cyita: OK. Preview https://yinatestdoc3.readthedocs.io/en/latest/doc/PythonAPI/Friesian/feature.html

cyita:
Jenkins passed http://10.239.47.210:18888/view/ZOO-PR-PIPELINE/job/ZOO-PR-Validation/5399/

PR #3831, opened 2021-04-21
Fix DeConv unit test on mac

hkvision:


hkvision:
Mac passed: http://10.239.47.210:18888/job/ZOO-NB-UnitTests-2.1-SCALA-MAC/327/

PR #3836, opened 2021-04-21
[PPML] Revised Scala README and fixed bugs

ColossusChang:
- Revised README.md for Scala to incorporate recent filename changes and enclave keys
- Fixed Dockerfile for Scala to copy `ppml-spark-submit.sh` to the docker image
- Added `ENCLAVE_KEY_PATH` when starting the scala-driver container


ppml/trusted-big-data-ml/scala/docker-graphene/README.md
```diff
@@ -1,10 +1,9 @@
 # Trusted Big Data ML
```
qiyuangong, line 1: Just a reminder: *.md and Dockerfile can be 644 permission (7 means runnable, which is not necessary for readme).

glorysdj (approved):
LGTM

PR #3841, opened 2021-04-22
[PPML] Revised Flink README and fixed bugs

ColossusChang:
- Revised README.md to incorporate new filename changes and enclave keys
- Slightly revised scripts to prevent containers from exiting after finishing their jobs


glorysdj (approved):
LGTM

PR #3844, opened 2021-04-22
Remove feature selection and use all features as default

yushan111:
User used to need including `all_available_features` in `search_space` while customizing a `recipe`. It might be a little confusing since users may only search with their customized models and don't need feature at all.

To improve usability, we remove the parameter and disable feature selection for now. And we will enable feature selection later with a better way to pass necessary parameters to search space in recipe.


yushan111:
Jenkins: http://10.239.47.210:18888/view/ZOO-PR/job/ZOO-PR-Validation/5509/

PR #3855, opened 2021-04-25
[PPML] Fixed some path and filename errors in READMEs

ColossusChang:
- The filename `environments.sh` is corrected to `environment.sh`
- The paths for generate-keys and generate-password are updated.


qiyuangong (approved):
LGTM

PR #3857, opened 2021-04-25
Add fill_int in Friesian fillna

cyita:


pyzoo/zoo/friesian/feature/table.py
```diff
@@ -100,20 +97,31 @@ def drop(self, *cols):
         """
         return self._clone(self.df.drop(*cols))
 
-    def fillna(self, value, columns):
+    def fillna(self, value, columns, fill_int=False):
```
jason-dai, line 100: why do we need the user to pass an extra parameter? can we infer that automatically?

cyita: The current implementation will raise exception when columns contain a long/double type column. We can pass value like 2.0 to fill double columns but for long columns, we cannot fill with -2147483648 <= value <= 2147483647.

cyita: Or we can check the type of columns here and if columns are all int type then we call fill_na_int

jason-dai: Yes, it's better to check the the type of columns on the Python side.

cyita:
Jenkins passed http://10.239.47.210:18888/job/ZOO-PR-Validation/5514/  http://10.239.47.210:18888/job/ZOO-PR-Validation/5505/

PR #3868, opened 2021-04-26
[PPML] Refine generate-keys permission and doc

qiyuangong:
* Refine generate-keys permission and doc. Add root permission warning
* Refine doc


glorysdj (approved):
LGTM

Issue #3772, opened 2021-04-13
Redesign automl recipe


yushan111:
The future usage of recipe might be
```python
# user customized model
def model_creator(config):
    model = tf.keras.models.Sequential([tf.keras.layers.Dense(config["hidden_size"])])
    model.compile()
    return model

# user customized search space corresponding to model hyper parameters
from zoo.orca.automl import hp
search_space = {
    # model parameters
    "hidden_size": hp.grid_search([32, 64, 128]),

    # optimization parameters
    "lr": hp.uniform(0.001, 0.01),
    "batch_size": hp.choice([32, 64, 128])}

# build auto estimator and fit
auto_est = AutoEstimator.from_keras(model_creator=model_creator,...)
auto_est.fit(data, 
             search_space=search_space,
             metric="mse",
             num_samples=1,
             early_stop={"max_iter": 5, "threshold": 0.01}, # same as stop in tune.run(),
             search_alg=,

            )

yushan111:
Breakdown tasks

- [ ] wrap all tune random distributions API in `zoo.orca.automl.hp`, change corresponding examples/documents.
- [ ] move `recipe.runtime_params` out into `searcher.compile` and expose in AutoEstimator.fit()
- [ ] move feature selection part into zouwu predictor and remove `all_available_features`  in `Recipe.search_space()`
- [ ] support model selection (with `hp.choice`?)

TheaperDeng:
Update 4/14/2021:
Breakdown tasks
- [ ] wrap all tune random distributions API in `zoo.orca.automl.hp`, change corresponding examples/documents. (#3776 Running Jenkins)
- [ ] move `recipe.runtime_params` out into `searcher.compile` and expose in `AutoEstimator.fit()` 

Currently we report three key in our ray tune trainable function: `training_iteration`, `reward_metric` and `checkpoint`. The first two can be seen as runtime stop recipe. `num_samples` I believe can be got from normal search_space with default value = 1. (Not sure)
BTW, the API we use to report `training_iteration`, `reward_metric` and `checkpoint` has deprecated. #3787

yushan111:
> Currently we report three key in our ray tune trainable function: `training_iteration`, `reward_metric` and `checkpoint`. The first two can be seen as runtime stop recipe. `num_samples` can be exposed to AutoEstimator as well.

[...]

PR #3879, opened 2021-04-27
Change AutoEstimator recipe API from recipe class to searchspace dicts and other params

TheaperDeng:
related to issue #3772


pyzoo/zoo/orca/automl/auto_estimator.py
```diff
@@ -98,8 +98,11 @@ def from_keras(*,
 
     def fit(self,
```
jason-dai, line 99: make it consistent with orca estimator (https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/base_estimator.py#L22) as much as possible

TheaperDeng:
pr validation pass:
http://10.239.176.111:18888/job/ZOO-PR-Validation/5576/

PR #3921, opened 2021-05-12
parquet add write coco

GitEasonXu:
add coco_dataset.py.


pyzoo/zoo/orca/data/image/coco_dataset.py
```diff
@@ -0,0 +1,125 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import numpy as np
+from PIL import Image
+from os import path as osp
+from pycocotools.coco import COCO
```
yangw1234, line 19: this package have to be installed right?

pyzoo/test/zoo/orca/data/test_write_parquet.py
```diff
@@ -163,6 +163,27 @@ def test_write_voc(orca_context_fixture):
...
+        ann_file = os.path.join(resource_path,
+                                "COCO/annotations/instances_{}.json".format(split_names))
+        output_path = os.path.join(temp_dir, "output_dataset")
+        write_coco(img_path, ann_file, output_path="file://" + output_path)
+        data, schema = ParquetDataset._read_as_dict_rdd("file://" + output_path)
+        data = data.collect()[0]
+        image_path = data["image_id"]
+        with open(image_path, "rb") as f:
+            image_bytes = f.read()
+        assert image_bytes == data['image']
```
yangw1234, line 181: maybe also test on label?

yangw1234:
@GitEasonXu as a general recommendation, I think it is better to rebase on master before developing a new feature and open a PR. This way we can get much more clear commit history.

pyzoo/zoo/orca/data/image/coco_dataset.py
gtfaiwxm: yes, the package must be installed.

gtfaiwxm: the package is normal API for reading coco format.

GitEasonXu:
Ok, I'll pay attention to it next time.

pyzoo/test/zoo/orca/data/test_write_parquet.py
gtfaiwxm: The test on image can know if there is an error in reading and writing. I think it is an optional operation.

pyzoo/zoo/orca/data/image/coco_dataset.py
yangw1234: Hi @jason-dai , if additional package must be installed do you prefer to add it to `install_required` or add it to `extra_requires`.

image

yangw1234: @GitEasonXu @gtfaiwxm how about we test if users have already installed pycocotools, if not, raise a error to users to install it.

yangw1234:
Could you also add "coco" here https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/data/image/parquet_dataset.py#L256

GitEasonXu:
OK, we will modify "coco" according your way.

PR #3934, opened 2021-05-14
Orca Data Image Write ImageNet Dataset

GZHoffie:
#3909.


yangw1234:
let's move the imagenet part to anther PR.

pyzoo/zoo/orca/data/image/parquet_dataset.py
```diff
@@ -246,3 +260,30 @@ def make_generator():
...
+
+def write_parquet(format, output_path, *args, **kwargs):
+    supported_format = {"imagenet", "mnist", "image_folder", "voc"}
+    if format not in supported_format:
+        raise ValueError(format + " is not supported, should be one of 'imagenet', 'mnist',"
+                         "'image_folder' and 'voc'.")
+
+    if format == "mnist":
+        _check_arguments(format, kwargs, "image_file", "label_file")
+        write_mnist(output_path=output_path, *args, **kwargs)
```
yangw1234, line 278: Maybe you can write a dict like 
```python
{
"mnist": (write_mnit, ["image_file", "label_file"]),
"image_folder": (write_from_directory, ["directory", "label_map"])
}
```
Then you just look up the function and the required arguments in the the dict without writing many if else bracnhes.

Issue #3933, opened 2021-05-14
jenkins: random failed


pinggao187:
Running zouwu tests

________________ TestZouwuModelAnomaly.test_threshold_gaussian _________________

self = <test_model_anomaly.TestZouwuModelAnomaly testMethod=test_threshold_gaussian>

    def test_threshold_gaussian(self):
        sample_num = 500
        # actual value
        y_test = np.full(sample_num, 2)
        mu, sigma, ratio = 3, 0.1, 0.01
        s = np.random.normal(mu, sigma, sample_num)
        y = y_test + s
    
        threshold = ThresholdEstimator().fit(y, y_test, mode="gaussian", ratio=ratio)
        from scipy.stats import norm
>       assert abs(threshold-(norm.ppf(1-ratio)*sigma+mu)) < 0.02
E       AssertionError: assert 0.025225962559095816 < 0.02
E        +  where 0.025225962559095816 = abs((3.2074088248449883 - ((2.3263478740408408 * 0.1) + 3)))
E        +    where 2.3263478740408408 = <bound method rv_continuous.ppf of <scipy.stats._continuous_distns.norm_gen object at 0x7f8c2205e518>>((1 - 0.01))
E        +      where <bound method rv_continuous.ppf of <scipy.stats._continuous_distns.norm_gen object at 0x7f8c2205e518>> = <scipy.stats._continuous_distns.norm_gen object at 0x7f8c2205e518>.ppf

../test/zoo/zouwu/model/anomaly/test_model_anomaly.py:131: AssertionError

jenkins link:
http://10.239.176.111:18888/job/ZOO-NB-UnitTests-2.4-PYTHON-MAC/336/console

TheaperDeng:
I will check this ut. It might be caused by an unnecessarily strict threshold. Please re-run for now.

pinggao187:
rerun
http://10.239.176.111:18888/view/ZOO-NB/job/ZOO-NB-UnitTests-2.4-PYTHON-MAC/337/

PR #3935, opened 2021-05-14
relax the threshold to fix anomaly detection ut random fail

TheaperDeng:
related to #3933 
It seems that we compare the estimators' value to a theoretical value. Relaxing the threshold will significantly reduce the random fail.


TheaperDeng:
pr validation combine pass:
http://10.239.176.111:18888/job/ZOO-PR-Validation/5577/
http://10.239.176.111:18888/job/ZOO-PR-Python-ExampleTests-py37-Spark3/1038/

PR #3946, opened 2021-05-18
align mtnet build/save/restore with base keras model

yushan111:


yushan111:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5604/

PR #3950, opened 2021-05-18
fix rayxshards for single node spark standalone

yushan111:


pyzoo/zoo/orca/data/ray_xshards.py
```diff
@@ -331,6 +331,9 @@ def _from_spark_xshards_ray_api(spark_xshards):
                 # if running in cluster, filter out driver ip
                 if ray_ctx.is_local or key != f"node:{driver_ip}":
```
hkvision, line 332: remove ray_ctx.is_local here?

pyzoo/zoo/orca/data/ray_xshards.py
```diff
@@ -331,6 +331,9 @@ def _from_spark_xshards_ray_api(spark_xshards):
...
                     nodes.append(key)
+        # for the case of single node spark standalone
```
hkvision, line 334: this can cover local mode and single node spark standalone

pyzoo/zoo/orca/data/ray_xshards.py
```diff
@@ -331,6 +331,9 @@ def _from_spark_xshards_ray_api(spark_xshards):
...
+        if not nodes:
+            nodes.append(f"node:{driver_ip}")
```
hkvision, line 336: Also check node:{driver_ip} in keys? Do we need to raise an error if it is not in keys and nodes=None?

pyzoo/zoo/orca/data/ray_xshards.py
yushan111: Good idea!

pyzoo/zoo/orca/data/ray_xshards.py
yushan111: Ray may make sure at least one node in resources, so nodes will not be empty. In that case, maybe we don't need to raise error?

yushan111: If node:{driver_ip} isn't in keys, it should be added to nodes in the condition of `key != f"node:{driver_ip}`

hkvision: ok Then we don't need to raise error.

hkvision (approved):

yushan111:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5624/
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5630/

PR #3951, opened 2021-05-19
modify k8s.py

Adria777:
add 3 notes in run az examples in k8s


jason-dai:
Don't change the entire file; only update the places where changes are needed, so that it is easier for others to review.

glorysdj:
> Don't change the entire file; only update the places where changes are needed, so that it is easier for others to review.

Lingqi will make a new pr to avoid this and also add detailed description

Issue #3944, opened 2021-05-18
FillZeroImpute does not call an inplace method, ut is also needed


TheaperDeng:
https://github.com/intel-analytics/analytics-zoo/blob/04d991d9b588cc2ede43f23648aec422280d1a61/pyzoo/zoo/zouwu/preprocessing/impute/impute.py#L37-L43

should be
```python
input_df.fillna(0, inplace=True)
```

PR #3954, opened 2021-05-19
Fix fillZeroImpute with implace method, add the unit test

cabuliwallah:
#3944


pyzoo/test/zoo/zouwu/preprocessing/test_fill.py
```diff
@@ -52,5 +52,12 @@ def test_lastfill(self):
         assert imputed_data.isna().sum().sum() == 0
         mse = last_fill.evaluate(imputed_data, 0.1)
 
+    def test_fillzero(self):
+        data = [[np.nan, 1], [np.nan, 2]]
+        df = pd.DataFrame(data)
+        imputor = FillZeroImpute()
+        imputor.impute(df)
+        assert df.isna().sum().sum()== 0
```
yushan111, line 60: maybe we shouldn't directly change df. The expected behavior should be: 
```python
imputed_df = imputor.imputed(df)
assert imputed_df.isna().sum().sum()== 0
assert df.isna().sum().sum() != 0

TheaperDeng (commented):
LGTM. 
We may move these imputors to utils function (e.g. zouwu/transform/impute.py) in next pr since they are stateless.

yushan111 (approved):

cabuliwallah:
http://10.239.176.111:18888/job/ZOO-PR-Validation/5658/

PR #3972, opened 2021-05-20
[Graphene] Update graphene to new version

DiegoCao:
Update Makefile, java.manifest.template(python version) and Dockerfile, corresponding with new version graphene


glorysdj:
a issue is opened at graohene:
https://github.com/oscarlab/graphene/issues/2388
after the issue is fixed, we should tests all the workload again, and merge this pr.

PR #3975, opened 2021-05-21
move datetime feature generation from feature transformer to chronos.data.utils

TheaperDeng:
This PR is kind of like an example to show how we will move data transforming code to utils functions in zouwu.preprocessing(or rename to zouwu.transform later) that will be called individually or by TSDataset

The code is adapted from 
https://github.com/intel-analytics/analytics-zoo/blob/b420b041ad4574a1b978104e625a8299f7842580/pyzoo/zoo/zouwu/feature/time_sequence.py#L535-L562


TheaperDeng:
pr validation combine pass
http://10.239.176.111:18888/job/ZOO-PR-Python-ExampleTests-py37-Spark3/1158/
http://10.239.176.111:18888/job/ZOO-PR-Validation/5690/

PR #4014, opened 2021-05-31
Add support for non-barrier mode to launch ray

hkvision:
First launch the ray master using spark and then launch raylets in two stages.

### DIFF
--- a/pyzoo/zoo/orca/common.py
+++ b/pyzoo/zoo/orca/common.py
@@ -257,7 +257,8 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
                          "but got: %s".format(cluster_mode))
     ray_args = {}
     for key in ["redis_port", "password", "object_store_memory", "verbose", "env",
-                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui", "barrier_mode"]:
+                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui",
+                "barrier_mode"]:
         if key in kwargs:
             ray_args[key] = kwargs[key]
     if "barrier_mode" not in ray_args:
--- a/pyzoo/zoo/orca/common.py
+++ b/pyzoo/zoo/orca/common.py
@@ -25,6 +25,7 @@ class OrcaContextMeta(type):
     _serialize_data_creator = False
     _train_data_store = "DRAM"
     __shard_size = None
+    _barrier_mode = True
 
     @property
     def log_output(cls):
@@ -119,6 +120,20 @@ class OrcaContextMeta(type):
             assert isinstance(value, int) and value > 0, \
                 "shard size should be either None or a positive integer."
         cls.__shard_size = value
+
+    @property
+    def barrier_mode(cls):
+        """
+        Whether to use Spark barrier mode to launch Ray, which is supported in Spark 2.4+ and when
+        dynamic allocation is disabled.
+        Default to be True.
+        """
+        return cls._barrier_mode
+
+    @barrier_mode.setter
+    def barrier_mode(cls, value):
+        assert isinstance(value, bool), "barrier_mode should either be True or False"
+        cls._barrier_mode = value
 
 
 class OrcaContext(metaclass=OrcaContextMeta):
@@ -242,9 +257,11 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
                          "but got: %s".format(cluster_mode))
     ray_args = {}
     for key in ["redis_port", "password", "object_store_memory", "verbose", "env",
-                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui"]:
+                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui", "barrier_mode"]:
         if key in kwargs:
             ray_args[key] = kwargs[key]
+    if "barrier_mode" not in ray_args:
+        ray_args["barrier_mode"] = OrcaContext.barrier_mode
     from zoo.ray import RayContext
     ray_ctx = RayContext(sc, **ray_args)
     if init_ray_on_spark:
--- a/pyzoo/zoo/orca/common.py
+++ b/pyzoo/zoo/orca/common.py
@@ -25,7 +25,6 @@ class OrcaContextMeta(type):
     _serialize_data_creator = False
     _train_data_store = "DRAM"
     __shard_size = None
-    _barrier_mode = True
 
     @property
     def log_output(cls):
@@ -128,12 +127,11 @@ class OrcaContextMeta(type):
         dynamic allocation is disabled.
         Default to be True.
         """
-        return cls._barrier_mode
+        return ZooContext.barrier_mode
 
     @barrier_mode.setter
     def barrier_mode(cls, value):
-        assert isinstance(value, bool), "barrier_mode should either be True or False"
-        cls._barrier_mode = value
+        ZooContext.barrier_mode = value
 
 
 class OrcaContext(metaclass=OrcaContextMeta):
@@ -257,12 +255,9 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
                          "but got: %s".format(cluster_mode))
     ray_args = {}
     for key in ["redis_port", "password", "object_store_memory", "verbose", "env",
-                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui",
-                "barrier_mode"]:
+                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui"]:
         if key in kwargs:
             ray_args[key] = kwargs[key]
-    if "barrier_mode" not in ray_args:
-        ray_args["barrier_mode"] = OrcaContext.barrier_mode
     from zoo.ray import RayContext
     ray_ctx = RayContext(sc, **ray_args)
     if init_ray_on_spark:
--- a/pyzoo/zoo/ray/raycontext.py
+++ b/pyzoo/zoo/ray/raycontext.py
@@ -213,6 +213,8 @@ class RayServiceFuncGenerator(object):
         if "envs" in self.python_loc:  # conda environment
             python_bin_dir = "/".join(self.python_loc.split("/")[:-1])
             return "{}/python {}/ray".format(python_bin_dir, python_bin_dir)
+        elif self.python_loc == "python_env/bin/python":  # conda yarn archive on the executor
+            return "python_env/bin/python python_env/bin/ray"
         else:  # system environment with ray installed; for example: /usr/local/bin/ray
             return "ray"
 
--- a/pyzoo/zoo/ray/raycontext.py
+++ b/pyzoo/zoo/ray/raycontext.py
@@ -20,6 +20,8 @@ import re
 import random
 import signal
 import warnings
+import tempfile
+import filelock
 import multiprocessing
 
 from zoo.ray.process import session_execute, ProcessMonitor
@@ -143,6 +145,12 @@ class RayServiceFuncGenerator(object):
         self.labels = \
             """--resources '{"_mxnet_worker": %s, "_mxnet_server": %s, "_reserved": %s}'""" \
             % (1, 1, 2)
+        base_path = tempfile.gettempdir()  # Temp path to save lock and flag files for launching ray.
+        import uuid
+        # Add a unique id to the flag so that this won't affect other programs even if it is not removed.
+        self.ray_master_flag = os.path.join(base_path, "ray_master_" + uuid.uuid4().hex)
+        self.ray_master_lock = os.path.join(base_path, "ray_master_start.lock")
+        self.raylet_lock = os.path.join(base_path, "raylet_start.lock")
 
     def gen_stop(self):
         def _stop(iter):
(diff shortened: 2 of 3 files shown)


pyzoo/zoo/ray/raycontext.py
```diff
@@ -500,9 +551,21 @@ def _start_cluster(self):
         print("Start to launch ray on cluster")
         ray_rdd = self.sc.range(0, self.num_ray_nodes,
                                 numSlices=self.num_ray_nodes)
-        # The first ip would be used to launch ray master.
-        process_infos = ray_rdd.barrier().mapPartitions(
-            self.ray_service.gen_ray_start(self.cluster_ips[0])).collect()
+        if self.barrier_mode:
+            # The first ip would be used to launch ray master.
+            process_infos = ray_rdd.barrier().mapPartitions(
+                self.ray_service.gen_ray_start(self.cluster_ips[0])).collect()
```
hkvision, line 557: See this previous PR: https://github.com/intel-analytics/analytics-zoo/pull/2836
If running in k8s, BarrierTaskContext gets wrong container ips so I abort the logic of making partition 0 task as ray master and launch the ray master given the master ip address (i.e. the first cluster ip address). All the raylet tasks need to know the redis address beforehand and thus the master address is passed as an argument.

hkvision:
@jason-dai To at maximum not affect the original logic, I write some new code segments for non-barrier mode. We can do code refactor afterwards when the new code is stable.

pyzoo/zoo/ray/raycontext.py
```diff
@@ -431,7 +482,7 @@ def purge(self):
         else:
             self.sc.range(0,
                           self.num_ray_nodes,
-                          numSlices=self.num_ray_nodes).barrier().mapPartitions(
```
hkvision, line 434: removing this barrier won't affect right? @jason-dai

pyzoo/zoo/ray/raycontext.py
```diff
@@ -500,9 +551,21 @@ def _start_cluster(self):
...
                                 numSlices=self.num_ray_nodes)
-        # The first ip would be used to launch ray master.
-        process_infos = ray_rdd.barrier().mapPartitions(
-            self.ray_service.gen_ray_start(self.cluster_ips[0])).collect()
+        if self.barrier_mode:
+            # The first ip would be used to launch ray master.
+            process_infos = ray_rdd.barrier().mapPartitions(
+                self.ray_service.gen_ray_start(self.cluster_ips[0])).collect()
+        else:
+            master_process_infos = ray_rdd.mapPartitionsWithIndex(self.ray_service.gen_ray_master_start()).collect()
```
hkvision, line 559: When launching ray master first, there is no such issue above and thus I use the first partition task for the ray master.

pyzoo/zoo/ray/raycontext.py
```diff
@@ -207,6 +215,52 @@ def _get_ray_exec(self):
...
+                            redis_address=redis_address,
+                            ray_exec=self.ray_exec,
+                            password=self.password,
+                            ray_node_cpu_cores=self.ray_node_cpu_cores,
+                            labels=self.labels,
+                            object_store_memory=self.object_store_memory,
+                            extra_params=self.extra_params),
+                        tag="raylet")
+                    kill_redundant_log_monitors(redis_address=redis_address)
+            # Cannot remove ray_master_flag at the end of this task since no barrier is guaranteed.
```
hkvision, line 259: We can either launch another task to remove the flag, or just leave it here. Since I add a unique id to the flag, this should not affect other Ray clusters.

hkvision:
Tested on single node standalone; WIP for yarn.

pyzoo/zoo/ray/raycontext.py
```diff
@@ -143,6 +145,12 @@ def __init__(self, python_loc, redis_port, ray_node_cpu_cores,
         self.labels = \
             """--resources '{"_mxnet_worker": %s, "_mxnet_server": %s, "_reserved": %s}'""" \
             % (1, 1, 2)
+        base_path = tempfile.gettempdir()  # Temp path for lock and flag files when launching ray.
+        import uuid
+        # Add a unique id to the flag so that it won't affect other programs even if not removed.
+        self.ray_master_flag = os.path.join(base_path, "ray_master_" + uuid.uuid4().hex)
```
jason-dai, line 151: executors in the same node for the same job will have different uuid, no?

hkvision: This uuid is generated on the driver, one application will have a uuid right?

yangw1234:
It is nice to remove barrier mode and support RayOnSpark below spark 2.4. However, does it solve the dynamic allocation problem?

After RayOnSpark is started, many of our workloads just scheduling jobs using ray. While tasks are runing on ray workers, from spark point of view, those executors are completely idle. With dynamic allocation enabled, I am afraid those executors might be released and caused the tasks in ray workers to fail. (Except we ask users to set `spark.dynamicAllocation.minExecutors = spark.dynamicAllocation.maxExecutors = spark.dynamicAllocation.initialExecutors` )

hkvision:
> It is nice to remove barrier mode and support RayOnSpark below spark 2.4. However, does it solve the dynamic allocation problem?
> 
> After RayOnSpark is started, many of our workloads just scheduling jobs using ray. While tasks are runing on ray workers, from spark point of view, those executors are completely idle. With dynamic allocation enabled, I am afraid those executors might be released and caused the tasks in ray workers to fail. (Except we ask users to set `spark.dynamicAllocation.minExecutors = spark.dynamicAllocation.maxExecutors = spark.dynamicAllocation.initialExecutors` )

Okay. You mean if user cannot set dynamic allocation to be false, even if they can start ray, the ray processes would be killed if spark executors are released? This can be a problem... Do you have any other ways besides setting `spark.dynamicAllocation.minExecutors = spark.dynamicAllocation.maxExecutors = spark.dynamicAllocation.initialExecutors`?

Anyway, aborting barrier mode can at least support spark <2.4...

hkvision:
If no other Spark applications want to share resources with the current applications, will the resources be released?

pyzoo/zoo/orca/common.py
```diff
@@ -242,9 +257,12 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
                          "but got: %s".format(cluster_mode))
     ray_args = {}
     for key in ["redis_port", "password", "object_store_memory", "verbose", "env",
-                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui"]:
+                "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui",
+                "barrier_mode"]:
         if key in kwargs:
             ray_args[key] = kwargs[key]
+    if "barrier_mode" not in ray_args:
+        ray_args["barrier_mode"] = OrcaContext.barrier_mode
```
jason-dai, line 265: Do we need to pass `barrier_mode` as parameter to `RayContext`? We may direct check `OrcaContext.barrier_mode` in `RayContext`

pyzoo/zoo/ray/raycontext.py
```diff
@@ -500,9 +551,22 @@ def _start_cluster(self):
         print("Start to launch ray on cluster")
         ray_rdd = self.sc.range(0, self.num_ray_nodes,
                                 numSlices=self.num_ray_nodes)
-        # The first ip would be used to launch ray master.
-        process_infos = ray_rdd.barrier().mapPartitions(
-            self.ray_service.gen_ray_start(self.cluster_ips[0])).collect()
+        if self.barrier_mode:
```
jason-dai, line 554: just check `OrcaContext.barrier_mode` here

pyzoo/zoo/ray/raycontext.py
```diff
@@ -402,7 +453,7 @@ def info_fn(iter):
             yield get_node_ip()
 
         ips = self.sc.range(0, self.num_ray_nodes,
-                            numSlices=self.num_ray_nodes).barrier().mapPartitions(info_fn).collect()
+                            numSlices=self.num_ray_nodes).mapPartitions(info_fn).collect()
```
jason-dai, line 456: it might be safer to use `sc.range(0, nodes * cores)` and then preform `distinct()`?

pyzoo/zoo/ray/raycontext.py
jason-dai: Similarly, it might be safer to use `sc.range(0, nodes * cores)`?

pyzoo/zoo/ray/raycontext.py
```diff
@@ -143,6 +145,12 @@ def __init__(self, python_loc, redis_port, ray_node_cpu_cores,
         self.labels = \
             """--resources '{"_mxnet_worker": %s, "_mxnet_server": %s, "_reserved": %s}'""" \
             % (1, 1, 2)
+        base_path = tempfile.gettempdir()  # Temp path for lock and flag files when launching ray.
```
jason-dai, line 148: The `tempdir` on each worker might be different from the diver node or other workers? I think you should still try to retrieve it on each worker node.

pyzoo/zoo/ray/raycontext.py
jason-dai: ok - then why don't you also use it for `ray_master_lock` and `raylet_lock`?

pyzoo/zoo/ray/raycontext.py
hkvision: okay.

pyzoo/zoo/ray/raycontext.py
hkvision: Since the flag may block other applications permanently; while the locks may only make other applications wait for a bit while.
Anyway, we can do this for all of them to make safe.

pyzoo/zoo/ray/raycontext.py
hkvision: oh yeah, makes sense. I will change back. So for the base_path it is generated on each executor and we can generate the uuids on the driver, one for each appliation.

pyzoo/zoo/orca/common.py
hkvision: But the logic is OrcaContext depends on RayContext, it is strange to put OrcaContext related things in RayContext?

pyzoo/zoo/ray/raycontext.py
hkvision: But the logic is OrcaContext depends on RayContext, it is strange to put OrcaContext related things in RayContext?

jason-dai: > But the logic is OrcaContext depends on RayContext, it is strange to put OrcaContext related things in RayContext?

The you may put it in `ZooContext`, just as `log_output`

pyzoo/zoo/ray/raycontext.py
hkvision: changed. the same as in mpi_runner

pyzoo/zoo/ray/raycontext.py
hkvision: yes. calling ray stop on a node for multiple times should have no issue. updated.

pyzoo/zoo/ray/raycontext.py
hkvision: Good idea! Fixed.

hkvision:
All comments address. Take a further look? @jason-dai

pyzoo/zoo/ray/raycontext.py
```diff
@@ -217,29 +276,29 @@ def _start_ray_services(iter):
             print("master address {}".format(master_ip))
             redis_address = "{}:{}".format(master_ip, self.redis_port)
             process_info = None
-            import tempfile
-            import filelock
             base_path = tempfile.gettempdir()
-            master_flag_path = os.path.join(base_path, "ray_master_initialized")
+            ray_master_flag_path = os.path.join(base_path, self.ray_master_flag)
```
jason-dai, line 280: move this to `if current_ip == master_ip`?

pyzoo/zoo/ray/raycontext.py
```diff
@@ -381,6 +440,7 @@ def __init__(self, sc, redis_port=None, password="123456", object_store_memory=N
                 include_webui=self.include_webui,
                 extra_params=self.extra_params)
         RayContext._active_ray_context = self
+        self.total_cores = self.num_ray_nodes * self.ray_node_cpu_cores
```
jason-dai, line 443: use "spark_executors * spark_cores", instead of "ray_nodes*ray_cores"

pyzoo/zoo/ray/raycontext.py
hkvision: As in Line 315-316 we are to remove the flag after the barrier (after the master has launched), and this is outside the if branch. So this flag needs to be visible not only in the if branch?

pyzoo/zoo/ray/raycontext.py
hkvision: Actually the logic here https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/ray/raycontext.py#L333-L368 by default makes ray_nodes = num_executors and ray_cores = executor_cores. It is possible in some cases reported by customers that executor cores or num executors can not be obtained from the SparkContext...

pyzoo/zoo/ray/raycontext.py
jason-dai: ok

hkvision:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5744/ jenkins passed

pyzoo/zoo/ray/raycontext.py
jason-dai: The user can override `num_executors` and `executor_cores` when initializing `RayContext`? Maybe you should records these in `RayContext`.

hkvision: Normally users shouldn't do this, only if when the config can't be obtained from sc, users need to provide by themselves. I have written in the document: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/ray/raycontext.py#L292

hkvision:
http://10.239.176.111:18888/job/ZOO-PR-Validation/5748/

pyzoo/zoo/ray/raycontext.py
jason-dai: I think in AutoML we sometimes do this (to manage the CPU resources for trials) @yushan111 ? Anyway you cannot stop user from specifying these values.

yushan111: For automl trial resources scheduling, we may not change the executors and cores in `RayContext`, since they are the total resources for ray. We schedule resources for each trial base on the `num_executors` and `executor_cores`. (e.g We have `num_executors` = 2, `executor_cores` = 8 in `RayContext`, for automl, if we want 4 trials run in parallel, we would set `resources_per_trial`={‘cpu’: 4} for `AutoEstimator`).

hkvision:
Tested on yarn. If multiple executors are on the same node, it is possible that one executor launches both the master and the raylet(s).

jason-dai:
> Tested on yarn. If multiple executors are on the same node, it is possible that one executor launches both the master and the raylet(s).

Does it matter?

hkvision:
> > Tested on yarn. If multiple executors are on the same node, it is possible that one executor launches both the master and the raylet(s).
> 
> Does it matter?

It wouldn't matter in theory. Just record this fact here to track if necessary :)

hkvision:
http://10.239.176.111:18888/job/ZOO-PR-Validation/5776/ jenkins passed

hkvision:
Merge it first. Will further modify if any issues detected.

PR #4025, opened 2021-06-02
[OpenVINO] Change predict output from JTensor to List<JTensor>

qiyuangong:
* Change predict output from JTensor to List<JTensor>
* Fix maskrcnn, Yolo output issues.


glorysdj (approved):
LGTM

PR #4030, opened 2021-06-02
Add FeatureTable operations for dummy pipeline

hkvision:
To fully support all the preprocessing operations in the dummy pipeline.
Would remove the example before merge.


pyzoo/zoo/friesian/feature/table.py
```diff
@@ -59,6 +58,18 @@ def _read_json(paths, cols):
                 raise Exception("cols should be a column name or list of column names")
         return df
 
+    @staticmethod
+    def _read_csv(paths, delimiter=",", schema=None):
+        # TODO: better support schema instead of exposing Spark DataTypes?
```
jason-dai, line 63: similar to xshards

pyzoo/zoo/examples/friesian/feature/dummy_pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+from pyspark.sql.types import StructField, StructType, StringType, IntegerType, FloatType
+
+sc = init_orca_context(cores="*")
+
+LABEL_COL = "_c0"
+INT_COLS = ["_c%d" % i for i in list(range(1, 14))]
+CAT_COLS = ["_c%d" % i for i in list(range(14, 40))]
+
+tbl = FeatureTable.read_csv("/home/kai/RS/reco_dummy_pipeline/data/small.csv")
+tbl = tbl.cast([LABEL_COL] + INT_COLS, "int")
```
jason-dai, line 13: What's the original type? And can we add an argument `dtype` to `read_csv` (similar to pandas).

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -160,6 +172,9 @@ def dropna(self, columns, how='any', thresh=None):
         """
         return self._clone(self.df.dropna(how, thresh, subset=columns))
 
+    def drop_duplicates(self, columns=None):
```
jason-dai, line 175: use `subset` instead of `columns` (same as pandas and spark)

pyzoo/zoo/examples/friesian/feature/dummy_pipeline.py
hkvision: When Aaron save the sampled data to csv, he converts all int fields to float. When reading the csv if we don't provide the schema, all the fields are read as string.
OK. I will take a look at how pandas handles dtype.

pyzoo/zoo/friesian/feature/table.py
hkvision: ok.

pyzoo/zoo/examples/friesian/feature/dummy_pipeline.py
jason-dai: I think in `XShards.read_csv`, we added some schema support?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -380,9 +399,50 @@ def encode_string(self, columns, indices):
...
+
+    def one_hot_encode(self, columns, sizes=None, keep_original_columns=False):
+        # Assume the columns have been encoded to index and the dimension is small enough for one hot.
+        if not isinstance(columns, list):
+            columns = [columns]
+        if sizes and not isinstance(sizes, list):
+            sizes = [sizes]
+            assert len(columns) == len(sizes)
+        data_df = self.df
+
```
jason-dai, line 414: Instead of calling udf many times (which can be expensive), call the udf only once: `onehot_encode(columns, sizes)` can take two arrays as input and return an array (or struct):
```python
data_df = data_df.withColumn("fresian_onehot", onehot_encode(array(cols), sizes))

for i in range(len(columns)):
    for j in range(sizes[i] + 1):
         col_name = ...
         k = ...
         data_df =  data_df.withColumn(col_name, data_df.fresian_onehot[k])

if not keep_original_columns:
    data_df.drop(...)
```

pyzoo/zoo/friesian/feature/table.py
hkvision: Moved to this PR: https://github.com/intel-analytics/analytics-zoo/pull/4128
Will remove in this one.

pyzoo/zoo/friesian/feature/table.py
hkvision: In XShards, we previously referred to Koalas for some pandas argument mapping: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/data/pandas/preprocessing.py#L89
But the result is an RDD of pandas DataFrames.

pyzoo/zoo/examples/friesian/feature/dummy_pipeline.py
hkvision: In XShards, we already have schema support through dtypes: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/data/pandas/preprocessing.py#L234

hkvision: To make this PR smaller, I will split read_csv to a new PR for discussion.

hkvision: Spark by default reads all columns as string types. Refer to XShards, what about the following arguments in the first stage:
- inferSchema by default to be true (this may be slower, but most likely the types will be correct)
- header by default to be infer
- names to input column names in case header can't be inferred
- dtypes as a string or a dict in case schema is not inferred.

pyzoo/zoo/friesian/feature/table.py
hkvision: Updated. Now only uses one udf to do all the encodings.

hkvision:
Put the preprocessing script to here: https://github.com/analytics-zoo/friesian/blob/dummy-pipeline/reco_dummy_pipeline/pipeline/az_preprocess.py

hkvision:
@jason-dai I will merge this first. If you have any further comments on read_csv and one_hot_encode, I will address them in a following PR.

pyzoo/zoo/examples/friesian/feature/dummy_pipeline.py
hkvision: Updated. Take a look.

hkvision:
http://10.239.176.111:18888/job/ZOO-PR-Python-Spark-2.4-py36/4380/console

hkvision:
jenkins: http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5939/
the failures are irrelevant.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -449,15 +505,123 @@ def encode_string(self, columns, indices):
...
+                one_hot_cols.append(one_hot_col)
+                data_df = data_df.withColumn(one_hot_col,
+                                             data_df.friesian_onehot[i][j])
+            if keep_original_columns:
+                all_columns = cols_before + [col_name] + one_hot_cols + cols_after
+            else:
+                all_columns = cols_before + one_hot_cols + cols_after
+            data_df = data_df.select(*all_columns)
+        data_df = data_df.drop("friesian_onehot")
+        return FeatureTable(data_df)
```
jenniew, line 613: Why one hot encode will generate multiple columns? why not array or vector?

hkvision: This aligns with the dummy pipeline implementation and their model.

PR #4037, opened 2021-06-03
FeatureTable join-groupby

yaxinchen666:


pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,41 @@ def gen_cat(items):
             df = df.withColumn(c.replace("item", "category"), cat_udf(col(c)))
         return FeatureTable(df)
 
+    def join_groupby(self, cat_cols, cont_cols, stats="count"):
```
jason-dai, line 636: Instead of implementing `join_groupby`, we can implement a `groupby` as follows (used full for both recsys and wechat dataset):

```python
# run "group_by" on "columns", which can be arbitrary columns
# for each grouped dataframe, optionally run "agg", which can be either string (list)
#        or dictionary (similar to Pandas or Spark)
# optionally join the result with the original dataframe
def group_by(columns = ..., agg = ..., join = false)
```

yaxinchen666: https://spark.apache.org/docs/latest/api/python/_modules/pyspark/sql/dataframe.html#DataFrame.groupBy
The built-in `groupBy` of spark DataFrame returns `GroupedData` instead of DataFrame, while agg/join after groupby returns DataFrame. Is it ok to put them together?
Or can we make `agg` as a required argument instead of an optional argument?

hkvision: Yeah, I think we need to make agg required, default to be sum to make the output a DataFrame?

jason-dai: Is it useful to perform a `groupBy` only? We can simply return `GroupedData._df` if `agg` is None.

yaxinchen666: Ususally `groupBy` is followed by `agg` and returns a `DataFrame` finally, except for one method, `pivot`, of `GroupedData`, which returns a `GroupedData`. https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.sql.GroupedData.pivot.html
Is it necessary to support `pivot`? Or maybe we can make it as an optional argument of `groupBy`, because `pivot` is also usually followed by `agg`?
Btw, it seems `GroupedData._df` is the same as the original dataframe before performing `groupBy`.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +632,38 @@ def gen_cat(items):
...
+    def group_by(self, columns=None, agg_exprs={"*":"count"}, join=False):
+        """
+        Group the Table with specified columns and then run aggregation. Optionally join the result
+        with the original Table.
+
+        :param columns: str or list of str. Columns to group the Table.
+        :param agg_exprs: dict or list. Aggragation functions to be applied to grouped Table.
+               If agg_exprs is a single dict mapping from string to string, then the key is
+               the column to perform aggregation on, and the value is the aggregate function.
+               Alternatively, agg_exprs can also be a list of aggregate Column expressions.
```
hkvision, line 644: list some aggregate column experessions in the doc?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +632,38 @@ def gen_cat(items):
             df = df.withColumn(c.replace("item", "category"), cat_udf(col(c)))
         return FeatureTable(df)
 
+    def group_by(self, columns=None, agg_exprs={"*":"count"}, join=False):
```
hkvision, line 635: can the default is str instead of a dict? str or list of str is more commonly used?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +632,38 @@ def gen_cat(items):
...
+        """
+        if columns is None:
+            grouped_data = self.df.groupBy()
+        else:
+            grouped_data = self.df.groupBy(columns)
+
+        if isinstance(agg_exprs, dict):
+            agg_df = grouped_data.agg(agg_exprs)
+        elif isinstance(agg_exprs, list):
+            agg_df = grouped_data.agg(*agg_exprs)
```
hkvision, line 657: is a single string supported and we can convert it into a list?

pyzoo/zoo/friesian/feature/table.py
hkvision: agg_exprs -> agg?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +632,38 @@ def gen_cat(items):
...
+        with the original Table.
+
+        :param columns: str or list of str. Columns to group the Table.
+        :param agg_exprs: dict or list. Aggragation functions to be applied to grouped Table.
+               If agg_exprs is a single dict mapping from string to string, then the key is
+               the column to perform aggregation on, and the value is the aggregate function.
+               Alternatively, agg_exprs can also be a list of aggregate Column expressions.
+        :param join: boolean. If join is True, join the aggragation result with original Table.
+
+        :return: A new Table.
```
hkvision, line 647: A new Table with aggregated column fields.

pyzoo/zoo/friesian/feature/table.py
yaxinchen666: The built-in agg can support only dict or list of sql.functions, because it needs to specify both operation (count/sum/...) and the column to apply this operation?

yaxinchen666: I feel kind of weird if writing df.agg(agg)... is it ok?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,66 @@ def gen_cat(items):
...
+        :param columns: str or list of str. Columns to group the Table.
+        :param agg: str, dict or list. Aggragation functions to be applied to grouped Table.
+               If agg is a str, then agg is the aggregate function and the aggregation is performed
+               on all columns that are not in `columns`.
+               If agg is a single dict mapping from string to string, then the key is the column
+               to perform aggregation on, and the value is the aggregate function.
+               If agg is a single dict mapping from string to string/list, then the key is the
+               column to perform aggregation on, and the value is the aggregate function(s) among
+               ['count', 'sum', 'mean', 'std', 'variance', 'max', 'min'].
+               Alternatively, agg can also be a list of aggregate Column expressions.
```
jason-dai, line 650: We don't need to expose Spark aggregation functions in the doc; just document aggregation strings that we support.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,66 @@ def gen_cat(items):
...
+        Group the Table with specified columns and then run aggregation. Optionally join the result
+        with the original Table.
+
+        :param columns: str or list of str. Columns to group the Table.
+        :param agg: str, dict or list. Aggragation functions to be applied to grouped Table.
+               If agg is a str, then agg is the aggregate function and the aggregation is performed
+               on all columns that are not in `columns`.
+               If agg is a single dict mapping from string to string, then the key is the column
+               to perform aggregation on, and the value is the aggregate function.
+               If agg is a single dict mapping from string to string/list, then the key is the
```
jason-dai, line 647: mapping from string to list

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,66 @@ def gen_cat(items):
...
 
+    def group_by(self, columns=[], agg="count", join=False):
+        """
+        Group the Table with specified columns and then run aggregation. Optionally join the result
+        with the original Table.
+
+        :param columns: str or list of str. Columns to group the Table.
+        :param agg: str, dict or list. Aggragation functions to be applied to grouped Table.
+               If agg is a str, then agg is the aggregate function and the aggregation is performed
+               on all columns that are not in `columns`.
```
jason-dai, line 644: Which strings do we support for agg?

yaxinchen666: If agg is string or dict mapping string to string, it calls the built-in agg of pyspark and supports what is marked as “Aggregate function” in https://github.com/apache/spark/blob/master/python/pyspark/sql/functions.py. There are 32 aggregate functions.
If agg is a dict mapping from string to list, it uses a dict `stats_func` to map each string in list to aggregate function and supports ['count', 'sum', 'mean', 'std', 'variance', 'max', 'min'].
https://github.com/intel-analytics/analytics-zoo/pull/4037/files#diff-f627bae6816b7dc510932e3da8944b9970604771f47f9785e80943c79aaede75R673
Do I need to put all aggregate functions supported by pyspark into `stats_func`?

jason-dai: We should not expect users to read Spark document; instead, just list all the valid string that we support in the Python API doc.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,66 @@ def gen_cat(items):
             df = df.withColumn(c.replace("item", "category"), cat_udf(col(c)))
         return FeatureTable(df)
 
+    def group_by(self, columns=[], agg="count", join=False):
```
hkvision, line 636: why the default value is an empty list? what does it mean?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,66 @@ def gen_cat(items):
...
+               column to perform aggregation on, and the value is the aggregate function(s) among
+               ['count', 'sum', 'mean', 'std', 'variance', 'max', 'min'].
+               Alternatively, agg can also be a list of aggregate Column expressions.
+
+               Examples:
+               agg="sum"
+               agg={"*":"count"}
+               agg={"col_1":"sum", "col_2":["count", "mean"]}
+               import pyspark.sql.functions as F
+               agg=[F.last("col_1"), F.sum("col_2"), F.mean("col_2")]
```
hkvision, line 657: remove spark related in the doc.

pyzoo/zoo/friesian/feature/table.py
yaxinchen666: Got it.

pyzoo/zoo/friesian/feature/table.py
yaxinchen666: df.groupBy([]).agg is the same as df.groupBy().agg, which gives the statistics of the whole column.

pyzoo/zoo/friesian/feature/utils.py
```diff
@@ -62,3 +62,10 @@ def check_col_exists(df, columns):
     col_not_exist = list(filter(lambda x: x not in df_cols, columns))
     if len(col_not_exist) > 0:
         raise ValueError(str(col_not_exist) + " do not exist in this Table")
+
+
+def str_to_list(arg, arg_name):
```
hkvision, line 67: no need this?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,76 @@ def gen_cat(items):
...
+                stat_func = getattr(F, stat)
+                agg_exprs_list += [stat_func(agg_column) for agg_column in self.df.columns \
+                        if agg_column not in columns]
+            agg_df = grouped_data.agg(*agg_exprs_list)
+        elif isinstance(agg, dict):
+            if all(isinstance(agg[agg_column], str) for agg_column in agg):
+                agg_df = grouped_data.agg(agg)
+            else:
+                agg_exprs_list = []
+                for agg_column in agg:
```
hkvision, line 687: change to `for col, stats in agg.items()`?
also for line 683

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -632,6 +633,76 @@ def gen_cat(items):
...
+                    stats = agg[agg_column]
+                    if isinstance(stats, str):
+                        stats = [stats]
+                    assert isinstance(stats, list), "value in agg should be str or list of str"
+                    for stat in stats:
+                        stat_func = getattr(F, stat)
+                        agg_exprs_list += [stat_func(agg_column)]
+                agg_df = grouped_data.agg(*agg_exprs_list)
+        else:
+            raise TypeError("agg should be dict or list")
```
hkvision, line 697: agg should be a str, list of str or a dict

hkvision (approved):

hkvision:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5801/

PR #4040, opened 2021-06-03
Change signature for automl base model and move model builders

yushan111:
This PR mainly change the interface of zoo.automl.abstract.BaseModel

- change `fit_eval`
  from
  ```python
    def fit_eval(self, data, validation_data=None, mc=False, verbose=0, **config):
  ``` 
  to
  ```python
    def fit_eval(self, data, validation_data=None):
  ``` 
- remove `predict` and `evaluate`

This PR also moves all `model_builder`s to their corresponding `model` module.


yushan111:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5727/

PR #4051, opened 2021-06-04
FeatureTable TargetEncode

yaxinchen666:


pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1150,6 +1151,239 @@ def split(self, ratio, seed=None):
...
+    def gen_target(self, cat_cols, target_cols, target_mean=None, smooth=20, kfold=2, fold_seed=42,
+                   fold_col="__fold__", out_cols=None, name_sep="_"):
+        """
+        For each categorical column / column group in cat_cols, calculate the mean of target
+        columns in target_cols.
+        :param cat_cols: str, or list of (str or list of str). Categorical columns / column groups
+               to target encode. If an element in the list is a str, then it is a categorical
+               column; otherwise if it is a list of str, then it is a categorical column group.
+        :param target_cols: str, or list of str. Numeric target column to calculate the mean.
+        :param target_mean: dict. {target column : mean}. Provides mean of target column to
```
hkvision, line 1163: Provides global mean of the target column(s)
calculation typo.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1150,6 +1151,239 @@ def split(self, ratio, seed=None):
...
+        # generate fold_col
+        result_df = self.df
+        if kfold > 1:
+            if fold_col not in self.df.columns:
+                if fold_seed is None:
+                    windowSpec = Window.orderBy(F.lit(1))
+                    result_df = result_df.withColumn(
+                        fold_col,
+                        F.row_number().over(windowSpec) % F.lit(kfold)
+                    )
```
hkvision, line 1235: will this result in a large window?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1150,6 +1151,239 @@ def split(self, ratio, seed=None):
...
+                org_all_df = result_df.groupBy(cat_col)
+            else:
+                org_all_df = result_df.groupBy(*cat_col)
+            org_all_df = org_all_df.agg(*sum_list, F.count("*").alias(cat_col_name + "_all_count"))
+            all_df = org_all_df
+            for target_col, out_col in zip(target_cols, out_col_list):
+                global_target_mean = target_mean_dict[target_col]
+                all_func = udf(
+                    lambda s, count:
+                    None if s is None else (s + global_target_mean * smooth) / (count + smooth),
```
hkvision, line 1261: what case will s be None?

hkvision: s -> cat_sum
count -> cat_count?

hkvision:
We may need to
- Add a method probably `target_encode()` to combine generate targets and encode targets?
- Add a save method for TargetEncode?

pyzoo/zoo/friesian/feature/table.py
yaxinchen666: Yes. It takes the whole dataframe as window. By default (when fold_seed is provided), it will use rand to generate fold column instead of using window.

pyzoo/zoo/friesian/feature/table.py
yaxinchen666: s is None when all values in target column of this category are None, but generally target column should not have None value. Should I remove this?

yaxinchen666:
> We may need to
> 
> * Add a method probably `target_encode()` to combine generate targets and encode targets?
> * Add a save method for TargetEncode?

- `target_encode()` takes another Table (test set) as input and returns new train set & test set?
- I think it is a little bit difficult to add a save method? Because it contains not only two dataframes & but category column(s) (str/list) & global means (dict) & kfold. Is there a good way to store these information?

pyzoo/zoo/friesian/feature/table.py
hkvision: We can use partition id + row number for each window and make default seed to be None.

pyzoo/zoo/friesian/feature/table.py
hkvision: I think we can remove this. Target columns are supposed to be not None in theory.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1150,6 +1151,239 @@ def split(self, ratio, seed=None):
         tbl_list = [FeatureTable(df) for df in df_list]
         return tuple(tbl_list)
 
+    def gen_target(self, cat_cols, target_cols, target_mean=None, smooth=20, kfold=2,
+                   fold_seed=None, fold_col="__fold__", out_cols=None, name_sep="_"):
```
hkvision, line 1155: may not need the option name_sep?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1150,6 +1151,239 @@ def split(self, ratio, seed=None):
...
+                new_out_target_mean = {}
+                for out_col, target_mean in out_target_mean.items():
+                    if target_mean[0] not in target_cols:
+                        join_df = join_df.drop(out_col)
+                    else:
+                        new_out_target_mean[out_col] = target_mean
+                out_target_mean = new_out_target_mean
+
+            if not is_train or target_code.kfold == 1:
+                result_df = result_df.join(join_df, cat_col, how="left")
```
hkvision, line 1357: for the following operations, can we directly use join/fillna in FeatureTable? As these operations may be optimized in FeatureTable.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1237,3 +1471,82 @@ def write_parquet(self, path, mode="overwrite"):
...
+            column_mean = out_target_mean[column][1]
+            assert isinstance(column_mean, int) or isinstance(column_mean, float), \
+                "mean in target_mean should be numeric but get {} of type {}" \
+                " in {}".format(column_mean, type(column_mean), out_target_mean)
+
+    def _clone(self, df):
+        return TargetCode(df, self.all_df, self.cat_col, self.out_target_mean, self.kfold,
+                          self.fold_col)
+
+    def rename(self, columns):
```
hkvision, line 1529: is rename used?

hkvision:
```
def target_encode(self, columns, target_columns, target_mean=None, smooth=20, kfold=2, fold_seed=42,
            fold_col="__fold__", keep_fold=False, keep_original_columns=False, out_cols=None):
return new Target Encoded FeatureTable, a list of TargetCodes on full dataset only
# The TargetCodes on folds are abandoned since they won't be used any more?

# This will not include any fold information.
def encode_target(self, columns, targets, target_columns=None, keep_original_columns=False)
return new Target Encoded FeatureTable
```

And change your current encode_target to a utility function, which can handle both with and without fold. Then the above two methods can call this with different options.

hkvision:
`TargetCode` only needs df (target encodes on all data), cat_col, out_target_mean.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1150,6 +1151,239 @@ def split(self, ratio, seed=None):
         tbl_list = [FeatureTable(df) for df in df_list]
         return tuple(tbl_list)
 
+    def gen_target(self, cat_cols, target_cols, target_mean=None, smooth=20, kfold=2,
+                   fold_seed=None, fold_col="__fold__", out_cols=None, name_sep="_"):
+        """
+        For each categorical column / column group in cat_cols, calculate the mean of target
+        columns in target_cols.
+        :param cat_cols: str, or list of (str or list of str). Categorical columns / column groups
```
hkvision, line 1159: new line before this

pyzoo/zoo/friesian/feature/table.py
yaxinchen666: I refered to Nvidia's api https://nvidia.github.io/NVTabular/v0.3.0/api/ops/targetencoding.html

hkvision: We don't need to exactly follow their implementation.

yaxinchen666:
Jenkins passed:
http://10.112.231.51:18888/job/ZOO-PR-Validation/6150/

pyzoo/zoo/friesian/feature/table.py
hkvision: put rename here as it inherits Table.

hkvision:
This PR is large, merge this version first.

PR #4057, opened 2021-06-07
Add getattr and col for FeatureTable

hkvision:


hkvision:
http://10.239.176.111:18888/job/ZOO-PR-Validation/5743/
one failure is irrelevant. Merge it first.

PR #4065, opened 2021-06-08
Change return value of fit_eval and change default metric values in automl.BaseModel

yushan111:
This PR includes
- Change return value from `eval_result` value to dictionary of {`metric_name`: `eval_result`} in `fit_eval` for all `automl.BaseModel` inheritors 
- Change default `metric` value in `BaseKerasModel` from "mse" to `None`. If use leaves the `metric` in `fit_eval` as default, we will use the compiled metric name and value as evaluation result if user only used one metric to compile the model. We will raise error if user used multiple metrics in `model_creator` and didn't specify a specific target metric for `AutoEstimator` to optimize.
- Change default `metric` value in `BasePytorchModel` from "mse" to `None`.

### DIFF
--- a/pyzoo/test/zoo/automl/model/test_base_keras_model.py
+++ b/pyzoo/test/zoo/automl/model/test_base_keras_model.py
@@ -50,12 +50,12 @@ class TestBaseKerasModel(TestCase):
         model = modelBuilder_keras.build(config={
             "lr": 1e-2,
             "batch_size": 32,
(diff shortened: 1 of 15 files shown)


yushan111:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5777/
http://10.239.176.111:18888/job/ZOO-PR-Validation-Other/5808/

PR #4069, opened 2021-06-08
Fix xgboost metric

yushan111:
Fix bug: `metric` value passed through xgboost `fit_eval` doesn't take effect.


yushan111:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5751/

PR #4105, opened 2021-06-15
AutoTSTrainer with tuning of model hyperparams and data preprocessing

TheaperDeng:
This is a bug-fix version to #4102

- [x] built-in model tuning
- [x] built-in model + feature selection
 - user-defined model (next PR)

### What we can tune in AutoTSTrainer (currently).

1. Model Hyper Parameter (e.g. hidden_dim, ...) 
   1. Specific to each model
2. Model training Hyper Parameter (e.g. lr, batchsize, ...)
3. Data related Hyper Parameter
   1. Selected Feature ("selected_feature" in search_space, which will affect "input_feature_num")
   2. Lookback ("past_seq_len" in search_space)

### Why there is not any more hyper parameters can be tuned in TSDataset?
impute, deduplicate, resample and scale will affect the test set as well, which makes the comparison unfair.

###  Workload inside/outside AutoTSTrainer
impute, deduplicate, resample, scale and all feature generation should be completed by user themselves **before** sending the tsdataset to AutoTSTrainer. That is
```python
tsdata_train, tsdata_valid, _ = TSDataset.from_pandas(df, with_split=True, ...)
tsdata_train.impute().resample()...
tsdata_valid.impute().resample()...
autotstrainer.fit(data=tsdata_train, validation_data=tsdata_valid)
```

Reasons for this is listed as following
1. TSDataset API is exposed to users with clear document
2. If the workload of all these operation is completed before AutoTSTrainer, we don't need to do them again and again for each trail.
3. There is no need to wrap the TSDataset settings in AutoTSTrainer (e.g. the `preprocessing` parameter in AutoTSTrainer or even more complex design). That will be complex and less flexible (again, TSDataset has clear document).

### Search space for Data related Hyper Parameter
1. How to handle shape hyper parameters (e.g. "past_seq_len", "future_seq_len", "input_feature_num", "output_feature_num")
We should ask user to explicitly input them all. (Need discussion here)
"past_seq_len": can be tune, such as hp.randint(10,12)
"future_seq_len": fixed number, this should be provided by user since we can not infer it from our current data creator.
"input_feature_num": fixed number, this hp will be ignored if "selected_feature" is used in search space.
"output_feature_num": fixed number, this should be provided by user since we can not infer it from our current data creator.
2. How to handle selected_feature:
Maybe we need to set some special variable such as hp.SAMPLE_ALL_FEATURE ... (Need discussion here)


pyzoo/zoo/chronos/autots/experimental_trainer.py
```diff
@@ -0,0 +1,240 @@
+#
```
shane-huang, line 1: exprimental_trainer -> experimental (file name is too long now). Laster we can just replace the original forecast.

pyzoo/zoo/chronos/autots/experimental_trainer.py
```diff
@@ -0,0 +1,240 @@
...
+                             "loss": loss,
+                             "metric": metric,
+                             "optimizer": optimizer,
+                             "backend": backend,
+                             "logs_dir": logs_dir,
+                             "cpus_per_trial": cpus_per_trial,
+                             "name": name})
+
+        # create auto model from name
+        self.model = AutoModelFactory.create_auto_model(name=model,
```
shane-huang, line 103: This would assume all auto models use the same interface. Actually it is not garanteed, as they are not inherited from the same base. Anyway we can use it this way for now.

pyzoo/zoo/chronos/autots/experimental_trainer.py
```diff
@@ -0,0 +1,240 @@
...
+        :param val_data: validation data
+        :return: data creators from train and validation data
+        """
+        import torch
+        from torch.utils.data import TensorDataset, DataLoader
+        import ray
+
+        # append feature selection into search space
+        all_features = train_data.feature_col
+        if search_space.get('selected_features') is not None:
```
shane-huang, line 184: Remove this not None test and just check if the value of "selected_features" is in ("all", "auto")

pyzoo/zoo/chronos/autots/experimental_trainer.py
```diff
@@ -0,0 +1,240 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from zoo.chronos.data import TSDataset
+import zoo.orca.automl.hp as hp
+from zoo.chronos.autots.model import AutoModelFactory
+
+AUTOTS_DEFAULT_LOOKBACK = 5
```
shane-huang, line 21: Maybe set default lookback to 2? As I suppose there're at least 2 in an input sequence.

pyzoo/zoo/orca/automl/hp.py
```diff
@@ -109,6 +110,24 @@ def choice(categories):
...
 
+def choice_n(categories, min_items, max_items):
+    """
+    Sample a subset from a list
+    :param categories: A list to be sampled
+    :param min_items: minimum number of items to be sampled
+    :param max_items: maximum number of items to be sampled
+    """
+    return tune.sample_from(
+        lambda spec: list(
```
shane-huang, line 121: did shan said we could remove the list() ?

TheaperDeng: we need list here or the result would be a np.array, which is not supported by TSDataset.

pyzoo/test/zoo/chronos/autots/test_experimental_trainer.py
```diff
@@ -0,0 +1,161 @@
...
+                                     logs_dir="/tmp/auto_trainer",
+                                     cpus_per_trial=2,
+                                     name="auto_trainer")
+        auto_trainer.fit(data=get_tsdataset().gen_dt_feature(),
+                         epochs=1,
+                         batch_size=hp.choice([32, 64]),
+                         validation_data=get_tsdataset().gen_dt_feature(),
+                         n_sampling=1
+                         )
+        config = auto_trainer.get_best_config()
```
shane-huang, line 128: we may need a short full example that runs through the entire process to mimic the way our customer use it. When we're done fitting, we need to use the model for prediction. If we chose not to return a pipeline, we need to tell use how to reconstruct the pipeline using the configs and use it for prediction/evaluation.

pyzoo/zoo/chronos/autots/experimental_trainer.py
TheaperDeng: I think we can move the dirty preprocessing of different interface to AutoModelFactory so that we only need to change AutoModelFactory when we want to add another model rather than adding 10~15 lines to AutoTSTrainer.

pyzoo/test/zoo/chronos/autots/test_experimental_trainer.py
TheaperDeng: sure, I can add a **temp** example here. We can discuss the ppl to be developed in detail later (this may involve a historical operation record and a replay operation in TSDataset)

pyzoo/zoo/chronos/autots/experimental.py
```diff
@@ -0,0 +1,240 @@
...
+            search_space['selected_features'] = all_features
+
+        # put train/val data in ray
+        train_data_id = ray.put(train_data)
+        valid_data_id = ray.put(val_data)
+
+        def train_data_creator(config):
+            train_d = ray.get(train_data_id)
+
+            x, y = train_d.roll(lookback=config.get('past_seq_len', AUTOTS_DEFAULT_LOOKBACK),
```
shane-huang, line 201: If we allows default values to `past_seq_len` and `future_seq_len`, shall we somehow let user know what the default values we gave them? Also, if roll and model give different default values by using confi.get() to past_seq_len and future_seq, would it be a problem? So a better way is to just give default values at the AutoTSTrainer constructor, this way you don't need those constants. And just use config.get('past_seq_len') intead of get('past_seq_len', default)

pyzoo/zoo/chronos/autots/experimental_trainer.py
shane-huang: my only concern is that the way of using Model.fit maybe be different for different auto models as they don't share the same base model. And that is not part of the AutoModelFactory.

pyzoo/test/zoo/chronos/autots/test_experimental.py
```diff
@@ -0,0 +1,182 @@
...
+        tsdata_test.roll(lookback=best_config["past_seq_len"],
+                         horizon=1,
+                         feature_col=best_config["selected_features"])
+        x_test, y_test = tsdata_test.to_numpy()
+        y_pred = best_model.predict(x_test)
+        best_model.save("best.ckpt")
+        from zoo.automl.model.base_pytorch_model import PytorchModelBuilder
+        restore_model = PytorchModelBuilder(model_creator=best_model.model_creator,
+                                            optimizer_creator="Adam",
+                                            loss_creator=torch.nn.MSELoss()).build(best_config)
```
yushan111, line 146: Shall we add a method like `get_best_forecaster` later to make that easier or just return the best forecaster in `fit` or other options

TheaperDeng: I will propose a "pipeline" substitute later after this pr. We can not just provide a best forecaster since we might need to replay the tsdata operations as well.

TheaperDeng:
PR validation combine pass:
http://10.239.176.111:18888/job/ZOO-PR-Python-Spark-2.3-py36/5000/
http://10.239.176.111:18888/job/ZOO-PR-Validation/5877/

PR #4114, opened 2021-06-16
Add Ray doc for non-barrier mode

hkvision:
- Add barrier_mode in doc
- Remove pyarrow==0.17.0


hkvision:
https://ray-non-barrier.readthedocs.io/en/latest/doc/Ray/Overview/ray.html

Issue #4093, opened 2021-06-11
File NotFoundError for Wechat Baseline Model Using AutoSearch 


Elena-Qiu:
In the wechat baseline model using autosearch, I add the search space about crossed columns, where I implement as follows
```
def get_search_space():
    from zoo.orca.automl import hp
    cross_column_candidates = []
    cross_column_candidates.append(["userid", "bgm_singer_id", "bgm_song_id"])
    cross_column_candidates.append(["userid", "authorid"])
    cross_column_candidates.append(["userid", "feedid"])
    cross_column_candidates.append(["authorid", "bgm_singer_id", "bgm_song_id"])
    cross_column_candidates.append(["feedid", "bgm_singer_id", "bgm_song_id"])
    cross_column_candidates.append(["feedid", "authorid"])
    return {
        'embed_dim': hp.choice([20, 40, 80]),
        'embed_l2': None,
        'learning_rate': hp.choice([0.001, 0.01, 0.05, 0.1]),
        "batch_size": hp.choice([64, 128, 256]),
        "crossed_columns": hp.sample_from(lambda :
                                            list(np.random.choice(
                                                cross_column_candidates,
                                                size=np.random.randint(
                                                    low=3,
                                                    high=len(cross_column_candidates)),
                                                replace=False)))
    }
```

Then the filename of trails under auto_wnd seems too long and is not complete, it is like
```
/auto_wnd/wnd/train_func_a9181_00000_0_batch_size=256,crossed_columns=[['userid', 'bgm_singer_id', 'bgm_song_id'], ['userid', 'feedid'], ['feedi_2021-06-10_18-28-56
```

Then, I will get the following error:

```
INFO:tensorflow:Done calling model_fn.
2021-06-10 19:22:13,395 - INFO - Done calling model_fn.
INFO:tensorflow:Warm-starting with WarmStartSettings: WarmStartSettings(ckpt_to_initialize_from="/home/yuqing/repo/friesian/wechat2021/baseline/wideANDdeep/data/auto_wnd/wnd/train_func_a9181_00000_0_batch_size=256,crossed_columns=[['userid', 'bgm_singer_id', 'bgm_song_id'], ['userid', 'feedid'], ['feedi_2021-06-10_18-28-56/./data/model_ckpt/offline_train/read_comment/model.ckpt-2422", vars_to_warm_start='.*', var_name_to_vocab_info={}, var_name_to_prev_var_name={})
[...]

PR #4116, opened 2021-06-16
change default log dir name for searcher

yushan111:


yushan111:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5803/

yushan111:
The default directory name of ray tune contains configs in `search_space`. If the searched values is a list, it may be hard to identify the directory name for several applications and cause error during restore.

For tensorflow,

- `tf.train.Checkpoint()` can not deal with "[" for model path. [Related issue](https://github.com/tensorflow/tensorflow/issues/32614)

- `tf.train.saver()`  [Related issue](https://github.com/tensorflow/tensorflow/issues/6082)
- `tf.estimator(warm_start_from='path_without_[]')`, issue #4093 

Also
hdfs get doesn't support unbalanced number of "[" and "]"

PR #4117, opened 2021-06-17
Modify master and worker start scripts to support bigdl example

Zesheng16:
- Modify master and worker start scripts to support bigdl example


glorysdj (approved):
LGTM

Issue #4083, opened 2021-06-10
Exception happened when running the example notebook in latest zoo image


TheaperDeng:
same issue as #3428 happens in intelanalytics/analytics-zoo as well

glorysdj:
need to update dockerfile of hyperzoo and zoo to install jupyterlab
 sudo -H pip install jupyterlab -U

PR #4118, opened 2021-06-17
reslove version conflict between tornado and jupyter

zzti-bsj:
fixed [4083 issue](https://github.com/intel-analytics/analytics-zoo/issues/4083), version conflict between tornado==5.1.1 and juypter notebook was resloved


glorysdj (approved):
LGTM

PR #4121, opened 2021-06-18
add bigdl lenet example to README

Zesheng16:
- add bigdl lenet example to README
- replace secret key in command line with "your_secret_key"


glorysdj (approved):
LGTM

PR #4145, opened 2021-06-24
Add rank and size  in config for training and validation data creator in TFRunner

jenniew:
Add rank and size in config for data creator to use


yangw1234:
generally LGTM, however maybe we should add this to the configs in all other estimators (or create an issue to as a reminder)

jenniew:
Add https://github.com/intel-analytics/analytics-zoo/issues/4231 to remind

jenniew:
Jenkins: http://10.239.176.111:18888/job/ZOO-PR-Validation/6049/

PR #4140, opened 2021-06-23
Revised quick start for forecasting (TSDataset + Forecaster / AutoTS)

TheaperDeng:
- [x] TCN problem #4136 (work around)
- [x] Setup problem #4137 #4153
- [x] _unscale_numpy -> unscale_numpy #4144  (jenkins)
- [x] split to two notebooks (one for TSDataset + Forecaster, one for AutoTS)
- eliminate `exit()` (will be fixed later)

preview: https://junweitestdoc.readthedocs.io/en/quick-start-revise-forecasting/doc/Chronos/QuickStart/chronos-tsdataset-forecaster-quickstart.html

Issue #4137, opened 2021-06-22
analytics-zoo[automl] nightly built setup issue on google colab


TheaperDeng:
In theory, `pip install --pre --upgrade analytics-zoo[automl]` should setup all the dependencies for nightly built version for Chronos without any issue.

On google colab, we still need to `!pip install scipy==1.5 aioredis==1.3.1` to solve version confliction.

colab has default scipy==1.4.1 which is incompatible with tsfresh. aioredis=2.0 is incompatible with ray dashboard.

TheaperDeng:
wait for validation after nightly built

PR #4153, opened 2021-06-25
Change versions of scipy and aioredis in automl extra require

TheaperDeng:
related to issue #4137 and pr #4140


TheaperDeng:
PR validation combine pass:
http://10.239.176.111:18888/job/ZOO-PR-Validation/5918/
http://10.239.176.111:18888/job/ZOO-PR-Python-Spark-2.3-py36/5044/

PR #4169, opened 2021-06-28
Always keep driver core as 0 when start ray on driver in cluster mode

jenniew:
If start ray on driver with more than 0 cores, the TFRunner or other actors would run on driver which is not expected.
So not pass driver cores values when calling ray_context.init()

### DIFF
--- a/pyzoo/zoo/orca/common.py
+++ b/pyzoo/zoo/orca/common.py
@@ -262,8 +262,6 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
     ray_ctx = RayContext(sc, **ray_args)
     if init_ray_on_spark:
         driver_cores = 0  # This is the default value.
-        if "driver_cores" in kwargs:
-            driver_cores = kwargs["driver_cores"]
         ray_ctx.init(driver_cores=driver_cores)
     return sc
 


pyzoo/zoo/orca/common.py
```diff
@@ -262,8 +262,6 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
     ray_ctx = RayContext(sc, **ray_args)
     if init_ray_on_spark:
         driver_cores = 0  # This is the default value.
-        if "driver_cores" in kwargs:
```
hkvision, line 265: instead of removing it, can we change to `ray_driver_cores` to differ with spark?

jenniew: Do we need to need to provide option to set driver cores if start ray on driver? I think we need to hardcode to 0 to avoid actors running on driver.

hkvision: Sometimes if the driver has the capability to run actors, we can?

jenniew: I think running actors on driver is a little strange, as in general in cluster environment, the tasks are often running on executors not driver. In addition, we may run multiple jobs on the same driver and all the jobs would share the resources that may cause problem. In that case, ray_driver_cores  should be less than or equal to spark driver_cores.

hkvision (approved):

jenniew:
Jenkins: http://10.239.176.111:18888/job/ZOO-PR-Validation/6058/

PR #4189, opened 2021-06-30
remove max_length calculation in orca estimators

### LINKED ISSUE #4176: TF2Estimator fit DataFrame need to do repartition before converting to SparkXShards?
https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/learn/tf2/estimator.py#L221

The original logic is: if we get a Spark DataFrame, for each partition, we convert into a dict and we get a SparkXShards. Then before converting to RayXShards, we do repartition on the number of workers.

But if the original Spark DataFrame only has 1 partition, then after converting to SparkXShards, there is only one dict element. So if I launch two workers on one node, after repartition(2), there will be an empty partition for a worker, which will cause index error when retrieving data. Even I set OrcaContext._shard_size to be small, there a more than one dict elements, I still get an empty partition.

In my case, the val data only has 1000 records with 1 partition and I get the error when training on a single node with two workers.

We'd better put repartition at the very beginning before converting to SparkXShards to avoid this issue?

yangw1234:
fix #4176

### DIFF
--- a/pyzoo/zoo/orca/learn/tf2/estimator.py
+++ b/pyzoo/zoo/orca/learn/tf2/estimator.py
@@ -82,13 +82,8 @@ def process_spark_xshards(spark_xshards, num_workers):
     data = spark_xshards
     if data.num_partitions() != num_workers:
         data = data.repartition(num_workers)
-
-    # todo currently we need this information to pad the short partitions
-    # so that every model run exactly the same number of steps in one epoch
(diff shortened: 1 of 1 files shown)


yangw1234:
does it work for you @hkvision

yangw1234:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5975/

Issue #4190, opened 2021-06-30
Problem with import Estimator from zoo.orca.learn.tf.estimator


cs201503:
Hi, 
I am trying out orca, and encountered some problems during the first step:
![image](https://user-images.githubusercontent.com/78426877/124010547-39778280-d9df-11eb-8d04-f48cc2350d24.png)

I am using Python 3.6.13.
Anyone know how I can solve the problem? Please help!
Thank you very much in advance!

yangw1234:
hi @cs201503 thank you for reporting this issue but I was not able to reproduce it on our side. Would you mind sharing more details about your environment,
e.g.
1. what's the version of analytics zoo?
2. what's the version of tensorflow?

yushan111:
@cs201503 
Thanks for your issue.

Your tensorflow version might be 2.x and you could try tensorflow==1.15.0 to use orca tf estimator. And we have orca tf2 estimator for tensorflow2.x. You can import with `from zoo.orca.learn.tf2 import Estimator` and find more details in [tf2 estimator API guide](https://analytics-zoo.readthedocs.io/en/latest/doc/PythonAPI/Orca/orca.html#module-zoo.orca.learn.tf2.estimator) and [tf2 estimator quick start](https://analytics-zoo.readthedocs.io/en/latest/doc/Orca/QuickStart/orca-tf2keras-quickstart.html)

cs201503:
> @cs201503
> Thanks for your issue.
> 
> Your tensorflow version might be 2.x and you could try tensorflow==1.15.0 to use orca tf estimator. And we have orca tf2 estimator for tensorflow2.x. You can import with `from zoo.orca.learn.tf2 import Estimator` and find more details in [tf2 estimator API guide](https://analytics-zoo.readthedocs.io/en/latest/doc/PythonAPI/Orca/orca.html#module-zoo.orca.learn.tf2.estimator) and [tf2 estimator quick start](https://analytics-zoo.readthedocs.io/en/latest/doc/Orca/QuickStart/orca-tf2keras-quickstart.html)

Hi @yushan111 , thank you very much for your help!
I have created a new environment and installed analytics zoo again according to TF2 instructions (https://analytics-zoo.readthedocs.io/en/latest/doc/Orca/QuickStart/orca-tf2keras-quickstart.html).
I have Python 3.7.0, Tensorflow 2.3.0.
When I am running the 2nd step: init_orca_context(cluster_mode="local", cores=4, memory="2g")
[...]

PR #4192, opened 2021-07-01
fix py36 tf version checking issue

yangw1234:
relate #4190


yangw1234:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/5975/

PR #4208, opened 2021-07-02
Add frontend base64 image input support

EvelynQiang:


EvelynQiang:
@Litchilitchy jenkin passed.

zoo/src/main/scala/com/intel/analytics/zoo/serving/preprocessing/PreProcessing.scala
```diff
@@ -62,18 +62,31 @@ class PreProcessing()
...
+          })
+      }
+      Seq(table)
     })
   }
 
   def decodeArrowBase64(key: String, s: String, serde: String = ""): Activity = {
     try {
 
-      val instance = if (serde == "stream") {
```
Litchilitchy, line 72: These kinds of if-else are not good. Could add the logics into `JsonInputDeserializer`. The input is a json string the output is the Tensor.

The image input does not always follows the pattern `image:` so it is better to move the logics into `JsonInputDeserializer`. In its implementation, you could consider all string inputs are image base64 string.

EvelynQiang: Updated

Litchilitchy (approved):

PR #4228, opened 2021-07-06
[WIP] Change the data creation method of test_dataset

liangs6212:
add function get_data and class tsdataset.


TheaperDeng:
We can revise our test file after all tests are added. So we will hold this PR for a while

PR #4237, opened 2021-07-07
Check conda pack status, throw error while return status is not 0

yangw1234:


yangw1234:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/6039/

PR #4248, opened 2021-07-09
Change gen_string_index output to a single StringIndex if there's only one column

hkvision:


hkvision:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/6061/

Issue #4256, opened 2021-07-12
jenkins: AttributeError: 'Prophet' object has no attribute 'model'


pinggao187:
branch-0.11.0-RC1 test
error: 
~~~
#25 start app test for chronos-stock-prediction-prophet
AttributeError: 'Prophet' object has no attribute 'model'
~~~
but prophet 1.0.1 has been installed on the node.

jenkins link:
http://10.239.176.111:18888/job/ZOO-NB-Pip-AppTests-part4/482/console
http://10.239.176.111:18888/job/ZOO-NB-Pip-AppTests-MAC-py36/443/console

pinggao187:
@PeterQiu0516  pls have a look

TheaperDeng:
This bug may related to pr #4226, while at that time @PeterQiu0516 run pass http://10.239.176.111:18888/job/ZOO-PR-Pip-Apps-Integration-Part4/105/ Is there any difference between these tests?

PeterQiu0516:
This problem results from the change of API for `AutoEstimator.get_best_model()` here: https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/automl/auto_estimator.py#L199. 

For PR that includes changes of API which come with notebook examples, we should run app test in addition to the Zoo-PR-Validation as well.

pinggao187:
fixed
http://10.239.176.111:18888/view/ZOO-NB/job/ZOO-NB-Pip-AppTests-part4/483/

PR #4260, opened 2021-07-12
Fix App Test Error from AutoEstimator API Change

PeterQiu0516:
Fix App Test Error from AutoEstimator API Change.
Resolve Issue #4256.


TheaperDeng:
app test fix
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Pip-Apps-Integration-Part4/107/

PR #4317, opened 2021-07-16
add orca tf text example to python graphene and ppml doc

Zesheng16:


ppml/trusted-big-data-ml/python/docker-graphene/start-distributed-spark-driver.sh
```diff
@@ -23,8 +23,8 @@ ssh root@$MASTER "docker run -itd \
       -v $ZOO_PATH:/ppml/trusted-big-data-ml/work/zoo \
       --name=spark-driver \
       -e LOCAL_IP=$MASTER \
-      -e SGX_MEM_SIZE=32G \
+      -e SGX_MEM_SIZE=64G \
```
glorysdj, line 26: this is only changed for orca tf text example, so we'd better add doc to ask user to modify this when user need to run orca text on spark standalone

glorysdj (approved):
LGTM

PR #4318, opened 2021-07-16
Fix datatype when creating stringindex from dict

hkvision:
Default to be Long, change to int


hkvision:
Tested on recsys preprocess.

hkvision:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/6094/

PR #4329, opened 2021-07-20
support pyspark.sql.types.StringType column

nyamashi:
This modification has been tested using a tf2/keras  model through Orca TF2 APIs.


hkvision:
Thanks so much for your contribution. We will test an example for string input soon. Stay tuned.

PR #4331, opened 2021-07-20
[Fix #4328]Repartition validation data for zip reduce RayXShards

### LINKED ISSUE #4328: RayXShards zip reduce shouldn't assume the same number of partitions for train and test data
https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/orca/data/ray_xshards.py#L232

Currently logic is:
- If train data partitions < num workers, we repartition to the number of workers. Otherwise the partition number will be unchanged.
- So if originally train rdd and test rdd have different partitions larger than the number of workers, the error "the rdds to be zipped must have the same number of partitions" will always occur.

We need to handle the partition issue by ourselves in our estimator.

yushan111:
This PR is to fix #4328

### DIFF
--- a/pyzoo/test/zoo/orca/learn/ray/tf/test_tf_ray_estimator.py
+++ b/pyzoo/test/zoo/orca/learn/ray/tf/test_tf_ray_estimator.py
@@ -457,6 +457,7 @@ class TestTFRayEstimator(TestCase):
         assert df.rdd.getNumPartitions() < trainer.num_workers
 
         trainer.fit(df, epochs=1, batch_size=4, steps_per_epoch=25,
+                    validation_data=df, validation_steps=1,
                     feature_cols=["feature"],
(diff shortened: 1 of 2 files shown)


yushan111:
@hkvision  
I repartition the `validatition_data` to the same partition number as `data`
Does that work for your issue?

hkvision:
I think it is ok. But will repartition be expensive? Are there any better ways?

yangw1234:
LGTM

yangw1234 (approved):

yushan111:
> I think it is ok. But will repartition be expensive? Are there any better ways?

According to current design, repartition might be the most convenient way. We could enhance that in later PRs.

yushan111:
http://10.239.176.111:18888/view/ZOO-PR/job/ZOO-PR-Validation/6127/

Issue #4330, opened 2021-07-20
jenkins: apptest failed


pinggao187:
#20 start app test for chronos-anomaly-detect-unsupervised-forecast-based
~~~
File "/var/jenkins_home/workspace/ZOO-NB-Pip-AppTests-MAC-py36/dist/../pyzoo/zoo/chronos/use-case/AIOps/AIOps_anomaly_detect_unsupervised_forecast_based.py", line 113, in <module>
    tsdata.resample(interval='1min', merge_mode="mean")          .impute(mode="last")          .gen_dt_feature()          .scale(standard_scaler, fit=(tsdata is tsdata_train))          .roll(lookback=look_back, horizon=horizon, feature_col = ["HOUR(time_step)"])
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/zoo/chronos/data/tsdataset.py", line 451, in roll
    .apply(lambda df: roll_timeseries_dataframe(df=df,
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 859, in apply
    result = self._python_apply_general(f, self._selected_obj)
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 892, in _python_apply_general
    keys, values, mutated = self.grouper.apply(f, data, self.axis)
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/pandas/core/groupby/ops.py", line 220, in apply
    res = f(group)
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/zoo/chronos/data/tsdataset.py", line 456, in <lambda>
    target_col=target_col))
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/zoo/chronos/data/utils/roll.py", line 63, in roll_timeseries_dataframe
    target_col)
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/zoo/chronos/data/utils/roll.py", line 108, in _roll_timeseries_dataframe_train
    x = df[:-max_horizon].loc[:, target_col+feature_col].values
  File "/Users/arda/anaconda3/envs/py36pip/lib/python3.6/site-packages/pandas/core/indexing.py", line 873, in __getitem__
    return self._getitem_tuple(key)
[...]

PR #4332, opened 2021-07-20
Chronos: fix broken notebook

TheaperDeng:
will run app test 4 before merging.

related to issue #4330


TheaperDeng:
app test pass:
http://10.239.176.111:18888/job/ZOO-PR-Pip-Apps-Integration-Part4/110/

PR #4336, opened 2021-07-20
Add string type option in convert_row_to_numpy

hkvision:
Logically, only tf2 estimator and pytorch ray estimator can support string input. BigDL backends use Sample and string is not supported.

### DIFF
--- a/pyzoo/zoo/orca/learn/openvino/estimator.py
+++ b/pyzoo/zoo/orca/learn/openvino/estimator.py
@@ -111,7 +111,8 @@ class OpenvinoEstimator(SparkEstimator):
                                               validation_data=None,
                                               feature_cols=feature_cols,
                                               label_cols=None,
-                                              mode="predict")
+                                              mode="predict",
+                                              accept_str_col=False)
             transformed_data = xshards.transform_shard(predict_transform, self.batch_size)
             result_rdd = self.model.distributed_predict(transformed_data.rdd, sc)
             return convert_predict_rdd_to_dataframe(data, result_rdd.flatMap(lambda data: data))
--- a/pyzoo/zoo/orca/learn/openvino/estimator.py
+++ b/pyzoo/zoo/orca/learn/openvino/estimator.py
@@ -111,8 +111,7 @@ class OpenvinoEstimator(SparkEstimator):
                                               validation_data=None,
                                               feature_cols=feature_cols,
                                               label_cols=None,
-                                              mode="predict",
-                                              accept_str_col=False)
+                                              mode="predict")
             transformed_data = xshards.transform_shard(predict_transform, self.batch_size)
             result_rdd = self.model.distributed_predict(transformed_data.rdd, sc)
             return convert_predict_rdd_to_dataframe(data, result_rdd.flatMap(lambda data: data))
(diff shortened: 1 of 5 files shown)


pyzoo/zoo/orca/learn/openvino/estimator.py
```diff
@@ -111,7 +111,8 @@ def predict_transform(dict_data, batch_size):
                                               validation_data=None,
                                               feature_cols=feature_cols,
                                               label_cols=None,
-                                              mode="predict")
+                                              mode="predict",
+                                              accept_str_col=False)
```
jason-dai, line 115: why do we need this?

hkvision: OpenVINO doesn't support string, no?

jason-dai: it defaults to `false`, no?

pyzoo/zoo/orca/learn/utils.py
```diff
@@ -279,7 +280,7 @@ def _dataframe_to_xshards(data, feature_cols, label_cols=None):
 
 
 def dataframe_to_xshards(data, validation_data, feature_cols, label_cols, mode="fit",
-                         num_workers=None):
+                         num_workers=None, accept_str_col=True):
```
jason-dai, line 283: we should make it default to `false`, and only make it `true` for `TF2RayEstimator`

pyzoo/zoo/orca/learn/utils.py
```diff
@@ -263,14 +263,15 @@ def process_xshards_of_pandas_dataframe(data, feature_cols, label_cols=None, val
         return data
 
 
-def _dataframe_to_xshards(data, feature_cols, label_cols=None):
+def _dataframe_to_xshards(data, feature_cols, label_cols=None, accept_str_col=True):
```
jason-dai, line 266: default to `false`

pyzoo/zoo/orca/learn/utils.py
hkvision: oh really, but as we discussed, dataframe to xshards logically can support any datatype so I think it is more reasonable to make default True?

hkvision:
Added a dummy unit test @jason-dai

hkvision:
http://10.239.176.111:18888/job/ZOO-PR-Python-Spark-2.4-py37-Horovod/2021/console

pyzoo/zoo/orca/learn/utils.py
jason-dai: I think we should only support string type for `TF2RayEstimator` at this moment; it makes sense to set it to false for all the other cases.

jason-dai:
> Added a dummy unit test @jason-dai

It only tests the positive case, not the negative case (e.g., passing string for tf1 estimator); I think you should keep the original behavior (i.e., string not supported) for all cases except the tf2 estimaror

pyzoo/zoo/orca/learn/utils.py
hkvision: okay.

jason-dai:
LGTM

hkvision:
jenkins passed: http://10.239.176.111:18888/job/ZOO-PR-Validation/6117/

PR #4339, opened 2021-07-21
Update init_orca_context

sgwhat:


pyzoo/zoo/orca/common.py
```diff
@@ -190,22 +190,37 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
     print("Initializing orca context")
     import atexit
     atexit.register(stop_orca_context)
-    cluster_mode = cluster_mode.lower()
+    from pyspark import SparkContext, SparkConf
+    conf = SparkConf()
```
hkvision, line 194: why create a conf by yourself? You need to take the conf from sc if it is active

pyzoo/zoo/orca/common.py
```diff
@@ -190,22 +190,37 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
     import atexit
     atexit.register(stop_orca_context)
-    cluster_mode = cluster_mode.lower()
+    from pyspark import SparkContext
+    activate_sc = SparkContext._active_spark_context is not None
+    if cluster_mode is not None:
+        cluster_mode = cluster_mode.lower()
+        if activate_sc:
+            sc = SparkContext.getOrCreate()
+            cores = sc._conf.get("spark.executor.cores") if cores is None else cores
```
hkvision, line 199: change _conf to getConf()?

pyzoo/zoo/orca/common.py
```diff
@@ -190,22 +190,37 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
     print("Initializing orca context")
     import atexit
     atexit.register(stop_orca_context)
-    cluster_mode = cluster_mode.lower()
+    from pyspark import SparkContext
+    activate_sc = SparkContext._active_spark_context is not None
+    if cluster_mode is not None:
```
hkvision, line 195: just `if cluster_mode`?

pyzoo/zoo/orca/common.py
```diff
@@ -190,22 +190,37 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
+        cluster_mode = cluster_mode.lower()
+        if activate_sc:
+            sc = SparkContext.getOrCreate()
+            cores = sc._conf.get("spark.executor.cores") if cores is None else cores
+            memory = sc._conf.get("spark.executor.memory") if memory is None else memory
+            num_nodes = sc._conf.get("spark.executor.instances") if num_nodes is None else num_nodes
```
hkvision, line 201: Does it make sense to specify cores/memory is sc is not None? They won't take any effect right? I think it is better that we just check if cores/memory/num_nodes are not None, they whether they match the actual conf in sc?

pyzoo/zoo/orca/common.py
sgwhat: yes, when cluster_mode is not None,  which is also not local mode, then we could make sure if there is activate sc.

pyzoo/zoo/orca/common.py
```diff
@@ -158,7 +158,7 @@ def get_ray_context():
         return RayContext.get()
 
 
-def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
+def init_orca_context(cluster_mode=None, cores=None, memory=None, num_nodes=None,
```
jason-dai, line 161: I think we can keep the original default values for `cores`, `memory` and `num_nodes`, and report a warning when Spark context already exists?

pyzoo/zoo/orca/common.py
```diff
@@ -190,21 +190,40 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
     atexit.register(stop_orca_context)
-    cluster_mode = cluster_mode.lower()
+    from pyspark import SparkContext
+    activate_sc = SparkContext._active_spark_context is not None
+    if cluster_mode is not None:
+        cluster_mode = cluster_mode.lower()
+        if cluster_mode != "local" and not activate_sc:
+                cores = 4 if cores is None else cores
+                memory = "2g" if memory is None else memory
+                num_nodes = 2 if num_nodes is None else num_nodes
```
jason-dai, line 200: better to keep these in the default parameters

sgwhat: Sure, I will keep them in the default parameters and add warnning report.

pyzoo/zoo/orca/common.py
```diff
@@ -190,21 +190,40 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
+        elif cluster_mode != "local" and activate_sc:
+                scc = SparkContext.getOrCreate()
+                cores = scc.getConf().get("spark.executor.cores")
+                memory = scc.getConf().get("spark.executor.memory")
+                num_nodes = scc.getConf().get("spark.executor.instances")
```
jason-dai, line 205: I don't think this works for `cores` and `num_nodes`, as difference cluster model (local, standalone, yarn, etc.) will use different ways to specify these values?

sgwhat: yeah...that's a bad idea. Sorry I didn't consider this stuff like standalone doesn't call executor, I think I should better follow the original method.

pyzoo/zoo/orca/common.py
```diff
@@ -190,21 +190,40 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
+        num_nodes = 1 if num_nodes is None else num_nodes
         assert num_nodes == 1, "For Spark local mode, num_nodes should be 1"
         os.environ["SPARK_DRIVER_MEMORY"] = memory
         if "python_location" in kwargs:
             spark_args["python_location"] = kwargs["python_location"]
         from zoo import init_spark_on_local
         sc = init_spark_on_local(cores, **spark_args)
+    elif cluster_mode == "spark-submit":
+        from zoo import init_nncontext
+        sc = init_nncontext(**spark_args)
```
jason-dai, line 226: I think the logic above is too complex; maybe simply
```python
if activate_sc
   report warning if cluster_mode is not None or "spark_submit"
   #and also deprecate "spark_submit"
   init_nncontext
else
   cluster_model = "local" if cluster_model is None
   #repot error if cluster_model is "spark_submit"?
   then follow the original code path
```

sgwhat: Sure, I will modify the logic soon.

sgwhat: `repot error if cluster_mode is "spark_submit"` if not activate_sc, are we going to remove this mode whether activate_sc or not?

pyzoo/zoo/orca/common.py
```diff
@@ -190,69 +190,81 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
+        if cluster_mode == "local":
+            assert num_nodes == 1, "For Spark local mode, num_nodes should be 1"
+            os.environ["SPARK_DRIVER_MEMORY"] = memory
+            if "python_location" in kwargs:
+                spark_args["python_location"] = kwargs["python_location"]
+            from zoo import init_spark_on_local
+            sc = init_spark_on_local(cores, **spark_args) 
+        elif cluster_mode == "spark-submit":
+            from zoo import init_nncontext
+            sc = init_nncontext(**spark_args)
```
hkvision, line 219: why spark-submit is in this branch? It should be in activate sc branch, no?

pyzoo/zoo/orca/common.py
```diff
@@ -190,69 +190,81 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
+                                'and submit the application via spark-submit instead')
+            assert "master" in kwargs, "Please specify master for k8s-client mode"
+            assert "container_image" in kwargs, "Please specify container_image for k8s-client mode"
+            for key in ["driver_cores", "driver_memory", "extra_executor_memory_for_ray",
+                        "extra_python_lib", "jars", "python_location"]:
+                if key in kwargs:
+                    spark_args[key] = kwargs[key]
+            from zoo import init_spark_on_k8s
+            sc = init_spark_on_k8s(master=kwargs["master"],
+                                container_image=kwargs["container_image"],
```
hkvision, line 254: style... keep aligned

pyzoo/zoo/orca/common.py
```diff
@@ -190,69 +190,81 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
     for key in ["conf", "spark_log_level", "redirect_spark_log"]:
         if key in kwargs:
             spark_args[key] = kwargs[key]
-    if cluster_mode == "spark-submit":
+    if cluster_mode is not None:
+        cluster_mode = cluster_mode.lower()
+    activate_sc = SparkContext._active_spark_context is not None
+    if activate_sc:
+        if cluster_mode is not None:
+            warnings.warn("Using an Existing SparkContext, cluster_mode is useless now", Warning)
```
hkvision, line 204: Existing -> existing
cluster_mode is determined by the existing SparkContext

pyzoo/zoo/orca/common.py
```diff
@@ -190,69 +191,80 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
     for key in ["conf", "spark_log_level", "redirect_spark_log"]:
         if key in kwargs:
             spark_args[key] = kwargs[key]
-    if cluster_mode == "spark-submit":
+    if cluster_mode is not None:
+        cluster_mode = cluster_mode.lower()
+    activate_sc = SparkContext._active_spark_context is not None
+    if activate_sc:
+        if cluster_mode is not None:
+            warnings.warn("cluster_mode is determined by the existing SparkContext", Warning)
```
hkvision, line 205: Use an existing SparkContext, cluster_mode is determined by the existing SparkContext

pyzoo/zoo/orca/common.py
```diff
@@ -190,69 +191,80 @@ def init_orca_context(cluster_mode="local", cores=2, memory="2g", num_nodes=1,
...
+        conf = SparkConf()
```
hkvision, line 206: will this conf take effect?

sgwhat: yes, already passed test.

Issue #4327, opened 2021-07-19
min_max_scale should transform data given min and max values


hkvision:
After obtaining min/max values from train dataset, we need to use these values to transform the test dataset.
Need to add something like:
```
for column, (col_min, col_max) in min_max_dic.items():
    normalize = lambda x: (x-col_min)/(col_max-col_min)
    test_tbl = test_tbl.apply(column, column, normalize, "float")
```

PR #4343, opened 2021-07-21
add transform min_max scale

jenniew:
Apply min max scale with provide min, max value.
Fix https://github.com/intel-analytics/analytics-zoo/issues/4327


hkvision:
What's the status of this PR?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +929,58 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
 
         return FeatureTable(df), min_max_dic
 
+    def transform_min_max_scale(self, columns, min_max_dic):
+        """
+        Rescale each column individually with [min, max] range.
+
+        :param columns: a column name or a list of column names
```
hkvision, line 936: str or a list of str. The column(s) to be rescaled.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +929,58 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
...
+        :param min_max_dic: mapping of (min, max) for each column
```
hkvision, line 937: dict, the key is the column name, and the value is a min, max tuple of  this column.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +929,58 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
...
+        :return: Rescaled FeatureTable
```
hkvision, line 938: A new FeatureTable with rescaled column(s).

hkvision:
@jenniew Also polish the doc of `min_max_scale`

pyzoo/test/zoo/friesian/feature/test_table.py
```diff
@@ -338,8 +338,14 @@ def test_norm(self):
...
         tbl2 = FeatureTable(feature_tbl.df.withColumn("col2-col3", array(["col_2", "col_3"])))
-        normalized_tbl2, min_max = tbl2.min_max_scale(["col_2", "col2-col3"])
-        normalized_tbl2.compute()
+        normalized_tbl2, min_max_dic2 = tbl2.min_max_scale(["col_2", "col2-col3"])
+
+        test_file_path = os.path.join(self.resource_path, "friesian/feature/parquet/data3.parquet")
+        test_tbl = FeatureTable.read_parquet(test_file_path).fillna(0, ["col_2", "col_3"])
+        scaled = test_tbl.transform_min_max_scale(["col_2"], min_max_dic)
+        test_tbl2 = FeatureTable(test_tbl.df.withColumn("col2-col3", array(["col_2", "col_3"])))
+        scaled2 = test_tbl2.transform_min_max_scale(["col_2", "col2-col3"], min_max_dic2)
```
hkvision, line 347: Add some test to ensure the result value is correct?

pyzoo/zoo/friesian/feature/table.py
jenniew: updated

pyzoo/test/zoo/friesian/feature/test_table.py
jenniew: updated.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -865,7 +865,8 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
         :param min: Lower bound after transformation, shared by all columns. 0.0 by default.
         :param max: Upper bound after transformation, shared by all columns. 1.0 by default.
 
-        :return: FeatureTable and mapping = {c: (originalMin, originalMax) for c in columns}
+        :return: new FeatureTable with scaled columns
```
hkvision, line 868: A tuple of a new FeatureTable .... and the mapping ...

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +930,61 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
 
         return FeatureTable(df), min_max_dic
 
+    def transform_min_max_scale(self, columns, min_max_dic):
+        """
+        Rescale each column individually with given [min, max] range of each column.
+
+        :param columns: a str or a list of str. The column(s) to be rescaled.
```
hkvision, line 937: str (not **a** str, to keep the same as elsewhere)

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +930,61 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
...
+        :param min_max_dic: a dictionary of min, max values of each column.
+         The key is the column name, and the value is (min, max) of this column.
```
hkvision, line 939: `The key ...` Keep align with `min_max_dict`

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +930,61 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
 
         return FeatureTable(df), min_max_dic
 
+    def transform_min_max_scale(self, columns, min_max_dic):
```
hkvision, line 933: min_max_dic -> min_max_dict

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +930,61 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
...
+        """
+        Rescale each column individually with given [min, max] range of each column.
```
hkvision, line 935: within the given

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -865,7 +865,8 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
         :param min: Lower bound after transformation, shared by all columns. 0.0 by default.
```
hkvision, line 865: Line 864, do we support columns a single list? If yes, change the doc to `str or a list of str. The column(s) to be rescaled.` (same as transform_min_max_scale)?

hkvision: Change to: int, lower/upper ... Default to be 0.

hkvision: Line 862, remove the space before `column`

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +930,61 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
 
         return FeatureTable(df), min_max_dic
```
jason-dai, line 931: It is better to only return a `FeatureTable(df)` here, and pass in `min_max_dic` as a function parameter? Then it's possible to write something like:
```
tbl.min_max_scale(...).cross_columns(...)
```

hkvision: But min_max_dict is the meta data of this FeatureTable, it is calculated and how to pass it as a function parameter?

jason-dai: the user need to pass an empty dictionary, and then method can update the dictionary

hkvision: oh, I think it is a bit strange doing this? 0 0

jason-dai: > oh, I think it is a bit strange doing this? 0 0

Maybe, but we also want to make all operations consistent (i.e., returning the updated table so that the operations can be chained in sequence)

hkvision: I have created an issue for this: https://github.com/intel-analytics/analytics-zoo/issues/4403
Currently several other operations return a tuple as well. We can discuss and think of a good way to handle this later.

hkvision: But actually for complicated preprocessing, it may be reasonable for users to split the steps instead of chaining everything into a single line?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -929,6 +930,61 @@ def min_max_scale(self, columns, min=0.0, max=1.0):
...
 
+    def transform_min_max_scale(self, columns, min_max_dic):
+        """
+        Rescale each column individually with given [min, max] range of each column.
+
+        :param columns: a str or a list of str. The column(s) to be rescaled.
+        :param min_max_dic: a dictionary of min, max values of each column.
+         The key is the column name, and the value is (min, max) of this column.
+        :return: A new FeatureTable with rescaled column(s).
```
hkvision, line 940: new line before return

jenniew:
Jenkins: http://10.112.231.51:18888/job/ZOO-PR-Validation/6226/

PR #4355, opened 2021-07-23
add how to run user's own pyspark program to ppml doc

Zesheng16:
- add how to run user's own pyspark program to ppml doc

This PR is to add the instructions for user to run his self-developed pyspark program.
We will also add similar instructions to scala/java part.
It's tested with a sample pyspark program.


jason-dai:
LGTM

PR #4360, opened 2021-07-23
Grpc serving

Gavin-Sun555:
Add frontend GRPC Serving. 
DIEN model roughly 28ms with inference model on local computer.

Performance Metrics
(with env var TF_DISABLE_MKL=1)
**On Clx 15**
testing...


zoo/pom.xml
```diff
@@ -96,9 +101,9 @@
 
         <profile>
             <id>spark_2.4+</id>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
+            <!--<activation>-->
```
glorysdj, line 104: this should be reverted

zoo/pom.xml
```diff
@@ -149,6 +154,9 @@
 
         <profile>
             <id>spark_3.x</id>
+            <activation>
```
glorysdj, line 157: this should be removed

zoo/pom.xml
```diff
@@ -33,6 +34,9 @@
         <scala-library-scope>compile</scala-library-scope>
         <spark-mllib-scope>compile</spark-mllib-scope>
         <scala.major.version>2.11</scala.major.version>
+        <akka.http.version>10.1.11</akka.http.version>
```
glorysdj, line 37: why need this since you have put it in another profile

glorysdj:
please clean the code, dont send the unnecessary changes.

glorysdj:
and refer to https://github.com/intel-analytics/analytics-zoo/commit/b1c42529ec73eb553ece868d3e79832663fde7b4
may include/extend ZooGrpcServer to define FrontEndGRPCApp
also align the maven dependency versions with versions in the commit
please define the protos in a grpc.protos package
put the generated code in a grpc.services/generated package
only the FrontEndGRPCApp in grpc package

zoo/pom.xml
Gavin-Sun555: originally, we have the dependency
```
<dependency>
            <groupId>com.typesafe.akka</groupId>
            <artifactId>akka-http_${scala.major.version}</artifactId>
            <version>10.1.11</version>
            <version>${akka.http.version}</version>
        </dependency>
```
I need to preserve it for the default profile, while this version is not compatible with our new grpc serving. Hence, I need to add this to maintain different dependency version

jason-dai:
We should use https://github.com/intel-analytics/analytics-zoo/pull/4273 as the base for all GRPC code in Zoo

zoo/src/main/scala/com/intel/analytics/zoo/serving/grpc/FrontEndGRPCServiceImpl.scala
```diff
@@ -0,0 +1,435 @@
...
+import com.intel.analytics.zoo.serving.grpc.service.generated._
+import io.grpc.stub.StreamObserver
+
+import scala.collection.mutable
+import scala.concurrent.Future
+
+object FrontEndGRPCServiceImpl extends Supportive{
+  def main(args: Array[String]): Unit = {
+
+    // WARN: Firstly, only single connection is supported
```
glorysdj, line 35: only single connection is supported? what does this mean?

zoo/src/test/scala/com/intel/analytics/zoo/serving/http/MockMultipleServingHttpClient.scala
```diff
@@ -26,6 +26,8 @@ import akka.http.scaladsl.model._
 import akka.http.scaladsl.unmarshalling.Unmarshal
```
glorysdj, line 26: change this to unit tests? so the ut can be auto run in ci/cd jobs

glorysdj:
please also add a readme to show:
how to build the jar
how to start grapc server
how to tests the grpc perf
etc

Gavin-Sun555:
Jenkins Link: http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-Validation/6337/

PR #4362, opened 2021-07-26
Chronos: distributed fashion TSDataset (XShardTSDataset) first push

TheaperDeng:
XShardsTSDataset set internal data structure as XShards of pandas dataframe. And data parallelization is on each sub df based on id_col.

1. fully utilized `chronos.data.utils`, which makes `XShardTSDataset` can be easily developed with less effort and have **nearly all the features as `TSDataset` with the same behaviour**.
2. We can utilize `orca.data.pandas` to read files from **both local or distributed source**.
3. Since xshard is distributed, `XshardTSDataset` makes sure that our Chronos workflow doesn't need to collect all data to a single node for processing or feature engineering 
4. `XShardTSDataset` has better performance when we have multiple ids and each ids has a long time series.
5. Only `from_xshard`, `roll`(except for id_sensitive=True), `to_xshard` is implemented for now.


pyzoo/zoo/chronos/data/experimental/xshards_tsdataset.py
```diff
@@ -0,0 +1,199 @@
...
+        return item
+    if item is None:
+        return []
+    _check_type(item, name, expect_type)
+    return [item]
+
+
+def _check_type(item, name, expect_type):
+    assert isinstance(item, expect_type),\
+        f"a {str(expect_type)} is expected for {name} but found {type(item)}"
```
yushan111, line 199: move to data/utils？ since tsdataset might also use

pyzoo/zoo/chronos/data/experimental/xshards_tsdataset.py
```diff
@@ -0,0 +1,199 @@
...
+        feature_col = _to_list(feature_col, "feature_col") if feature_col is not None \
+            else self.feature_col
+        target_col = _to_list(target_col, "target_col") if target_col is not None \
+            else self.target_col
+        self.numpy_shards = self.shards.transform_shard(roll_timeseries_dataframe,
+                                                        None, lookback, horizon,
+                                                        feature_col, target_col)
+        return self
+
+    def to_xshards(self):
```
yushan111, line 175: Shall we change to `to_np_xshards` to avoid misinterpretations？

yushan111 (commented):
Others LGTM

pyzoo/zoo/chronos/data/experimental/xshards_tsdataset.py
TheaperDeng: I think this related to the customers' high level understanding of our product. If he/she is only a Chronos user, then `to_xshards` is fine and may be better. If he/she is a AZ user, then `to_xshards` may bring up questions like "to xshards of what?"

TheaperDeng:
pr validation combine pass: (except scala style):
http://10.112.231.51:18888/job/ZOO-PR-Validation/6212/
http://10.112.231.51:18888/job/ZOO-PR-Scala-Spark-3.1-k216/1646/
http://10.112.231.51:18888/job/ZOO-PR-Python-ExampleTests-py37-Spark3/1749/

Issue #3426, opened 2021-01-25
Automl leaderboard usage improvement


TheaperDeng:
Some issues/bugs needs to be fixed:

- [x] AutoTS should involve a param called `name` which helps the user to find the log files (#3434 )
- [x] Sorting malfunction in Tensorboard (#4620 )
- [x] Add additional address column (#3434 )
- [x] Each iter's result should be recorded (#3434 )
- [x] Remove the conditional import for tensorboardx (#3434 )
- [x] Change setup and doc tensorboardx version from 2.0 to 2.1

TheaperDeng:
For the Sorting malfunction in Tensorboard, native ray tune does not have this issue.

PR #4368, opened 2021-07-27
Fix tensorboard hparam sorting issue

cabuliwallah:
#3426 
When a hparam column contains non-numeric value, it leads to sorting malfunction in Tensorboard.
Still need work on showing these columns.


TheaperDeng:
string type is definitely supported by tensorboard hparams. Still we need some effort to do it. :(

TheaperDeng:
and we found it, #4620

PR #4370, opened 2021-07-28
add orca submit

Adria777:
orca-submit-train, orca-submit-cluster:
1.envs
2.check nfs server
3.deploy pvc
4.copy data
5.submit spark job

orca clean:
undeploy pvc and delete related pods


jason-dai:
Is this for K8s only?

glorysdj:
this is for k8s only, so please make a folder named with orca-submit-k8s, add a environment.sh and source it in orca-clean-k8s, orca-submit-k8s, try to unify client/cluster mode, and others per our discussion @Adria777

Adria777:
create a new orca-submit request, so this old one need to be close.

PR #4382, opened 2021-07-29
Chronos: rename the notebook path

TheaperDeng:


TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-Pip-Apps-Integration-Part4/112/

PR #4394, opened 2021-07-30
bigdl 2.0 scala dllib and common construct

dding3:


jason-dai:
Change the path `analytics-zoo/spark/dl/src/` to `analytics-zoo/scala/dllib/src/`

dding3:
> Change the path `analytics-zoo/spark/dl/src/` to `analytics-zoo/scala/dllib/src/`

Can we update the path to bigdl/dllib/src/main/scala? as we also have some java code which defined some classes and needed be used by scala. Checked spark mllib path, it is `spark/mllib/src/main/scala/org/apache/spark/mllib`

jason-dai:
> > Change the path `analytics-zoo/spark/dl/src/` to `analytics-zoo/scala/dllib/src/`
> 
> Can we update the path to bigdl/dllib/src/main/scala? as we also have some java code which defined some classes and needed be used by scala. Checked spark mllib path, it is `spark/mllib/src/main/scala/org/apache/spark/mllib`

Then how to you plan to add the python models? E.g., we will have both python dllib and scala dllib. One option is to put all python and scale modules under `bigdl/python` and `bigdl/scala` respectively; any other thoughts?

dding3:
> > > Change the path `analytics-zoo/spark/dl/src/` to `analytics-zoo/scala/dllib/src/`
> > 
> > 
> > Can we update the path to bigdl/dllib/src/main/scala? as we also have some java code which defined some classes and needed be used by scala. Checked spark mllib path, it is `spark/mllib/src/main/scala/org/apache/spark/mllib`
> 
> Then how to you plan to add the python models? E.g., we will have both python dllib and scala dllib. One option is to put all python and scala modules under `bigdl/python` and `bigdl/scala` respectively; any other thoughts?

Maybe we can put all python modules under "pybigdl" and all scala modules under ""bigdl"? Checked spark, seems its python modules are put under "python" folder.

jason-dai:
> > > > Change the path `analytics-zoo/spark/dl/src/` to `analytics-zoo/scala/dllib/src/`
> > > 
> > > 
> > > Can we update the path to bigdl/dllib/src/main/scala? as we also have some java code which defined some classes and needed be used by scala. Checked spark mllib path, it is `spark/mllib/src/main/scala/org/apache/spark/mllib`
> > 
> > 
> > Then how to you plan to add the python models? E.g., we will have both python dllib and scala dllib. One option is to put all python and scala modules under `bigdl/python` and `bigdl/scala` respectively; any other thoughts?
> 
> Maybe we can put all python modules under "pybigdl" and all scala modules under ""bigdl"? Checked spark, seems its python modules are put under "python" folder.

pybigdl is an awkward name; let's just use python and scala instead.

dding3:
Close it as it's duplicate to https://github.com/intel-analytics/analytics-zoo/pull/4448

PR #4400, opened 2021-08-03
create zoo-spark-k8s

zzti-bsj:
This pr is to reduce the size of hyperzoo imgae by removing the python dependencies and packaging them into a conda package, passing the conda package to a remote storage, and leverage the remote storage feature of spark 3.0+ to retrieve the python env from it.

Theoretically, it can be used by both client and cluster mode, if the plused python libs only imported in executor part and not in the driver part. please refer to
https://spark.apache.org/docs/latest/api/python/user_guide/python_packaging.html and https://conda.github.io/conda-pack/spark.html
But for analytics zoo examples, most driver code are using extra python libs, so only client mode is supported for analytics zoo on k8s.

This PR can reduce the image size from 4.9G to 1.9G, but need another conda package about 2G.

This PR is only for spark 3.0+.

BTW, this pr also supports running analytics zoo examples on k8s with code/jar/data file stored in remote storgae like http.


docker/zoo-spark-k8s/install-zoo-conda.sh
```diff
@@ -0,0 +1,27 @@
+# install conda env at client mode
+
+BASE_DIR=/opt/spark/work-dir
+PYTHON_VERSION=${PYTHON_VERSION}
+CONDA_ENV_NAME=${CONDA_ENV_NAME}
+MINICONDA=${MINICONDA}
+REQUIREMENTS=${BASE_DIR}/requirements.txt
```
glorysdj, line 7: where is the requiements.txt?

zzti-bsj: the `requirements.txt` needs to be created by users according to the python package they need after starting the container

PR #4410, opened 2021-08-04
grpc pom fix to exclude scala2.11 only support 2.12(spark 3.0)

Litchilitchy:


glorysdj:
should exclude in scala-maven-plugin

 <artifactId>scala-maven-plugin</artifactId>
                        <configuration>
                            <excludes>
                                <exclude>**/com/intel/analytics/zoo/examples/serving/**</exclude>
                                <exclude>**/com/intel/analytics/zoo/serving/**</exclude>
                                <exclude>**/com/intel/analytics/zoo/grpc/**</exclude>

Issue #4406, opened 2021-08-04
jenkins: python integration test failed


pinggao187:
~~~
#1 Start ray horovod pytorch example tests
Traceback (most recent call last):
  File "/opt/work/jenkins/workspace/ZOO-NB-Python-integration-test-horovod/pyzoo/zoo/examples/orca/learn/horovod/pytorch_estimator.py", line 149, in <module>
    train_example(workers_per_node=args.workers_per_node)
  File "/opt/work/jenkins/workspace/ZOO-NB-Python-integration-test-horovod/pyzoo/zoo/examples/orca/learn/horovod/pytorch_estimator.py", line 90, in train_example
    }, backend="horovod")
  File "/opt/work/conda/envs/py37-horovod-pytorch/lib/python3.7/site-packages/zoo/orca/learn/pytorch/estimator.py", line 95, in from_torch
    backend=backend)
  File "/opt/work/conda/envs/py37-horovod-pytorch/lib/python3.7/site-packages/zoo/orca/learn/pytorch/estimator.py", line 140, in __init__
    workers_per_node=workers_per_node)
  File "/opt/work/conda/envs/py37-horovod-pytorch/lib/python3.7/site-packages/zoo/orca/learn/pytorch/pytorch_ray_estimator.py", line 108, in __init__
    ray_ctx = RayContext.get()
  File "/opt/work/conda/envs/py37-horovod-pytorch/lib/python3.7/site-packages/zoo/ray/raycontext.py", line 453, in get
    ray_ctx.init()
  File "/opt/work/conda/envs/py37-horovod-pytorch/lib/python3.7/site-packages/zoo/ray/raycontext.py", line 549, in init
    redis_address=redis_address)
  File "/opt/work/conda/envs/py37-horovod-pytorch/lib/python3.7/site-packages/zoo/ray/raycontext.py", line 631, in _start_driver
    return ray.init(**init_params)
  File "/opt/work/conda/envs/py37-horovod-pytorch/lib/python3.7/site-packages/ray/_private/client_mode_hook.py", line 47, in wrapper
    return func(*args, **kwargs)
TypeError: init() got an unexpected keyword argument 'namespace'
~~~
jenkins link:
http://10.112.231.51:18888/view/ZOO-NB/job/ZOO-NB-Python-integration-test-horovod/183/

hkvision:
@yushan111 Take a look. Seems it is related to upgrading to ray 1.5.0

[...]

PR #4412, opened 2021-08-05
fix integration test fail due to compatible to ray 1.5.0 

yushan111:
To fix issue #4406


yushan111:
Running jenkins: 
http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-Python-integration-test/77/

PR #4426, opened 2021-08-06
Chronos: Revision on the forecasting section (concepts and forecaster) in Chronos user guide

TheaperDeng:
please preview the doc here: https://junweitestdoc.readthedocs.io/en/forecaster-tutorial-doc/doc/Chronos/Overview/chronos.html#forecasting


yushan111:
In the RR style image
image
Shall we use different colors or outlines to differentiate input and output? 

And we might also add a column to indicate whether the forecaster is based on deep learning model or statistical methods
image

shane-huang:
Below descripts says a lot about last observed timestamp. Maybe we should add a pointer to the last observed timestamp in the figure. 

>Traditionally, Time series forecasting problem was formulated with rich mathematical fundamentals and statistical models. Typically, one model can only handle one time series and fit on the whole time series before the last observed timestamp and predict the next few steps. Training(fit) is needed every time you change the last observed timestamp.

image

shane-huang:
In 4.1.2, there're descriptions about `input_feature_num` and `output_feature_num`, but there's no corresponding pointers in the RR style figure. Maybe we should add them. Or readers might be confused what's the relationship between targets, features, input_feature_num and output_feature_num.

TheaperDeng:
> In the RR style image
> image
> Shall we use different colors or outlines to differentiate input and output?
> 
> And we might also add a column to indicate whether the forecaster is based on deep learning model or statistical methods
> image

it's hard to say if the model is based on deep learning model or statistical methods, e.g. MTNet has autoregression as a path.

TheaperDeng:
pr validation pass:
http://10.112.231.51:18888/job/ZOO-PR-Validation/6268/

PR #4440, opened 2021-08-09
Fix cut bins

hkvision:
- Output type cast to int
- Support output col the same as input col


hkvision:
jenkins passed: http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-Validation/6252/
Tested on recsys data.

PR #4448, opened 2021-08-11
Migrate bigdl to bigdl2.0

dding3:


jason-dai:
This PR contains too many commits history from Zoo; and also why is there https://github.com/intel-analytics/analytics-zoo/tree/b54ff12550ad6f473051d31432f84fc181870dce/scala/dllib/src/main/java/com/intel/analytics/bigdl/grpc?

I think you should open a new PR with BigDL commits instead.

PR #4474, opened 2021-08-12
Add union column support for generate string index

jenniew:
Add union column support for generate string index.
The column input for generate string index can include {'src_cols": [xx, yy], 'col_name': zz}


jenniew:
Run Jenkins: http://10.112.231.51:18888/job/ZOO-PR-Validation/6273/
Failed with unrelated issue

PR #4484, opened 2021-08-13
Add sort, to_pandas, from_pandas in table

songhappy:


pyzoo/zoo/friesian/feature/table.py
```diff
@@ -658,6 +658,34 @@ def col(self, name):
...
+    @staticmethod
+    def from_pandas(pandas_df):
+        """
+        Returns the contents of this :class:`pandas.DataFrame` as Table
+        :param pandas_df: pandas dataframe
+        """
+        spark = OrcaContext.get_spark_session()
+        spark.conf.set("spark.sql.execution.arrow.enabled", "true")
+        sparkDF = spark.createDataFrame(pandas_df)
+        return Table(sparkDF)
```
jason-dai, line 688: `self._clone(sparkDF)`

songhappy: > self._clone(sparkDF)

there is no self in this method, cause it has to have a sparkDF to init Table, current implementation used a static method, similar to read_parquet()

songhappy:
http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-Validation/6297/
failure is irrelevant to this PR

pyzoo/zoo/friesian/feature/table.py
jason-dai: Then you should move this to `FeatureTable` and return a `FeatureTable` instead; it does not make sense to return a base class `Table`

PR #4497, opened 2021-08-17
add migration guide

dding3:


MigrationGuidance.md
```diff
@@ -0,0 +1,47 @@
...
+
+   And below classes remain the same:
+
+   ```com.intel.analytics.bigdl.utils.Engine```
+
+   ```com.intel.analytics.bigdl.utils.LoggerFilter```
+
+   ```com.intel.analytics.bigdl.utils.RandomGenerator```
+
+   ```com.intel.analytics.bigdl.utils.Shape```
```
jason-dai, line 28: Two questions:
1) Are these all the classes in `com.intel.analytics.bigdl.utils`? Can we say `com.intel.analytics.bigdl.utils.XYZ` remain the same?
2) What if we just make them `com.intel.analytics.bigdl.dllib.utils`?

dding3: Bigdl0.x com.intel.analytics.bigdl.utils.XYZ is split into two parts, some of them moved to dllib.utils as they need to use the classes defined in dllib(eg. tensor, table) While others moved to `common ` which doesn't need use dllib class.

We can move all bigdl0.x `com.intel.analytics.bigdl.utils` to `com.intel.analytics.bigdl.dllib.utils`, but there will be no `common` module.

PR #4501, opened 2021-08-18
Migrate bigdl python code to bigdl2.0 

dding3:
1. Migrate bigdl python code
2. Migrate bigdl python test code and scripts
3. Support create whl


dding3:
example test status:
python_example_test_status

dding3:
ut passed on local env.

PR #4507, opened 2021-08-18
Add method option to cross_hash_encode

hkvision:


hkvision:
jenkins: http://10.112.231.51:18888/job/ZOO-PR-Validation/6298/

PR #4511, opened 2021-08-18
uncomment EngineSpec

dding3:


dding3:
jekins pass http://10.112.231.51:18080/job/BigDL-PRVN-lite/1230/

PR #4516, opened 2021-08-19
Chronos: simplify the import line

TheaperDeng:
forcaster now can be directly import from `zoo.chronos.model.forecast`.
automodel now can be directly import from `zoo.chronos.autots.model`


yushan111:
Shall we import auto model with `zoo.chronos.autots`

PR #4524, opened 2021-08-19
Friesian Table add support for gen_string_index and encode_string  with split string

jenniew:
Friesian Table add support for gen_string_index and encode_string  with split string.


jenniew:
Jenkins: http://10.112.231.51:18888/job/ZOO-PR-Validation/6354/

PR #4526, opened 2021-08-20
Update Redis, Java entrypoint not changing yet

DiegoCao:
init-redis already verified, changing from bash

Need to changing java entrypoint (in java.manifest) to bash (follow change in scala)


glorysdj (approved):
LGTM

PR #4530, opened 2021-08-20
Move common.utils to dllib.utils

dding3:


jason-dai:
I think we should move `spark-version` to `dllib` too, as it is only used by dllib.

dding3:
> I think we should move `spark-version` to `dllib` too, as it is only used by dllib.

`spark-version` needs to has its own pom.xml so it can be built with different spark version(spark_2.x or spark_3.x) and be used as dependency of dllib.

If we move `spark-version` to `dllib`, is it ok if we change the module path as:
scala/dllib/dlib/src
scala/dllib/sparkversion/

dding3:
@jason-dai

dding3:
scala jekins pass http://10.112.231.51:18080/job/BigDL-PRVN-lite/1235/ while the python doctest failed, however it can pass on my local env.

Merge this PR first to unblock other work

PR #4534, opened 2021-08-23
[PPML] Fix script and readme in Graphene cluster serving

ColossusChang:
- Updated script to be compatible with Log4j version 2
- Updated readme scripts to forward ports


glorysdj (approved):
LGTM

Issue #4538, opened 2021-08-23
Chronos: some api doc accidentally broke because of numpy version in sphinx


TheaperDeng:
https://analytics-zoo.readthedocs.io/en/latest/doc/PythonAPI/Chronos/autotsestimator.html
https://analytics-zoo.readthedocs.io/en/latest/doc/PythonAPI/Chronos/tsdataset.html

PR #4539, opened 2021-08-23
Chronos: add Xshardstsdataset api doc and fix the api doc bug due to the numpy

TheaperDeng:
related to #4538


TheaperDeng:
pr validation passed:
http://10.112.231.51:18888/job/ZOO-PR-Validation/6314/

PR #4553, opened 2021-08-24
Bigdl-nano example: Modifications on Pl finetune

leonardozcm:


python/nano/example/pl-finetune/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Prepare the environment
+We recommend you to use [Anaconda](https://www.anaconda.com/distribution/#linux) to prepare the environment.
+```
+conda create -n nano python=3.7  # "nano" is conda environment name, you can use any name you like.
+conda activate nano
+pip install jsonargparse[signatures]
+conda install PyTurboJPEG -c conda-forge
+
+pip install bigdl-nano[pytorch]
```
jason-dai, line 14: Anyway to simply the installation?

python/nano/example/pl-finetune/README.md
```diff
@@ -0,0 +1,52 @@
+# Bigdl-nano Fine-tune example on Cat vs. Dog dataset
+
+This example illustrates how to apply bigdl-nano optimizations on a fine-tuning case based on pytorch-lightning framework. For the sake of this example, we train the proposed network(by default, a ResNet50 is used) on the [cats and dogs dataset](https://storage.googleapis.com/mledu-datasets/cats_and_dogs_filtered.zip), which consists both [freezen and unfreezen stages](https://github.com/PyTorchLightning/pytorch-lightning/blob/495812878dfe2e31ec2143c071127990afbb082b/pl_examples/domain_templates/computer_vision_fine_tuning.py#L21-L35).
```
jason-dai, line 3: We should say explicitly that using bigdl-nano, we will speedup the official PL finetune example by 5x with only 3 lines of change.

python/nano/example/pl-finetune/pl-finetune.py
```diff
@@ -80,6 +81,7 @@
 
 
 class MilestonesFinetuning(BaseFinetuning):
+
```
jason-dai, line 84: why this change?

python/nano/example/pl-finetune/pl-finetune.py
```diff
@@ -110,6 +112,7 @@ def to_dict(self):
 
 
 class CatDogImageDataModule(LightningDataModule):
+
```
jason-dai, line 115: why this change?

python/nano/example/pl-finetune/pl-finetune.py
```diff
@@ -180,6 +183,7 @@ def val_dataloader(self):
 
 
 class TransferLearningModel(pl.LightningModule):
+
```
jason-dai, line 186: why this change?

python/nano/example/pl-finetune/pl-finetune.py
```diff
@@ -274,8 +278,7 @@ def validation_step(self, batch, batch_idx):
         self.log("val_loss", self.loss(y_logits, y_true), prog_bar=True)
 
         # 3. Compute accuracy:
-        self.log("val_acc", self.valid_acc(
-            y_scores, y_true.int()), prog_bar=True)
+        self.log("val_acc", self.valid_acc(y_scores, y_true.int()), prog_bar=True)
```
jason-dai, line 281: why this change?

python/nano/example/pl-finetune/pl-finetune.py
```diff
@@ -292,6 +295,7 @@ def configure_optimizers(self):
 
 
 class MyLightningCLI(LightningCLI):
+
```
jason-dai, line 298: why this change?

python/nano/example/pl-finetune/pl-finetune.py
```diff
@@ -314,10 +318,13 @@ def instantiate_trainer(self):
 
 
 def cli_main():
+    from bigdl.nano.pytorch.trainer import Trainer
+
     MyLightningCLI(TransferLearningModel, CatDogImageDataModule,
-                   seed_everything_default=1234)
+                   seed_everything_default=1234, trainer_class=Trainer)
```
jason-dai, line 324: instead of `import`, can we add `trainer_class=bigdl.nano.pytorch.trainer.Trainer`?

python/nano/example/pl-finetune/pl-finetune.py
```diff
@@ -314,10 +318,13 @@ def instantiate_trainer(self):
...
 
 
 if __name__ == "__main__":
     cli_lightning_logo()
     cli_main()
+
```
jason-dai, line 330: why this change?

python/nano/example/pl-finetune/README.md
leonardozcm: > Anyway to simply the installation?

jsonargparse is excluded from our conda package, it is only used in this case.
`pip install` do not fully support PyTurboJPEG I think, these two lines could be replaced to
`conda install bigdl-nano.pytorch -c conda-forge -c pytorch -c chriskafka`

python/nano/example/pl-finetune/pl-finetune.py
leonardozcm: > instead of `import`, can we add `trainer_class=bigdl.nano.pytorch.trainer.Trainer`?

I think this also need to import bigdl.nano at least, you can't simply use classes in some package without telling interpreter where to locate the package.

python/nano/example/pl-finetune/README.md
leonardozcm: Yes, fixed

python/nano/example/pl-finetune/README.md
yangw1234: `jsonargparse` is specific to this example.  We should not need PyTurboJPEG if there is a `libjpegturbo.so` installed on the system. We are planning to include this library in our whl.

leonardozcm:
http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-NanoTests/

python/nano/example/pl-finetune/README.md
```diff
@@ -0,0 +1,52 @@
+# Bigdl-nano Fine-tune example on Cat vs. Dog dataset
+
+This example illustrates how to apply bigdl-nano optimizations on a fine-tuning case based on pytorch-lightning framework. For the sake of this example, we train the proposed network(by default, a ResNet50 is used) on the [cats and dogs dataset](https://storage.googleapis.com/mledu-datasets/cats_and_dogs_filtered.zip), which consists both [freezen and unfreezen stages](https://github.com/PyTorchLightning/pytorch-lightning/blob/495812878dfe2e31ec2143c071127990afbb082b/pl_examples/domain_templates/computer_vision_fine_tuning.py#L21-L35). With all the optimizations provided by bigdl-nano, we could achieve 5x times faster than the official PL finetune example in only 3 lines of changes.
```
yangw1234, line 3: should be `frozen and unfrozen`?

PR #4558, opened 2021-08-25
disable cpu binding by default

yangw1234:
Current stategy does not work in some cases on k8s. Make default to false first.


yangw1234:
http://10.112.231.51:18888/job/ZOO-PR-Validation-Other/6391/

Issue #4425, opened 2021-08-06
RayEstimator need more friendly error message to report memory error.


yushan111:
Better to guide user to tune their memory settings. 

- [ ] the way that yarn/cdh/local/... report memory error
- [ ] capture the error and more friendly message

yushan111:
Yarn report error as below.  It guide user to increase spark.yarn.executor.memoryOverhead
`ERROR YarnScheduler:70 - Lost executor 1 on Almaren-Node-164: Container killed by YARN for exceeding memory limits.  160.2 GB of 160 GB physical memory used. Consider boosting spark.yarn.executor.memoryOverhead or disabling yarn.nodemanager.vmem-check-enabled because of YARN-4714.`

However it may take some time for user to locate it.
The last error messages user may find are:
```
Traceback (most recent call last):
  File "train_wnd_tf2.py", line 258, in <module>
    label_cols=[column_info.label])
  File "/root/anaconda3/envs/recsys-kai/lib/python3.7/site-packages/zoo/orca/learn/tf2/estimator.py", line 231, in fit
    ray_xshards = process_spark_xshards(data, self.num_workers)
  File "/root/anaconda3/envs/recsys-kai/lib/python3.7/site-packages/zoo/orca/data/utils.py", line 300, in process_spark_xshards
    ray_xshards = RayXShards.from_spark_xshards(data)
  File "/root/anaconda3/envs/recsys-kai/lib/python3.7/site-packages/zoo/orca/data/ray_xshards.py", line 355, in from_spark_xshards
    return RayXShards._from_spark_xshards_ray_api(spark_xshards)
  File "/root/anaconda3/envs/recsys-kai/lib/python3.7/site-packages/zoo/orca/data/ray_xshards.py", line 383, in _from_spark_xshards_ray_api
    ray.get([v.get_partitions.remote() for v in partition_stores.values()])
  File "/root/anaconda3/envs/recsys-kai/lib/python3.7/site-packages/ray/_private/client_mode_hook.py", line 47, in wrapper
    return func(*args, **kwargs)
  File "/root/anaconda3/envs/recsys-kai/lib/python3.7/site-packages/ray/worker.py", line 1456, in get
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(RuntimeError): ray::LocalStore.get_partitions() (pid=24448, ip=172.16.0.169)
  File "python/ray/_raylet.pyx", line 439, in ray._raylet.execute_task
[...]

PR #4571, opened 2021-08-26
Add ray daemon to kill ray processes

yushan111:
## What is this PR for? 
While running `RayOnSpark` in cluster, if the spark executor is killed abruptly (e.g. killed by yarn due to memory-deficiency), the ray processes will keep running until meet further error (e.g "No module named zoo"). However, the ray processes should be killed immediately if the spark executor died. 

## Related issue
#4425 

## What does this PR include? 

1. This PR added a ray daemon to watch spark executor and kill ray processes after the spark executor is died.
2. This PR removes `shutdown_hook` and `clean_fn` in `ProcessMonitor` which are also for killing ray processes. And the functions don't work at all circumstances, such as `core_dump`.

## How is this PR tested? 
This PR has only been tested on **Linux**.
In both "standalone" mode and "yarn-client" mode:

- [x] exit normally
- [x] exit with errors which occur at different stages
- [x] interrupted with Ctrl + C 
- [x] kill -9 spark executor
- [x] kill -9 driver

In "yarn-client" mode, "core dump" after ray initialized has also been tested.

## Remained issue
1. Application keeps running after one spark executor is killed abruptly, since the processes on other nodes are not influenced. 
2. Test on MAC/Win
3. Better debug log for `ray_daemon`

### DIFF
--- a/pyzoo/zoo/ray/process.py
+++ b/pyzoo/zoo/ray/process.py
@@ -105,7 +105,6 @@ class ProcessMonitor:
                 self.master.append(process_info)
             else:
                 self.slaves.append(process_info)
-        ProcessMonitor.register_shutdown_hook(extra_close_fn=self.clean_fn)
         assert len(self.master) == 1, \
             "We should got 1 master only, but we got {}".format(len(self.master))
         self.master = self.master[0]
@@ -122,29 +121,3 @@ class ProcessMonitor:
             print(self.master)
             for slave in self.slaves:
                 print(slave)
-
-    def clean_fn(self):
-        if not self.raycontext.initialized:
-            return
-        import ray
-        ray.shutdown()
-        if not self.sc:
-            print("WARNING: SparkContext has been stopped before cleaning the Ray resources")
-        if not self.sc or is_local(self.sc):
-            gen_shutdown_per_node(self.pgids, self.node_ips)([])
-
-    @staticmethod
-    def register_shutdown_hook(pgid=None, extra_close_fn=None):
-        def _shutdown():
-            if pgid:
-                gen_shutdown_per_node(pgid)(0)
-            if extra_close_fn:
-                extra_close_fn()
-
-        def _signal_shutdown(_signo, _stack_frame):
-            _shutdown()
-            sys.exit(0)
-
-        atexit.register(_shutdown)
-        signal.signal(signal.SIGTERM, _signal_shutdown)
-        signal.signal(signal.SIGINT, _signal_shutdown)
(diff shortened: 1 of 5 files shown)


hkvision:
Generally LGTM.

hkvision:
1. Shall we put your test scripts somewhere as a backup?
2. If one spark executor dies, only the raylet combined with that executor will be killed right? So what if a new executor is relaunched by spark? the number of executors and the number of raylets will be different in this case?

yushan111:
>Shall we put your test scripts somewhere as a backup?

Actually the test script is pretty simple. I just added some `time.sleep` to ensure I could interrupt or kill the executor before the application finished.  The tests are mostly manually, like `ctrl + C` or `kill -9`. I could add the test script in repo like `test/zoo/ray/integration/` and disable the auto-test for back up. Or are there other places recommended?

yushan111:
>If one spark executor dies, only the raylet combined with that executor will be killed right? So what if a new executor is relaunched by spark? the number of executors and the number of raylets will be different in this case?

The right behavior might be if one spark executor dies, we shouldn't allow any retries and just make the whole application terminate. Because even though a new spark executor is relaunched, it is not associated with any ray actors so our `RayOnSpark` just doesn't work as expected.  In case of this PR, spark will relaunch executors until the max failure is reached. and it will take a long time.

As mentioned in the first remained issue, we could open another PR to terminate the whole application at the time one spark executor dies. To be honest, I haven't come up with a good solution for that yet. I tried modifying some spark confs listed in [comment](https://github.com/intel-analytics/analytics-zoo/pull/4571#issuecomment-911416426), and it disables retry. But it might be too strict.

hkvision:
> > Shall we put your test scripts somewhere as a backup?
> 
> Actually the test script is pretty simple. I just added some `time.sleep` to ensure I could interrupt or kill the executor before the application finished. The tests are mostly manually, like `ctrl + C` or `kill -9`. I could add the test script in repo like `test/zoo/ray/integration/` and disable the auto-test for back up. Or are there other places recommended?

So put it in `test/zoo/ray/integration/` and add some simple README in case to reproduce it later?

hkvision:
> As mentioned in the first remained issue, we could open another PR to terminate the whole application at the time one spark executor dies. To be honest, I haven't come up with a good solution for that yet. I tried modifying some spark confs listed in [comment](https://github.com/intel-analytics/analytics-zoo/pull/4571#issuecomment-911416426), and it disables retry. But it might be too strict.

It is reasonable to terminate the whole application when one spark executor dies. We may discuss whether we need to do this later.

yushan111:
I will merge this PR and add test scripts/solve the remained issues in new PRs.

yushan111:
Jenkins: http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-Validation/6425/

PR #4576, opened 2021-08-27
Add Original Segmentation Example

charlieJ107:
Create the original image segmentation example


python/nano/example/tensorflow/Segmentation/Segmentation.py
```diff
@@ -0,0 +1,142 @@
+##### Copyright 2019 The TensorFlow Authors.
```
yangw1234, line 1: You should also add bigdl license.

yangw1234:
Let's make the directory names and file names all in lower case

yangw1234:
LGTM

PR #4587, opened 2021-08-27
add gen_index_dict, reindex in FeatureTable

songhappy:


pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1357,6 +1357,69 @@ def gen_values(items):
             df = df.withColumn(c.replace("item", "category"), cat_udf(pyspark_col(c)))
         return FeatureTable(df)
 
+    def reindex(self, columns=[], index_dicts=[]):
```
jason-dai, line 1360: combine this with `add_value_features`

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1357,6 +1357,69 @@ def gen_values(items):
...
+        for i, c in enumerate(columns):
+            index_dict = index_dicts[i]
+            embed_in_dims[c] = max(index_dict.values()) + 1
+            spark = OrcaContext.get_spark_session()
+            index_dict_bc = spark.sparkContext.broadcast(index_dict)
+            index_lookup = lambda x: index_dict_bc.value.get(x, 0)
+            tbl = tbl.apply(c, c, index_lookup, "int")
+        return tbl, embed_in_dims
+
+    def gen_index_dicts(self, columns=[], freq_limit=10):
```
jason-dai, line 1389: need a more meaningful name for this method

pyzoo/zoo/friesian/feature/table.py
songhappy: although both add_value_features and reindex share a key value lookup functionality, the main logic is different. add_value_features takes one table of key value pairs for all the columns, it replaces the key with value, and add one corresponding column. Columns to be worked on are int or list[int], list[list[int]], and share the same key value pair table. On the other hand, reindex takes a list of dictionaries(could be tbls), each column has a specific dictionary, reindex replaces the old index with new index from a dictionary correspondingly, without adding any new columns.
It sounds more clear to me if we keep two separate methods.

pyzoo/zoo/friesian/feature/table.py
songhappy: gen_reindex_mapping

songhappy:
jenkins http://10.112.231.51:18888/job/ZOO-PR-Validation/6366/
abortions is irrelevant

pyzoo/zoo/friesian/feature/table.py
jason-dai: > although both add_value_features and reindex share a key value lookup functionality, the main logic is different. add_value_features takes one table of key value pairs for all the columns, it replaces the key with value, and add one corresponding column. Columns to be worked on are int or list[int], list[list[int]], and share the same key value pair table. On the other hand, reindex takes a list of dictionaries(could be tbls), each column has a specific dictionary, reindex replaces the old index with new index from a dictionary correspondingly, without adding any new columns.
> It sounds more clear to me if we keep two separate methods.

I think these logic can be easily combined; providing too many ops can be confusing to the users.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1357,6 +1357,69 @@ def gen_values(items):
...
+         :param freq_limit: int, dict or None. Indices with a count below freq_limit
+               will be omitted. Can be represented as either an integer or dict.
+               For instance, 15, {'col_4': 10, 'col_5': 2} etc. Default is 10,
+
+        :return: a dictionary of list of dictionaries, a mapping from old index to new index
+                new index starts from 1, save 0 for default
+         """
+        if isinstance(columns, str):
+            columns = [columns]
+        assert isinstance(columns, list), "columns should be str or a list of str"
```
jason-dai, line 1402: "but get a ..."

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1357,6 +1357,69 @@ def gen_values(items):
...
+        if isinstance(freq_limit, int):
+            freq_limit = {col: freq_limit for col in columns}
+        assert isinstance(freq_limit, dict), "freq_limit should be int or dict"
```
jason-dai, line 1405: "but get ..."

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1357,6 +1357,69 @@ def gen_values(items):
...
+
+        index_dicts = []
+        for c in columns:
+            c_count = self.select(c).group_by(c, agg={c: "count"}).rename(
+                {"count(" + c + ")": "count"})
+            c_count = c_count.filter(pyspark_col("count") >= freq_limit[c])\
+                .order_by("count", ascending=False)
+            c_count_pd = c_count.to_pandas()
```
jason-dai, line 1413: can we use Spark Dataframe operator instead of Pandas operator here?

pyzoo/zoo/friesian/feature/table.py
songhappy: merged into add_value_features()

pyzoo/zoo/friesian/feature/table.py
songhappy: yes, we can always spark DataFrame.
Two reasons that I chose pandas dataframe. 1, output of this methods go with models, should be saved in local, easier to handle. 2, the easiest way to convert spark dataframe to a dictionary is to convert it to pandas dataframe first.

jason-dai: Have you tested the performance, especially if there are many unique values for a column, or many columns?

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1316,75 +1318,65 @@ def join(self, table, on=None, how=None, lsuffix=None, rsuffix=None):
         joined_df = self.df.join(table.df, on=on, how=how)
         return FeatureTable(joined_df)
 
-    def add_value_features(self, key_cols, tbl, key, value):
+    def add_value_features(self, columns, tbls, key=None, value=None, reindex_only=False):
```
jason-dai, line 1321: This is too complex; I think you should keep the original `add_value_features` logic - looking up value from each key (no need to support a list of tables). The only difference is whether one needs in-place update or append operations.

pyzoo/zoo/friesian/feature/table.py
```diff
@@ -1316,75 +1318,65 @@ def join(self, table, on=None, how=None, lsuffix=None, rsuffix=None):
...
-        tbl = FeatureTable(self.df)
-        for i, c in enumerate(columns):
-            index_dict = index_dicts[i]
-            embed_in_dims[c] = max(index_dict.values()) + 1
-            spark = OrcaContext.get_spark_session()
-            index_dict_bc = spark.sparkContext.broadcast(index_dict)
-            index_lookup = lambda x: index_dict_bc.value.get(x, 0)
-            tbl = tbl.apply(c, c, index_lookup, "int")
-        return tbl, embed_in_dims
+        return FeatureTable(df), value_dims
```
jason-dai, line 1379: why do you need to return `value_dims` here? you can get it outside the method, yes?

pyzoo/zoo/friesian/feature/table.py
songhappy: reindex need a list of tables, since each column needs a dictionary to reindex. I will change back to the original version.
have two separate methods reindex and add_value_features

pyzoo/zoo/friesian/feature/table.py
songhappy: this is a byproduct of for reindex, return a reindexed table and its dimensions for later usage

pyzoo/zoo/friesian/feature/table.py
songhappy: do not have a performance number of this specific method, but for all the twitter data,     given embed_cols = ["user_id", "engaged_with_user_id", "hashtags", "present_links", "present_domains"], dimensions is around 48m, it worked fine.

pyzoo/zoo/friesian/feature/table.py
jason-dai: You may reindex one column at a time

pyzoo/zoo/friesian/feature/table.py
jason-dai: We should have the operator return the FeatureTable as much as we can, so that we may chain these operations together

pyzoo/zoo/friesian/feature/table.py
jason-dai: Please do detailed breakdown

pyzoo/zoo/friesian/feature/table.py
songhappy: that works too. 
up side: we have one method of add_value)_features() to cover original add_value_feaures and reindex functionalities.
downside: it's slightly confusing since add_value_features does all columns in the same time, but reindex only processes one column, it also makes the prepare_features slightly more ugly, and we need a loop whenever we call this method.

pyzoo/zoo/friesian/feature/table.py
songhappy: ok, removed value_dims

pyzoo/zoo/friesian/feature/table.py
jason-dai: We may then have the `reindex` method call `add_value_feaures`

BTW, why do you use Table in `add_value_feaures`, but dictionary in `reindex`?

pyzoo/zoo/friesian/feature/table.py
songhappy: given 100m users or items, the size of dictionary is 400m.

songhappy:
Jekins 
http://10.112.231.51:18888/job/ZOO-PR-Validation/6437/
failure is irrelevant.

PR #4594, opened 2021-08-30
TF2Estimator Fit Implementation by PySpark Backend

sgwhat:


yangw1234:
Could you create a new class instead of using the old ones?

PR #4597, opened 2021-08-30
migrate zoo common scala code

dding3:


dding3:
jekins pass. http://10.112.231.51:18080/job/BigDL-PRVN-lite/1245/

scala/dllib/pom.xml
```diff
@@ -82,6 +82,17 @@
             <version>0.14.0-SNAPSHOT</version>
             <scope>${bigdl-core-all-scope}</scope>
         </dependency>
+        <dependency>
+            <groupId>com.intel.analytics.zoo</groupId>
```
jason-dai, line 86: why depends on `analytics zoo`?

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/nnframes/DefaultParamsWriterWrapper.scala
```diff
@@ -0,0 +1,54 @@
+///*
```
jason-dai, line 1: Is it used?

scala/dllib/pom.xml
dding3: since we have migrate some zoo code and they will depened on analytics zoo core.

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/nnframes/DefaultParamsWriterWrapper.scala
dding3: not now. Will uncomment these files after all its dependent files are migrated(eg. common, feature). I leave these files here so we can save some efforts of migrating nnframes again :)

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/utils/Optim.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.intel.analytics.bigdl.dllib.optim.SGD.LearningRateSchedule
+//import com.intel.analytics.bigdl.dllib.zooKeras.models.{InternalOptimizerUtil}
+
+object Optim {
+
+  /**
+   * A fixed learning rate scheduler, always return the same learning rate
+   * @param lr learning rate
+   */
+  case class Fixed(lr: Double) extends LearningRateSchedule {
```
jason-dai, line 28: where is it used? and this should go to `dllib.optim`?

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/utils/PythonInterpreter.scala
```diff
@@ -0,0 +1,130 @@
+/*
```
jason-dai, line 1: I think we should move it to orca, not dllib

jason-dai:
I think some refactor and cleanup are needed; maybe we can move zoo-related common files to `bigdl.common`, so that we know where these files come from and clean them up later.

scala/dllib/pom.xml
jason-dai: I don't think dllib should depends on zoo0core, otherwise the python package will be too large.

dding3:
> I think some refactor and cleanup are needed; maybe we can move zoo-related common files to `bigdl.common`, so that we know where these files come from and clean them up later.

so zoo-related common files will be moved to `bigdl.common`, parallel with spark-version?

PR #4610, opened 2021-09-01
[PPML] Added back tm deployment

ColossusChang:
- Added tm deployment back, previously used stateful set to replace it


glorysdj (approved):
LGTM

PR #4613, opened 2021-09-01
fix zoo common comments

dding3:
1. move zootrigger from utils to dllib.optim
2. remove zoo core as dependency
3. move Optim from utils to dllib.optim
4. move pythoninterpreter from utils to orca
5. move zoo related utils to dllib.common


dding3:
jekins pass. http://10.112.231.51:18080/job/BigDL-PRVN-lite/1248/

PR #4600, opened 2021-08-31
Modifications on pytorch Segmentation example

leonardozcm:

PR #4619, opened 2021-09-02
Bigdl nano: Makes torch function replacement a plugin-like operation

leonardozcm:
For the development of ipex is on going, there are some operations or functions we can not call directly [#4600](https://github.com/intel-analytics/analytics-zoo/pull/4600#discussion_r699773873).
This pr provides a way for us to apply any workaround as an alternative if it fixs the issue temporarily. The replacement takes place whenever ipex accelerator is imported, in another word `use_ipex` is specfied as true in Trainer, but it will not affect default behaviors.

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/accelerators/ipex_torchfunctional.py
+++ b/python/nano/src/bigdl/nano/pytorch/accelerators/ipex_torchfunctional.py
@@ -1,3 +1,55 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# 
+# workaround_cross_entropy is adapted from 
+# 
+# https://github.com/pytorch/pytorch/blob/c7c711bfb88fcb0ef573125a5a8655c49156055b
+# /torch/nn/functional.py#L2767
+# 
+# Note: This license has also been called the "New BSD License" or "Modified BSD License". See also
+# the 2-clause BSD License.
+#
+# Copyright (c) 2019 The DeepGLO Project.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are permitted
+# provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of conditions
+# and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list of
+# conditions and the following disclaimer in the documentation and/or other materials provided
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/accelerators/ipex_torchfunctional.py
```diff
@@ -0,0 +1,54 @@
+import torch
```
yangw1234, line 1: license

python/nano/src/bigdl/nano/pytorch/accelerators/ipex_torchfunctional.py
```diff
@@ -0,0 +1,54 @@
...
+Tensor = torch.Tensor
+torch_nn_functional = torch.nn.functional
+_Reduction = torch.nn._reduction
+
+
+def replace_torch_function(function_name: str, replace_func: Callable):
+    setattr(torch.nn.functional, function_name, replace_func)
+
+
+def workaround_cross_entropy(
```
yangw1234, line 14: We'd better document what's different between this version and the pytorch version.

python/nano/src/bigdl/nano/pytorch/accelerators/ipex_torchfunctional.py
```diff
@@ -0,0 +1,54 @@
...
+            ignore_index=ignore_index,
+            reduce=reduce,
+            reduction=reduction,
+        )
+    if size_average is not None or reduce is not None:
+        reduction = _Reduction.legacy_get_string(size_average, reduce)
+
+    # `log_softmax(input, 1).contiguous()` makes multidimensional tensors continuous in memory
+    # May fix https://github.com/intel/intel-extension-for-pytorch/issues/175#issue-972312586
+    return torch_nn_functional.nll_loss(torch_nn_functional.log_softmax(input, 1).contiguous(),
```
yangw1234, line 40: Wouldn't it be simpler to just call `input.cpu()` and `target.cpu()` than call `torch_nn_functional.cross_entropy`?

python/nano/src/bigdl/nano/pytorch/accelerators/ipex_torchfunctional.py
yangw1234: You may also need to cite pytorch's license.

python/nano/src/bigdl/nano/pytorch/accelerators/ipex_torchfunctional.py
leonardozcm: > Wouldn't it be simpler to just call `input.cpu()` and `target.cpu()` than call `torch_nn_functional.cross_entropy`?

Will this leads to some performance loss when move tensor from xpu to cpu or losing optimizations in cross_entropy provided by ipex? I'm not sure, but only adding `contiguous()` makes minimal modifications compared with original version, so I think we can keep this.

leonardozcm:
http://10.112.231.51:18888/job/ZOO-PR-nano-ipex/20/
http://10.112.231.51:18888/job/ZOO-PR-nano-codestyle-test/26/
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/20/

yangw1234:
LGTM

jason-dai:
Please add detailed description (e.g., in the comments) for the functionality.

leonardozcm:
> Please add detailed description (e.g., in the comments) for the functionality.

sure, I will fix it

PR #4630, opened 2021-09-02
Add pip wheel support of Turbojpeg

leonardozcm:
This pr makes nano load libturbojpeg.so from wheel package itself instead of conda package, with which we can directly run `pip install bigdl-nano` without `conda install Pyturbojpeg`.


yangw1234:
LGTM

PR #4634, opened 2021-09-03
[PPML] Fix new graphene script

ColossusChang:
- Deleted bigdl related lines from bash.manifest.template
- In start-all.sh, changed `init-java.sh` and `init-redis.sh` to the actual file names
- Changed to new command to run in sgx mode


glorysdj (approved):
LGTM

PR #4638, opened 2021-09-03
[PPML] Escaped the quotes in sgx mode script

ColossusChang:
- Escaped the quotes, or else the `eval` command won't work.


glorysdj (approved):
LGTM

PR #4644, opened 2021-09-04
migrate Keras 

Le-Zheng:


jason-dai:
Need to update the migration guide (https://github.com/intel-analytics/analytics-zoo/blob/bigdl-2.0/MigrationGuidance.md)

PR #4646, opened 2021-09-06
update spark-submit

Adria777:
add "local://" in analytics-zoo-base.sh and spark-submit-python-with-zoo.sh


glorysdj (approved):
LGTM

yangw1234:
Maybe we should test it on different older versions of spark?

glorysdj:
> Maybe we should test it on different older versions of spark?

ok @Adria777  please test it in spark 2.3 version

PR #4672, opened 2021-09-07
update scripts

Adria777:
add "local" when spark version equal or greater than 2.4.4

### DIFF
--- a/scripts/analytics-zoo-base.sh
+++ b/scripts/analytics-zoo-base.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+SPARK_VERSION=2.4.3
+
 if [[ -z "${SPARK_CMD}" ]]; then
     echo "Please set SPARK_CMD environment variable"
     exit 1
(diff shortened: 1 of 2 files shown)


scripts/analytics-zoo-base.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+SPARK_VERSION=2.4.3
```
glorysdj, line 3: should not specify spark version here

scripts/analytics-zoo-base.sh
```diff
@@ -21,9 +23,22 @@ if [[ ! -f ${ANALYTICS_ZOO_JAR} ]]; then
...
     --conf spark.driver.extraClassPath=${ANALYTICS_ZOO_JAR} \
     --conf spark.executor.extraClassPath=${ANALYTICS_ZOO_JAR} \
     "$@"
+else
+    ${SPARK_HOME}/bin/${SPARK_CMD} \
+    --properties-file ${ANALYTICS_ZOO_CONF} \
+    --jars ${ANALYTICS_ZOO_JAR} \
+    --conf spark.driver.extraClassPath=${ANALYTICS_ZOO_JAR} \
+    --conf spark.executor.extraClassPath=${ANALYTICS_ZOO_JAR} \
+    "$@"
```
glorysdj, line 42: what if user pass local:// in $@?

PR #4674, opened 2021-09-07
Bigdl-Nano: Add description to ipex_torchfunctional

leonardozcm:
`from torch import Tensor`
This line tries to fix some typecheck issue in jenkins.


yangw1234:
LGTM

PR #4691, opened 2021-09-09
[PPML] Fixed helm chart files

ColossusChang:
- Moved backup files and scripts out of the `template` directory to avoid confusion for Helm


glorysdj (approved):
LGTM

PR #4700, opened 2021-09-10
Keras others uts

Le-Zheng:
remove nn/keras
uncomment predictor as dataset is migrated.


Le-Zheng:
jenkins passde: http://10.112.231.51:18080/view/BigDL-2.0/job/BigDL-PRVN-ZooKeras-scala/7/

PR #4704, opened 2021-09-10
Move orca/src/test to orca/test in BigDL2.0 

yushan111:
If not and with all `__init__.py` added to `orca/src/test`, `pytest` will automatedly prepend `orca/src` to `sys.path` (refer [here](https://docs.pytest.org/en/6.2.x/pythonpath.html#import-modes)). However, since `orca/src` contains `bigdl`, python will import all `bigdl` from `orca/src`,  which only includes `orca`, instead of from `PYTHONPATH`. In that case, we cannot import `bigdl.dllib`

For reference, the tree for bigdl2.0 `orca` is as below.

image


yushan111:
Updated version in #4712. Close this one.

PR #4715, opened 2021-09-12
fix scala style check and disable header check 

Le-Zheng:


Le-Zheng:
jenkins passed: http://10.112.231.51:18080/job/BigDL-PRVN-StyleCheck/1294/console

PR #4732, opened 2021-09-15
Chronos: test_autoprophet jenkins timeout (mac-only)

liangs6212:


TheaperDeng:
pls make sure it can fix the mac CI issue

PR #4735, opened 2021-09-15
migrate ZooModel to support 4dmm for TimeDistributed,TransformerLayer,Bert

Le-Zheng:


scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/models/common/ZooModel.scala
```diff
@@ -0,0 +1,154 @@
+/*
```
jason-dai, line 1: why do we need this class? where is it used?

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/keras/layers/TimeDistributed.scala
```diff
@@ -22,89 +22,88 @@ import com.intel.analytics.bigdl.dllib.keras.{KerasLayer}
 import com.intel.analytics.bigdl.dllib.tensor.Tensor
 import com.intel.analytics.bigdl.dllib.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.dllib.utils.Shape
-// import com.intel.analytics.zoo.models.common.ZooModel
+import com.intel.analytics.bigdl.dllib.models.common.ZooModel
```
jason-dai, line 25: why do we need this?

Le-Zheng: > why do we need this?

I found the chatbot example uses TimeDistributed. https://github.com/intel-analytics/analytics-zoo/blob/master/zoo/src/main/scala/com/intel/analytics/zoo/examples/chatbot/Train.scala#L128

jason-dai: But `Dense` is not a `ZooModel`, yes?

Le-Zheng: > But `Dense` is not a `ZooModel`, yes?

Yes. I will not migrate ZooModel and remove the ZooModel part in TimeDistributed.scala.

Le-Zheng:
Close this pr, and move changes in next pr https://github.com/intel-analytics/analytics-zoo/pull/4736

PR #4741, opened 2021-09-16
Chronos: add autoprophet example for easy user adoption

TheaperDeng:
add an example of autoprophet on nyc taxi, where users can improve their mse by 10% in 3min with not much resourse used.

The example will be changed to a notebook according to requestment later.


TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-Python-ExampleTests-py36/3824/
http://10.112.231.51:18888/job/ZOO-PR-Validation/6450/

PR #4748, opened 2021-09-17
Replace with the latest init.sh in taskmanager-deployment.yaml.back

Uxito-Ada:
Fix the mistake in real-time cluster serving


glorysdj (approved):
LGTM

PR #4762, opened 2021-09-18
Modify scripts in realtime cluster serving

Uxito-Ada:
1. Modify start-all-but-flink.sh to avoid check status before job submitted, which can cause a check fail.
2. Add some properties to make the yaml scripts more user-friendly.


ppml/trusted-realtime-ml/scala/docker-graphene/kubernetes/start-all-but-flink.sh
```diff
@@ -36,8 +36,8 @@ echo "http-frontend started"
 while ! nc -z $LOCAL_IP 10020; do
   sleep 1
 done
-./start-cluster-serving-job.sh &
```
glorysdj, line 39: why remove &?  it can be run at background.

glorysdj (approved):
LGTM

PR #4766, opened 2021-09-18
move bigdl python keras/ut to nn

Le-Zheng:


Le-Zheng:
jenkins passed: http://10.112.231.51:18080/job/BigDL-PRVN-pythontest-Spark-2.2-py36/1282/console

PR #4774, opened 2021-09-22
dien performance

songhappy:
1, add add_value_features into scala version
2, cache
3, stops before padding to have same data as recdp
4, read meta data from csv
5, have a choice to keep last seq of history
6, other minor change
now it is 4x faster than the original implementation on almaren.


songhappy:
jenkins passed
http://10.112.231.51:18888/job/ZOO-PR-Validation/6496/

PR #4777, opened 2021-09-22
Merge serving to bigdl 2.0

Litchilitchy:


jason-dai:
You should keep the commit history if possible.

Litchilitchy:
> You should keep the commit history if possible.

I keep the scala code history in #4803 and remained code here would be merged after that

Litchilitchy:
http://10.112.231.51:18888/view/ZOO-PR/job/Serving-PR-Python-Spark-2.4-py36/

PR #4782, opened 2021-09-22
tfpark/torch dependencies in Topology

Le-Zheng:


scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/nn/abstractnn/AbstractModule.scala
```diff
@@ -482,6 +482,14 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
     this.train
   }
 
+  /**
+   * Check if the model is single model (cores per node)
+   * @return
+   */
+  def isSingleModel(): Boolean = {
```
yangw1234, line 489: I think return the cpu cores the model requires might be more flexible. This way number of submodules can be computed by `cores_per_node/cores_per_model`. If the model needs all cores avaiable, it can return a -1 or 0. The default behavior is returning 1.

I think we can do this later.

jason-dai, line 489: change to `isPyTorch()` and `isTensorFlow()`

Le-Zheng:
dllib Jenkins passed: http://10.112.231.51:18888/job/BigDL-PR-Validation/3/
I will merge it first, then verify in orca testing.

PR #4789, opened 2021-09-22
Move dllib/src/test to dllib/test

Le-Zheng:
It's difficult to use git-mv-with-history  to move current floder. There is always a merge conflict of:
```
BUG: There are unmerged index entries:
BUG: 2 python/dllib/src/bigdl/dllib/keras/__init__.py
BUG: 3 python/dllib/src/bigdl/dllib/keras/__init__.py
fatal: BUG: unmerged index entries in merge-recursive.c
```
So I move the folder using `git mv`.  If you would like to check the git commit history, you need to use `git log --follow [file to path]` in cmd.


Le-Zheng:
simple jenkins test: http://10.112.231.51:18080/job/BigDL-PRVN-pythontest-Spark-2.1-py36/462/console

PR #4798, opened 2021-09-23
Change author name in license from Analytics Zoo to BigDL

yushan111:


Le-Zheng:
I disabled the header check before.  Could you help to uncomment the line https://github.com/intel-analytics/analytics-zoo/blob/bigdl-2.0/scala/scalastyle_config.xml#L37 and #L55 by removing `<!--` and `-->`
And then test scala style check on jenkins http://10.112.231.51:18888/job/BigDL-PRVN-StyleCheck/

PR #4801, opened 2021-09-23
Fix ray import psutil

### LINKED ISSUE #4791: When use OrcaContext and init_orca_context , the Ray package must be installed first
When we test OrcaContext and init_orca_context , we found they always asked to install Ray package but it should be a hard dependence to enable Ray on Spark , it should be configurable

hkvision:
Fix #4791 
The issue is brought from #4571

### DIFF
--- a/pyzoo/zoo/ray/utils.py
+++ b/pyzoo/zoo/ray/utils.py
@@ -15,7 +15,6 @@
 #
 
 import re
-import psutil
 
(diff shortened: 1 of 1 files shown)


hkvision:
http://10.112.231.51:18888/job/ZOO-PR-Validation/6461/ The failure is irrelevant

PR #4803, opened 2021-09-23
Migrate serving scala code with history

Litchilitchy:


jason-dai:
Please update license header. Otherwise LGTM.

PR #4810, opened 2021-09-23
fix scala style check in feature

Le-Zheng:


Le-Zheng:
jenkins passed: http://10.112.231.51:18888/job/BigDL-PRVN-StyleCheck/9/console

PR #4829, opened 2021-09-26
Enable orca python test with pip

yushan111:


yushan111:
Jenkins pass: 
http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-BigDL-Python-Spark-2.4-py36-ray/

BigDL2.0 orca python test with pip install has been enabled~

PR #4834, opened 2021-09-26
[BigDL-2.0]Move orca/test/dev to orca/dev/test

yushan111:


yushan111:
http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-BigDL-Python-Spark-2.4-py36-ray/7/

PR #4836, opened 2021-09-26
Enable automated example test for orca automl

yushan111:


yushan111:
http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-BigDL-Python-ExampleTests-py36-ray/2/

PR #4840, opened 2021-09-26
change jep test script path

yangw1234:


yangw1234:
http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-BigDL-Pip-JEP-UT-Spark2.4/9/

PR #4839, opened 2021-09-26
Exclude dllib from orca jar

yushan111:
After removing `bigdl-dllib`, orca jar size reduces from 389M to 245M

Before:
```
-rw-rw-r-- 1 shan shan 152822343 9月  24 10:18 bigdl-dllib-0.14.0-SNAPSHOT-jar-with-dependencies.jar
-rw-rw-r-- 1 shan shan    600532 9月  24 10:18 bigdl-dllib-0.14.0-SNAPSHOT-python-api.zip
-rw-rw-r-- 1 shan shan 389080878 9月  24 10:18 bigdl-orca-0.14.0-SNAPSHOT-jar-with-dependencies.jar
-rw-rw-r-- 1 shan shan    708745 9月  24 10:18 bigdl-orca-0.14.0-SNAPSHOT-python-api.zip

```
After:
```
-rw-rw-r-- 1 shan shan 122773321 9月  26 17:09 bigdl-dllib-0.14.0-SNAPSHOT-jar-with-dependencies.jar
-rw-rw-r-- 1 shan shan    600529 9月  26 17:09 bigdl-dllib-0.14.0-SNAPSHOT-python-api.zip
-rw-rw-r-- 1 shan shan 245061252 9月  26 17:09 bigdl-orca-0.14.0-SNAPSHOT-jar-with-dependencies.jar
-rw-rw-r-- 1 shan shan    818616 9月  26 17:09 bigdl-orca-0.14.0-SNAPSHOT-python-api.zip

```

PR #4854, opened 2021-09-28
Exclude dllib from orca jar

yushan111:
Related PR #4839 
This PR excluded dllib from orca jar and add dllib dependency in serving jar

With PR #4851, orca jar size can be reduced to 130M


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/7/

Issue #4847, opened 2021-09-27
Dllib several failed UTs


Le-Zheng:
Serval UTs are skipped temporarily on Jenkins nightly build

Check the error and mark it unignored. e.g replcaing `ignore should "ImageChannelNormalize with std not 1 work properly"` to `"ImageChannelNormalize with std not 1" should "work properly". Then run " mvn test -Dsuites=com.intel.analytics.bigdl.dllib.feature.FeatureSpec" on local to check the error.

1. com.intel.analytics.bigdl.dllib.feature.FeatureSpec.ImageChannelNormalize with std not 1 should work properly also failed on local
561000 was not equal to 234900

2. com.intel.analytics.bigdl.dllib.utils.EngineSpec should be right *** FAILED ***
  6 was not equal to 4 (EngineSpec.scala:139)
   readConf should skip blockTransferService if bigdl.network.nio is set to false *** FAILED ***

3. com.intel.analytics.bigdl.dllib.utils.EngineSpec.readConf should skip blockTransferService if bigdl.network.nio is set to false
5 was not equal to 3

4. dllib.nn.keras.nn com.intel.analytics.bigdl.keras.nn.KerasStyleSpec.save and reload model should work correctly requirement failed: unrecognized type Seq[com.intel.analytics.bigdl.dllib.nn.Graph.ModuleNode[T]]

5. dllib.nn.keras.nn com.intel.analytics.bigdl.keras.nn.KerasStyleSpec.save and reload sequential should work correctly com.intel.analytics.bigdl.dllib.utils.LayerException was thrown.

dding3:
The first failed ut can pass on my local env.

The left 4 have been fixed in https://github.com/intel-analytics/analytics-zoo/pull/4858

PR #4858, opened 2021-09-28
Fix failed dllib uts

dding3:
Fix failed ut in https://github.com/intel-analytics/analytics-zoo/issues/4847


dding3:
jekins pass. http://10.112.231.51:18888/job/BigDL-PR-Validation/18/

PR #4869, opened 2021-09-29
Upgrade IPEX to version 1.9.0

charlieJ107:
Upgrade IPEX from 1.8.0 to 1.9.0.

Pytroch upgrade from 1.8.0 to 1.9.0. 

torchvision upgrade from 0.9.0 to 0.10.0

Mannulay tested on clx-004.
According to test result, and benchmark, the thoughput will increase 5-6% without accuracy drop after upgrade.

### DIFF
--- a/python/nano/setup.py
+++ b/python/nano/setup.py
@@ -95,23 +95,26 @@ def setup_package():
 
     assert py_version.major == 3, "Nano only supports python3.x"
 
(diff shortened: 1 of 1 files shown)


charlieJ107:
Jekins Test passed in http://10.112.231.51:18888/job/ZOO-PR-nano-test/54/

PR #4883, opened 2021-10-04
migrate zoo nnframe example to dllib

dding3:


dding3:
http://10.112.231.51:18888/job/BigDL-PR-Validation/19/. one jekin job failed but it's not related to this pr.

PR #5087, opened 2021-12-21
fix 0.11.1 UT test

Le-Zheng:


Le-Zheng:
http://10.112.231.51:18888/job/ZOO-PR-Scala-Spark-2.1/6629/ passed

PR #5109, opened 2022-02-08
hotfix slfj scope on spark2

Le-Zheng:


Le-Zheng:
http://10.112.231.51:18888/job/ZOO-PR-Validation/6535/ has fixed log level not found issue

PR #5132, opened 2022-11-04
update BigDL components version

zzti-bsj:
update BigDL components version


Le-Zheng:
And other two issues in apps  https://github.com/intel-analytics/analytics-zoo/security/dependabot/179
https://github.com/intel-analytics/analytics-zoo/security/dependabot/180:
Need to add `<scope>provided</scope>`
https://github.com/intel-analytics/analytics-zoo/blob/master/apps/model-inference-examples/text-classification-inference/pom.xml#L61
https://github.com/intel-analytics/analytics-zoo/blob/master/apps/model-inference-examples/text-classification-training/pom.xml#L51

Le-Zheng:
LGTM

PR #5136, opened 2022-11-15
use PYSPARK_VERSION variable to replace pyspark version

zzti-bsj:
use PYSPARK_VERSION variable to replace pyspark version


glorysdj (approved):
LGTM

PR #5168, opened 2024-03-26
disable workflows' schedule event

liu-shaojun:
### 1. Why the change?

This submission disables workflow's scheduled event so that workflows won't run every night.


glorysdj (approved):
LGTM

PR #5169, opened 2024-03-26
release bigdl-llm

liu-shaojun:
### 1. Why the change?

This submission creates a workflow for nightly releases of bigdl-llm, we have just migrated to ipex-llm, we need to continue releasing bigdl-llm a few days.

### DIFF
--- a/python/llm/setup.py
+++ b/python/llm/setup.py
@@ -324,7 +324,7 @@ def setup_package():
                         "xpu-2-0": xpu_20_requires,
                         "xpu-2-1": xpu_21_requires,
                         "serving": serving_requires,
-                        "cpp": ["bigdl-core-cpp==" + VERSION]},
+                        "cpp": ["bigdl-core-cpp==2.5.0b20240324"]},
         classifiers=[
             'License :: OSI Approved :: Apache Software License',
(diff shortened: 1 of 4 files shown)


glorysdj (approved):
LGTM

python/llm/setup.py
```diff
@@ -168,7 +168,8 @@ def get_date_urls(base_url):
         except Exception as e:
             print("error - > ", base_url, e)
             pass
-        reg = "https://sourceforge.net/projects/analytics-zoo/files/bigdl-llm/[0-9]{4}-[0-9]{1,2}-[0-9]{1,2}/"
+        # reg = "https://sourceforge.net/projects/analytics-zoo/files/bigdl-llm/[0-9]{4}-[0-9]{1,2}-[0-9]{1,2}/"
+        reg = "https://sourceforge.net/projects/analytics-zoo/files/bigdl-llm/2024-03-24/"
```
Oscilloscope98, line 172: Why we fix the version here?

liu-shaojun: The code of bigdl-llm has not been updated since 20240324, there may be problems if the core-xe or binary version is not fixed.

PR #5174, opened 2024-04-11
Fix transformers version dependabot issues

liu-shaojun:
### 1. Why the change?

https://github.com/intel-analytics/BigDL-2.x/security/dependabot/498
https://github.com/intel-analytics/BigDL-2.x/security/dependabot/499
https://github.com/intel-analytics/BigDL-2.x/security/dependabot/497


glorysdj (approved):
LGTM

PR #5177, opened 2024-06-19
unpin scikit-learn version to fix dependabot issue

liu-shaojun:
### 1. Why the change?

Unpin scikit-learn version to fix dependabot issue: https://github.com/intel-analytics/BigDL-2.x/security/dependabot/501


plusbang (approved):

glorysdj (approved):
LGTM

